Issue #2, opened 2018-01-10
Create Initial 'Person' Entity


baileyckelly:
*User Story*
As a Kids First developer, I would like basic CRUD functionality to work with a Person.

*Tasks*
- [x] Create Model
- [ ] Resource
- [ ] API: Put, Patch, Delete, Get

*Acceptance Criteria*
Documentation
Have Put, Patch, Delete, Get
Unit Tests - Tests on Model, Resource, API

PR #10, opened 2018-01-10
✨ Create person entity

dankolbman:
Create person entity for #2 

- [x] - `PUT`
- [x] - `GET`
- [ ] - `PATCH`
- [ ] - `DELETE`
- [ ] - docs
- [ ] - tests


tests/test_person_api.py
```diff
@@ -0,0 +1,142 @@
...
+
+        resp = json.loads(response.data.decode("utf-8"))
+        self.assertEqual(response.status_code, 200)
+
+        response = self.client.get(url_for(EXISTING_PERSON_URL,
+                                           person_id=kf_id),
+                                   headers=self._api_headers())
+
+        resp = json.loads(response.data.decode("utf-8"))
+        self.assertEqual(response.status_code, 404)
```
dankolbman: Also worth checking that the actual row was dropped in the database. It's possible that the api could hold the transaction open making object deleted only in the api's session.

dataservice/api/person/person.py
```diff
@@ -0,0 +1,125 @@
...
+        Get all persons
+        """
+        persons = model.Person.query.all()
+        return {'status': 200,
+                'message': '{} persons'.format(len(persons)),
+                'content': {'persons': persons}}, 200
+
+
+@person_api.route('/')
+class NewPerson(Resource):
```
dankolbman: Any reason for the addition of this class rather than placing the 'post' on the `Persons` resource?

dataservice/api/person/person.py
```diff
@@ -0,0 +1,125 @@
...
+    @person_api.doc(responses={201: 'person updated',
+                               400: 'invalid data',
+                               404: 'person not found'})
+    @person_api.expect(person_model)
+    def put(self, person_id):
+        """
+        Update an existing person
+        """
+        body = request.json
+        person = model.Person.query.filter_by(kf_id=person_id).first_or_404()
```
dankolbman: `Person.query.get_or_404(id)` is equivalent.

Could also use `query.get()` and catch on the `orm.exc.NoResultFound` to abort with a relevant error message.

dataservice/api/person/person.py
znatty22: This resource class pattern didn't make sense to me before but I think now I understand it a little bit better. I'm gonna go with Person and PersonList classes

dataservice/api/person/person.py
znatty22: Looks like get and get_or_404 can only be used with primary keys

tests/test_person_api.py
```diff
@@ -0,0 +1,152 @@
...
+        body = {
+            'external_id': external_id
+        }
+        response = self.client.post(url_for(PERSON_LIST_URL),
+                                    headers=self._api_headers(),
+                                    data=json.dumps(body))
+
+        return response
+
+    def _test_response_content(self, resp, status_code):
```
dankolbman: Good use case for [paramatrized](https://docs.pytest.org/en/latest/example/parametrize.html) tests.

config.py
```diff
@@ -0,0 +1,69 @@
...
+class ProductionConfig(Config):
+    # Should use postgres
+    SQLALCHEMY_DATABASE_URI = os.environ.get("DATABASE_URL") or \
+            "sqlite:///" + os.path.join(basedir, "data.sqlite")
+
+    @classmethod
+    def init_app(cls, app):
+        Config.init_app(app)
+
+        # email errors to the administrators
```
znatty22, line 42: Not sure how emailing errors works, but is this necessary right now?

znatty22 (dismissed):
A lot of this might be OBE, so just going to complete review for now and figure out how to move forward

parimalak (dismissed):

PR #26, opened 2018-01-16
🔧 Refactor code base to be modular for additional entities

znatty22:


dataservice/api/__init__.py
```diff
@@ -1,7 +1,11 @@
+from flask import Blueprint
 from flask_restplus import Api
 from .person import person_api
 
-api = Api(title='Kids First Data Service',
+api_v1 = Blueprint('api', __name__, url_prefix='/api/v1')
```
dankolbman: We'll probably host under `api.kids-first.io`. We can use `/v1` for versioning, but let's drop the `/api` for now.

dataservice/api/common/model.py
```diff
@@ -1,9 +1,9 @@
 from sqlalchemy.ext.declarative import declared_attr
 
 from datetime import datetime
-from . import db
+from dataservice import db
```
dankolbman: why the change from relative to absolute imports? Should probably stay consistent throughout the code base.

dataservice/commands.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+"""Click commands."""
+
+import os
+
+import click
+
+HERE = os.path.abspath(os.path.dirname(__file__))
+PROJECT_ROOT = os.path.join(HERE, os.pardir)
+TEST_PATH = os.path.join(PROJECT_ROOT, 'tests')
```
dankolbman: Are these used anywhere?

dankolbman:
We can also remove flask-script from `requirements.txt` if it's being removed.

dataservice/commands.py
znatty22: Nope, I'll remove them

dataservice/api/common/model.py
znatty22: Ok as just discussed, we're going with absolute imports. Will make the necessary changes

dankolbman (approved):
lgtm

Issue #7, opened 2018-01-10
Create ID service/mechanism to assign kf_id's


baileyckelly:
A standard should be defined for what a Kid's First id looks like and how it is generated.

dankolbman:
We need to discuss the core set of ids and id formats that we assign to entities.
My thoughts are that each entity in our model should at least be assigned an kids-first id for general consumption. Something that indicates that it is the kids first id, the entity type, and the entity identity should be the minimum. For example:

A Person entity with `person_id: KF-PR000001`
An aliquot with `aliquot_id: KF-AL000001`

Another possibility is to namespace things by dataset and assign ids sequentially within. For example:

The first person in the Seidman dataset: `person_id: KF-SD-PR00001`
The twentieth sample in the Maris dataset: `sample_id: KF-MR-SA00001`

This has the upside of having more easily identifiable, comparable, and memorable ids, but would require every dataset to have a code assigned to it on submission.

Another option is to have internal ids in addition to the primary kids first id. These are only used for internal tracking and can be swapped around without users noticing.

allisonheath:
Just for the record, I think every row (aka entity) should have:

`uuid` a uuid4 for the entity
`kf_id` a name spaced kids first ID as described above, I like the format `KF-<two/three letter abbreviation for the entity>-<short but random alphanumeric id>

I like random alphanumeric generation similar to (down right now): https://blog.datacite.org/cool-dois/. Which iirc creates a base64 hash and then does some small tweaks to remove characters like l10O to help with readability. It's implemented via: https://github.com/datacite/cirneco to create the DOI suffix. 

Both the `uuid` and `kf_id` can be used as primary keys within the Kids First space. The nice thing about UUIDs is that they can truly be used universally, are often easily consumed by other libraries/distribution systems. And if we get to the point where we have enough entities it's starting to cause collision issues, it would be easy to move over towards. It's a lightweight ask to prevent future headaches. 

Files might have to be handled as a special case, as depending on the different data types you can many files per Person. Additionally, right now Gen3 is the authoritative auth service so we need to have synchronized IDs. In Gen3 it's a UUID, so could consider either it's `kf_id` or make an explicit secondary id `dos_id` (that's for digital object schema - one of the GA4GH standards that Gen3 is participating in).  

Also, I foresee having a lot of different "external IDs" to map to, I think in general it's better to keep these in a separate service / tables rather than right with the Kids First entity.

dankolbman:
[...]

PR #27, opened 2018-01-16
✨ Kids First Ids

dankolbman:
Addresses #7 by using a Crockford encoded number for `kf_id` and a uuid for `uuid`.

Waiting for #10  to merge

### DIFF
--- a/dataservice/api/common/id_service.py
+++ b/dataservice/api/common/id_service.py
@@ -1,5 +1,24 @@
 import uuid
+import random
+import base32_crockford as b32
 
 
-def assign_id():
+def uuid_generator():
+    """
+    Returns a stringified uuid of 36 characters
+    """
     return str(uuid.uuid4())
+
+
+def kf_id_generator():
+    """
+    Returns a (Crockford)[http://www.crockford.com/wrmg/base32.html] base 32
+    encoded number up to 8 characters left padded with 0
+
+    Ex:
+    '0004PEDE'
+    'D167JSHP'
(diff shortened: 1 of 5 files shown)


dataservice/api/common/id_service.py
```diff
@@ -1,5 +1,24 @@
 import uuid
+import random
+import base32_crockford as b32
```
znatty22, line 3: Add base32-crockford to requirements.txt

znatty22:
One test is failing: tests > person > test_models.py.

The changes below should fix it:
```python
    def test_create_person(self):
        """
        Test creation of person
        """

        self._create_person('Test_Person_0')

        dt = datetime.now()
        self.assertEqual(Person.query.count(), 1)
        new_person = Person.query.first()
        self.assertGreater(dt, new_person.created_at)
        self.assertGreater(dt, new_person.modified_at)
        self.assertEqual(len(new_person.kf_id), 8)
        self.assertIs(type(uuid.UUID(new_person.uuid)), uuid.UUID)
        self.assertEqual(new_person.external_id, "Test_Person_0")
```
Changes:
For some reason this test was passing before, even though the call to datetime.now() was above the call to create person. Probably has to do with timing of execution. Either way, I moved it below the create person method. 

Updated the tests for length of ids

dankolbman:
Timestamps weren't working because they were assigned on model definition rather than object creation.

znatty22 (dismissed):
Looks good!

PR #35, opened 2018-01-17
🐛 Update README

znatty22:
Add pip install -e . line to readme


dankolbman:
`manage` also works

znatty22:
Hmm ok, I could've sworn it didn't work for parimala when she ran one of the flask commands but maybe something else was wrong

dankolbman:
It will work if the module is installed. This brings up a point that we should probably rename `manage.py` or maybe move it into `dataservice/__init__.py` so it doesn't appear that the application is named `manage`.

znatty22:
Ok, so then should we add `pip install -e .` to the README.md?

dankolbman:
Yeah, I think that's better as that's how it's done in the dockerfile and jenkins.

PR #36, opened 2018-01-17
✨ Add CRUD Mixin 

znatty22:
As I was working on the demographic entity, I found myself writing repetitive code. 

Created a CRUD class which just extends the SQLAlchemy db.Model class and adds CRUD convenience methods.

This makes testing models easier and will also allow us to separate db functionality a little bit more from the web/resource layer.

### DIFF
--- a/dataservice/extensions.py
+++ b/dataservice/extensions.py
@@ -2,6 +2,45 @@ from flask_migrate import Migrate
 from flask_sqlalchemy import SQLAlchemy, Model
 
 
-db = SQLAlchemy()
+class CRUDMixin(Model):
+    """
+    Mixin that adds convenience methods for CRUD
+    (create, read, update, delete) operations.
+    """
+
+    @classmethod
+    def create(cls, **kwargs):
+        """Create a new record and save it the database."""
+        instance = cls(**kwargs)
+        return instance.save()
+
+    def update(self, commit=True, **kwargs):
+        """Update specific fields of a record."""
+        for attr, value in kwargs.items():
+            setattr(self, attr, value)
+        return commit and self.save() or self
+
+    @classmethod
+    def find_all(cls, **kwargs):
+        return [instance for instance in cls.query.filter_by(**kwargs)]
+
+    @classmethod
+    def find_one(cls, **kwargs):
+        return cls.query.filter_by(**kwargs).one_or_none() if kwargs else None
+
+    def save(self, commit=True):
+        """Save the record."""
+        db.session.add(self)
+        if commit:
+            db.session.commit()
+        return self
+
+    def delete(self, commit=True):
+        """Remove the record from the database."""
+        db.session.delete(self)
+        return commit and db.session.commit()
+
+
+db = SQLAlchemy(model_class=CRUDMixin)
 
 migrate = Migrate()
--- a/dataservice/extensions.py
+++ b/dataservice/extensions.py
@@ -10,33 +10,47 @@ class CRUDMixin(Model):
 
     @classmethod
     def create(cls, **kwargs):
-        """Create a new record and save it the database."""
+        """
+        Create a new record and save it the database.
+        """
         instance = cls(**kwargs)
         return instance.save()
 
     def update(self, commit=True, **kwargs):
-        """Update specific fields of a record."""
+        """
+        Update specific fields of a record.
+        """
         for attr, value in kwargs.items():
             setattr(self, attr, value)
         return commit and self.save() or self
 
     @classmethod
     def find_all(cls, **kwargs):
+        """
+        Find all records, filter by kwargs
+        """
         return [instance for instance in cls.query.filter_by(**kwargs)]
 
     @classmethod
     def find_one(cls, **kwargs):
+        """
+        Find one record, filter by kwargs
+        """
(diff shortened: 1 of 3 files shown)


dataservice/extensions.py
```diff
@@ -2,6 +2,59 @@
...
+    (create, read, update, delete) operations.
+    """
+
+    @classmethod
+    def create(cls, **kwargs):
+        """
+        Create a new record and save it the database.
+        """
+        instance = cls(**kwargs)
+        return instance.save()
```
dankolbman, line 17: What if your creating this object with a bunch of other objects?
Eg: I create a person with a demographic.
I probably would want to make one single transaction rather than two.

dataservice/extensions.py
```diff
@@ -2,6 +2,59 @@
...
+
+    def update(self, commit=True, **kwargs):
+        """
+        Update specific fields of a record.
+        """
+        for attr, value in kwargs.items():
+            setattr(self, attr, value)
```
dankolbman, line 24: We'll want some logic in here to make sure certain properties aren't illegally modified, ie `date_created`.

I've been thinking of how this could be added by extending [here](https://github.com/noirbizarre/flask-restplus/blob/master/flask_restplus/resource.py#L60)

dataservice/extensions.py
```diff
@@ -2,6 +2,59 @@
...
+        return commit and self.save() or self
+
+    @classmethod
+    def find_all(cls, **kwargs):
+        """
+        Find all records, filter by kwargs
+        """
+        return [instance for instance in cls.query.filter_by(**kwargs)]
```
dankolbman, line 32: Can get rid of the comprehension and use `all()`

```cls.query.filter_by(...).all()```

But probably best to replace this with pagination or `yield_per` if we really need to.

File: tests/person/test_models.py [hunk 4]
```diff
@@ -45,10 +30,10 @@ def test_get_person(self):
         """
         Test retrieving a person
         """
-        person = self._create_person('Test_Person_0')
+        person = Person.create(external_id='Test_Person_0')
         kf_id = person.kf_id
 
-        person = self._get_person(kf_id)
+        person = Person.find_one(kf_id=kf_id)
```
dankolbman, line 36: `get()` would be preferable for performance and readability, but there is an issue where we have multiple primary keys here.

See [get()](http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.get) for why it would be preferable.

dankolbman (commented):
I'm not too sure that abstracting over the query interface is a good idea. I could see it leading to a bunch of gotchas down the road.
I'd say we should opt for verbosity with elementary database operations, for the moment, and only explicitly execute against the db inside of the top level request to maintain atomicity.

The `filter_by` abstraction could be a nice point to branch into a `search` feature that contains logic for what fields are searchable and returns a `Query` object to be executed, rather than a group of materialized objects.

dataservice/extensions.py
znatty22: You can still do that. See tests for examples (in branch demographic-entity: tests > demographic > test_demographic_models)

Here's an example for convenience too:
```python
# Create demographic
        demo_id = 'demo_1'
        data = {
            'external_id': demo_id,
            'race': 'asian',
            'ethnicity': 'not hispanic',
            'gender': 'female',
            'sex': 'F'
        }
        d = Demographic(**data)

        # Create and save person with demographic
        person_id = 'Test subject 0'
        data = {
            'external_id': person_id,
            'demographic': d
        }
        p = Person.create(**data)
```
Is this what you're asking about?

dataservice/extensions.py
znatty22: EDIT - I think we could do something like the following: 

Do as much validation in the web layer (in http methods) using the namespace models/serializers (or later if we use marshmallow, we would use marshmallow schemas). For example if our request model does not include the `date_created` key, then that attribute will not be in kwargs, and will never be illegally modified in the db. Also, if we have common serializers (like you implemented already and used in person entity) then we can enforce this behavior for all our entities. Then presumably kwargs is a set of valid parameters to execute the db operation. 

I guess we need to discuss more..

dataservice/extensions.py
znatty22: Yep, agreed.

znatty22:
In general the CRUD mixin definitely needs to be expanded. This was just a start. I think this is where we can encapsulate some common exception handling across entities (i.e. a POST request's data is passes validation, but the foreign key value does not exist in the db. This is where we can handle those errors in a standard way for all entities.

dataservice/extensions.py
znatty22: One more thing, it looks like the namespace expect decorator can take a `validate` kwarg. So we can check for existence of required fields, etc.

dataservice/extensions.py
dankolbman: Yes, however, this is done in two transactions. It would be best practice to have these done in one atomic operation.
I see that you can flag whether to commit or not in the `save` method, but I'm of the opinion that this commit should be done explicitly at the end of the transaction (the request), at least for simple crud operations (batch is another topic).

PR #38, opened 2018-01-18
🐛 Update/fix Base class in models.py

znatty22:
Change Base to inherit only the model mixins (Base should not inherit db.Model)

If Base inherits db.Model then this is saying that Base is a db.Model/a
sqlalchemy declarative base. sqlalchemy will think that we are trying to
do single table inheritance, which I don't believe we are. We just want
a class representing the common columns across entities.

Even if we were trying to do single table inheritance, then we need to add
additional code to all child classes of Base to instruct sqlalchemy on
how to handle two entity models with the same column names.

See link for more details:
http://docs.sqlalchemy.org/en/latest/orm/extensions/declarative/inheritance.html

### DIFF
--- a/dataservice/api/common/model.py
+++ b/dataservice/api/common/model.py
@@ -32,14 +32,14 @@ class HasFileMixin:
         return db.relationship("File")
 
 
-class Base(db.Model, IDMixin, TimestampMixin):
+class Base(IDMixin, TimestampMixin):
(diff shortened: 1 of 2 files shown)


dankolbman (approved):
lgtm 👍

parimalak (approved):

PR #49, opened 2018-01-19
🔧 Update sqlalchemy config w foreign key constraint for sqlite

znatty22:
SQLite does not  enforce foreign key constraints by default. Need to explicitly tell sqlalchemy to turn on foreign key constraints when using sqlite.

Otherwise, tests to check foreign key constraints will fail (i.e. right now I can add a demographic for a person that does not exist)

See http://docs.sqlalchemy.org/en/latest/dialects/sqlite.html for more
details.

### DIFF
--- a/dataservice/__init__.py
+++ b/dataservice/__init__.py
@@ -53,6 +53,17 @@ def register_extensions(app):
     # SQLAlchemy
     db.init_app(app)
 
+    # If using sqlite, must instruct sqlalchemy to set foreign key constraint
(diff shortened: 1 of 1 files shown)


parimalak (approved):

dankolbman:
Interesting. We should probably work away from sqlite and using postgres even in testing as there are some other features that would be nice to us, eg `UUID` types.

dankolbman (approved):

PR #56, opened 2018-01-19
✨ Status page and rerouting

### LINKED ISSUE #39: Fix "/" path to return 200

dankolbman:
Adds `/status` endpoint to get information about the api health and version.
Removes the `/v1` prefix. Tried alias `/` to `/v1`, but restplus won't allow it. Ultimately, we probably don't want to maintain two different versions at a time like this and instead should opt for routing api versions on a service level.
Also introduces some general tests using py.test `parametrize` which are nice for covering simple checks like response formatting and status codes.

Fixes #39

### DIFF
--- a/dataservice/api/__init__.py
+++ b/dataservice/api/__init__.py
@@ -1,17 +1,48 @@
+import pkg_resources
+import subprocess
 from flask import Blueprint
 from flask_restplus import Api
 from dataservice.api.participant import participant_api
 
-api_v1 = Blueprint('api', __name__, url_prefix='/v1')
+api_v1 = Blueprint('api', __name__, url_prefix='')
 
 api = Api(api_v1,
           title='Kids First Data Service',
           description=open('dataservice/api/README.md').read(),
-          version='0.1',
+          version=_get_version(),
           default='',
           default_label='')
 
 api.add_namespace(participant_api)
+
+
+@api.route("/status")
+class Status(Resource):
+    """
+    Service Status
+    """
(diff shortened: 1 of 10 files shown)


dataservice/api/__init__.py
```diff
@@ -1,19 +1,63 @@
...
 
+
+api.models['VersionFields'] = _version_fields
+api.models['VersionResponse'] = version_response
+
 api.add_namespace(person_api)
 
 
+@api.route("/status")
+class Status(Resource):
```
dankolbman: Might want to move this somewhere else?

dataservice/api/common/serializers.py
```diff
@@ -44,3 +44,20 @@
 base_pagination = base_response.clone('BasePagination', {
     '_links': fields.Nested(_paginate_fields)
 })
+
+_version_fields = _status_fields.clone('VersionFields', {
+    'version': fields.String(
+        example='/persons?page=3',
```
znatty22: Looks like this might be a copy paste error?

dataservice/api/common/serializers.py
```diff
@@ -44,3 +44,20 @@
...
+        description='Location of the next page'),
```
znatty22: Maybe also copy paste error

dankolbman:
Need to update the README too

znatty22:
Looks good 👍

znatty22 (approved):

parimalak (approved):

PR #60, opened 2018-01-20
♻️ Rename Person to Participant in all relevant files

znatty22:


dankolbman:
Should probably go back and redo commit with git mv (I don't think it can be patched?). Otherwise it's going to be a lot of fun rebasing our other branches :p

znatty22:
So I tried git mv, but the git log looks the same. I think git actually automatically did the rename in my first attempt as well so there's no difference. Let me know if if you think there's a better approach?

dankolbman (approved):
Still rebases fine, and we don't have anything to rebase that makes changes on the participant/person, so should be ok..

I think the proper way to do it is to commit the move and modifications in separate commits, but it's functionally equivalent.

Issue #43, opened 2018-01-19
Create MVP ERD


baileyckelly:
Note: This is due EOD Thursday 1/25/18

We need to create an ERD for OICR of the MVP entities

* Participant
    * Rename Person to Participant
* Sample
* File
* Demographic
* Diagnosis
* ~Participant Relationships 
    * (Needs spec’d out)
* ~Phenotype 
    * Just HPO ID
    * Age at observation (Optional)
* ~Dataset
    * Needs spec’d out
* ~Outcome/Encounter/Event
    * Needs to be spec’d out

allisonheath:
Initial pass:
[DataModelMVP-20180125.pdf](https://github.com/kids-first/kf-api-dataservice/files/1664754/DataModelMVP-20180125.pdf)
[DataModelMVP-20180125.xml.zip](https://github.com/kids-first/kf-api-dataservice/files/1664755/DataModelMVP-20180125.xml.zip)

For now, left off types, but in general should be straightforward. The idea is to first get current data in hand into as-is (no enums for example) and that will give a baseline of where to focus harmonization efforts.

For sure we'll be iterating on this, and ultimately #40 will have all of the details, but this is to give an initial goal to works towards. Feedback welcome! 😄

allisonheath:
Also to further expand on the harmonization and the large arrows to the various ontologies and where we're generally headed at the moment:
 
For example, for phenotype right now we're just going to put the phenotypes we're directly getting to the investigators into the `phenotype` field as a string (e.g. `WEBBED_NECK`). Then we'll work on the analysis on how to harmonize those to HPO terms and likely create a `HPO_ID` field where that ID will be stored (e.g. [HP:0000465](http://compbio.charite.de/hpoweb/showterm?id=HP:0000465)) . The dataservice then can use that ID to materialize the standard term name out via the API, along with other features as needed such as synonyms, lay terms, hierarchies, xrefs as needed. 

Similar concept for diagnosis and MONDO, anatomic site and UBERON, etc.

allisonheath:
From conversations today, the only immediate change is adding `tumor_descriptor` to the sample entity. There were some suggestions on how to improve the family structure, but need to understand some more use cases to determine best path forward. Same for the study/dataset structure. Will open other tickets for those two. I think we can call this closed for the initial pass.

PR #61, opened 2018-01-22
✨ Add ERD generator

dankolbman:
Addresses #43 . Generates an erd from the current definition of the schema.


dev-requirements.txt
```diff
@@ -2,3 +2,4 @@ pytest==3.3.1
 pytest-cache==1.0
 pytest-cov==2.5.1
 pytest-pep8==1.0.6
+-e git+git@github.com:dankolbman/eralchemy.git@1058c12b828013144f861cb59f0e2c280a1a5ef6#egg=eralchemy
```
dankolbman: Waiting for [this](https://github.com/Alexis-benoist/eralchemy/pull/42) to merge.

znatty22 (changes requested):
This is pretty cool 👍,  Only comment is that it looks like File is getting rendered because its in dataservice.api.common.models. Maybe we should remove that since it'll go into its own package later

dankolbman:
Technically accurate to have File in there. As to whether that ever should have been commited is another question...

znatty22 (approved):

dev-requirements.txt
```diff
@@ -2,3 +2,4 @@ pytest==3.3.1
 pytest-cache==1.0
 pytest-cov==2.5.1
 pytest-pep8==1.0.6
+-e git+git@github.com:Alexis-benoist/eralchemy.git#egg=eralchemy
```
dankolbman, line 5: Will want to change this to track the pypy package once a release has been cut.

parimalak (approved):

PR #62, opened 2018-01-22
✨ Demographic resource and tests

znatty22:
Add a crud api for the demographic entity (get, post, put, delete methods) 
Add testing crud api

### DIFF
--- a/dataservice/api/__init__.py
+++ b/dataservice/api/__init__.py
@@ -1,6 +1,7 @@
 from flask import Blueprint
 from dataservice.api.status import StatusAPI
 from dataservice.api.participant import ParticipantAPI
+from dataservice.api.demographic import DemographicAPI
 
 
 def register_crud_resource(app, view, endpoint, url,
@@ -31,11 +32,16 @@ def register_crud_resource(app, view, endpoint, url,
     app.add_url_rule('{}<{}:{}>'.format(url, pk_type, pk), view_func=view_func,
                      methods=['GET', 'PUT', 'DELETE'])
 
+
 api = Blueprint('api', __name__, url_prefix='')
 
 # Status resource
 status_view = StatusAPI.as_view('status')
 api.add_url_rule('/', view_func=status_view, methods=['GET'])
 api.add_url_rule('/status', view_func=status_view, methods=['GET'])
+
 # Participant resource
 register_crud_resource(api, ParticipantAPI, 'participants', '/participants/')
+
+# Demographic resource
+register_crud_resource(api, DemographicAPI, 'demographics', '/demographics/')
--- a/dataservice/api/demographic/resources.py
+++ b/dataservice/api/demographic/resources.py
@@ -1,0 +1,133 @@
+from flask import (
+    abort,
+    request
+)
+from flask.views import MethodView
+from sqlalchemy.exc import IntegrityError
+from sqlalchemy.orm.exc import NoResultFound
+from marshmallow import ValidationError
+
+from dataservice.extensions import db
+from dataservice.api.errors import handle_integrity_error
+from dataservice.api.demographic.models import Demographic
+from dataservice.api.demographic.schemas import DemographicSchema
+
+
+class DemographicAPI(MethodView):
+    """
+    Demographic REST API
+    """
+
+    def get(self, kf_id):
+        """
+        Get a demographic by id or get all demographics
+
+        Get a demographic by Kids First id or get all demographics if
+        Kids First id is None
+        """
+
+        # Get all
+        if kf_id is None:
+            d = Demographic.query.all()
+            return DemographicSchema(many=True).jsonify(d)
+        # Get one
+        else:
+            try:
+                d = Demographic.query.filter_by(kf_id=kf_id).one()
+            # Not found in database
+            except NoResultFound:
+                abort(404, 'could not find {} `{}`'
+                      .format('demographic', kf_id))
+            return DemographicSchema().jsonify(d)
+
+    def post(self):
+        """
+        Create a new demographic
+        """
+
+        body = request.json
+
+        # Deserialize
+        try:
+            d = DemographicSchema(strict=True).load(body).data
+        # Request body not valid
+        except ValidationError as e:
+            abort(400, 'could not create demographic: {}'.format(e.messages))
+
+        # Add to and save in database
+        try:
+            db.session.add(d)
+            db.session.commit()
+        # Database error
+        except IntegrityError as e:
+            db.session.rollback()
+            context = {'method': 'create', 'entity': 'demographic',
+                       'ref_entity': 'participant', 'exception': e}
+            abort(400, handle_integrity_error(**context))
+
+        return DemographicSchema(201, 'demographic {} created'
+                                 .format(d.kf_id)).jsonify(d), 201
+
+    def put(self, kf_id):
+        """
+        Update existing demographic
+
+        Update an existing demographic given a Kids First id
+        """
+
+        body = request.json
+
+        # Check if demographic exists
+        try:
+            d1 = Demographic.query.filter_by(kf_id=kf_id).one()
+        # Not found in database
+        except NoResultFound:
+            abort(404, 'could not find {} `{}`'.format('demographic', kf_id))
+
+        # Validation only
+        try:
+            d = DemographicSchema(strict=True).load(body).data
+        # Request body not valid
+        except ValidationError as e:
+            abort(400, 'could not update demographic: {}'.format(e.messages))
+
+        # Deserialize
+        d1.external_id = body.get('external_id')
+        d1.race = body.get('race')
+        d1.gender = body.get('gender')
+        d1.ethnicity = body.get('ethnicity')
+        d1.participant_id = body.get('participant_id')
+
+        # Save to database
+        try:
+            db.session.commit()
+        # Database error
+        except IntegrityError as e:
+            db.session.rollback()
+            context = {'method': 'update', 'entity': 'demographic',
+                       'ref_entity': 'participant', 'exception': e}
+            abort(400, handle_integrity_error(**context))
+
+        return DemographicSchema(200, 'demographic {} updated'
+                                 .format(d1.kf_id)).jsonify(d1), 200
+
+    def delete(self, kf_id):
+        """
+        Delete demographic by id
+
+        Deletes a demographic given a Kids First id
+        """
+
+        # Check if demographic exists
+        try:
+            d = Demographic.query.filter_by(kf_id=kf_id).one()
+        # Not found in database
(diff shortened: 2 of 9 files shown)


dataservice/api/common/error_handler.py
```diff
@@ -0,0 +1,62 @@
+"""
+API level error handling/message formatting
+
+Includes helpers to parse the generic database integrity error
```
znatty22: There doesn't seem to be a better way to distinguish specific types of integrity errors. If someone else knows of anything better please let me know.

dataservice/api/demographic/resources.py
```diff
@@ -0,0 +1,150 @@
...
+        try:
+            db.session.add(demographic)
+            db.session.commit()
+        # Catch error and rollback db
+        except IntegrityError as e:
+            db.session.rollback()
+            context = {'method': 'create', 'entity': 'demographic',
+                       'ref_entity': 'participant', 'exception': e,
+                       'payload': body}
+            # raise BadInputError(context=context)
```
znatty22: Left this comment in here to show that error handling could also be handled with custom exceptions. With this approach, we would throw a custom exception and then handle it at the API level using the api.errorhandler decorator. Decided to avoid this approach for now due to some related known bugs with flask_restplus errorhandling. See  https://github.com/noirbizarre/flask-restplus/issues

dataservice/api/__init__.py
```diff
@@ -12,6 +15,25 @@
...
+def default_error_handler(error):
+    """
+    Default error handler
+    """
+    code = getattr(error, 'code', 500)
+    return kf_response(code=code, message=str(error))
+
+
+# TODO
+"""
```
znatty22: As a work around we may need to change our kf_response format a little bit to match the default restplus format. So instead of having 
```
    {
      "_links": {
        ...
      },
      "_status": {
        "code": 200,
        "message": "Success"
      },
      "results": [
        ...
      ]
    }
    ```
We might need to change to something like: 
```
    {
      "_links": {
        ...
      },
      "code": 200,
      "message": "Success"
      "results": [
        ...
      ]
    }
    ```

dataservice/api/demographic/resources.py
```diff
@@ -0,0 +1,150 @@
...
+        demographic = models.Demographic(**body)
+
+        # Try saving to db
+        try:
+            db.session.add(demographic)
+            db.session.commit()
+        # Catch error and rollback db
+        except IntegrityError as e:
+            db.session.rollback()
+            context = {'method': 'create', 'entity': 'demographic',
```
znatty22: Might be worth considering the crud mixin or something like it :). Reason being, if we had something to encapsulate our crud methods we could handle the db errors in that class and then avoid passing context parameters to another module (error_handler). The crud class would still call out to the error_handler to determine the type of error but it could take care of forming the actual error message with the context (crud method used, the entity, the reference entity, etc). Would also avoid duplicate code in every entity resource class

dataservice/api/__init__.py
dankolbman: Yeah, `kf_response` is a little restricted in what it actually applies to. I'm trying to move it down into the `Resource` but it may have to go deeper. For now, extending the Api with a modified [`error_handler`](https://github.com/noirbizarre/flask-restplus/blob/3f478d220b54ddb9d553a351e1b4327f601d4a01/flask_restplus/api.py#L559) may be easiest.

dataservice/api/common/error_handler.py
```diff
@@ -0,0 +1,62 @@
...
+
+UNIQUE_CONSTRAINT_ERROR_PREFIX = 'unique constraint failed'
+UNIQUE_CONSTRAINT_ERROR_MSG_TEMPLATE = \
+    'Cannot {0} {1}, {2} already has a {1}'
+
+INVALID_INPUT_MSG_TEMPLATE = 'Client error: invalid {} provided'
+NOT_FOUND_MSG_TEMPLATE = 'Client error: {} not found'
+
+
+def handle_invalid_input(context=None):
```
dankolbman: What is this function trying to accomplish? It looks like it's for formatting the error response from SQLAlchemy? In that case, does this really need to happen, and if so, could it happen by extending the SQLAlchemy exceptions?

dankolbman: Also, maybe the context should be kwargs? That way there's some sort of 'type' checking on the handle.

dataservice/api/demographic/resources.py
dankolbman: I think that exception handling from werkzeug and flask already encapsulate these things pretty well. It seems like the functionality here could be achieved by coercing to the standard error handling format, or by extending it as needed.

dataservice/api/demographic/resources.py
```diff
@@ -0,0 +1,150 @@
...
+    def get(self, kf_id):
+        """
+        Get a demographic by id
+        Gets a demographic given a Kids First id
+        """
+        demographic = models.Demographic.query.filter_by(
+            kf_id=kf_id).one_or_none()
+
+        if not demographic:
+            return self._not_found(kf_id)
```
dankolbman: This should probably be refactored in the future. It seems the main reason we're doing it this way is to make our responses fit our desired format. If we can figure out how to extend the handling to use our response, this line could be as easy as `raise HTTPException`

dataservice/api/demographic/resources.py
```diff
@@ -0,0 +1,150 @@
...
+        Update an existing demographic
+        """
+        demographic = models.Demographic.query.filter_by(kf_id=kf_id
+                                                         ).one_or_none()
+        body = request.json
+
+        if not demographic:
+            return self._not_found(kf_id)
+
+        demographic.external_id = body.get('external_id')
```
dankolbman: I've been thinking of moving the parsing functionality here to a `from_json` or `from_dict` on the ORM class. It Thoughts?

dataservice/api/demographic/serializers.py
```diff
@@ -0,0 +1,51 @@
...
+    'ethnicity': fields.String(
+        example='hispanic or latino',
+        description='The ethnicity of the person this demographic belongs to'),
+    'gender': fields.String(
+        example='female',
+        description='The gender of the person this demographic belongs to'),
+    'sex': fields.String(
+        example='F',
+        description='The sex of the person this demographic belongs to'),
+    'person_id': fields.String(
```
dankolbman: `participant_id`?

tests/demographic/test_demographic_models.py
```diff
@@ -0,0 +1,224 @@
...
+        }
+        dt = datetime.now()
+        d = Demographic(**data)
+        db.session.add(d)
+        db.session.commit()
+
+        # Check that demographic was created correctly
+        self.assertEqual(Demographic.query.count(), 1)
+        new_demographic = Demographic.query.first()
+        self.assertGreater(new_demographic.created_at, dt)
```
dankolbman: We could probably draw out all these created_at and modified_at checks to a parametrized test.

dankolbman (commented):
I think a lot of the abstractions your suggesting could be taken care of by extending the response handling in flask and flask-restplus to fit our needs. This approach seems more natural to me as it would still use the same familiar interface of a flask application rather than a high level library that ultimately deals with low level functionality.

I've been looking at how the resource responses could be handled [here](https://github.com/noirbizarre/flask-restplus/blob/3f478d220b54ddb9d553a351e1b4327f601d4a01/flask_restplus/resource.py) but there's some nuances with restplus' marshaling that make it difficult. It may be better to look more deeply into flask for how that may be achieved.

dataservice/api/demographic/resources.py
znatty22: Yep, sounds like a good idea

dataservice/api/demographic/serializers.py
znatty22: Will update this PR now that #60 has been merged

znatty22:
Thanks for the comments/suggestions. Let's discuss more on error handling and response formatting

znatty22:
After discussing, we agreed that flask_restplus has too many issues with error handling and response marshaling. We've decided to go with regular flask instead of flask_restplus which will solve the error handling issues. And we are going with marshmallow to solve the request/response validation/formatting. See #63 for details.

dataservice/api/demographic/resources.py
```diff
@@ -0,0 +1,115 @@
...
+        Update existing demographic
+
+        Update an existing demographic given a Kids First id
+        """
+        body = request.json
+        try:
+            # Check if demographic exists
+            d1 = Demographic.query.filter_by(kf_id=kf_id).one()
+            # Deserialize
+            d1.external_id = body.get('external_id')
```
znatty22: This can be done more dynamically/better in the future. Tried using sqlalchemy session.merge with the loaded obj from ma schema but this achieves patch/partial update behavior. To do a full replace, would have to iterate over the model's columns and relationships and set them back to defaults. This isn't trivial so will look into this in the near future.

dataservice/api/demographic/resources.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        Update existing demographic
+
+        Update an existing demographic given a Kids First id
+        """
+        body = request.json
+        try:
+            # Check if demographic exists
+            d1 = Demographic.query.filter_by(kf_id=kf_id).one()
+            # Deserialize
```
znatty22: This can be done more dynamically/better in the future. Tried using sqlalchemy session.merge with the loaded obj from marshmallow but this achieves patch/partial update behavior. To do a full replace, would first have to iterate over the d1's columns and relationships and set them back to defaults and then perform merge with the deserialized object from marshmallow. This isn't trivial so will look into this in the near future.

tests/demographic/test_demographic_resources.py
```diff
@@ -0,0 +1,357 @@
...
+                                    data=json.dumps(d2))
+        # Check status code
+        self.assertEqual(response.status_code, 400)
+        # Check response body
+        response = json.loads(response.data.decode("utf-8"))
+        # Check error message
+        message = 'Cannot create demographic, participant already'
+        self.assertIn(message, response['_status']['message'])
+
+        # Check database
```
dankolbman, line 133: No database checking?

dankolbman (commented):
It looks like a lot of the test functions could easily be replaced with 1-2 lines added to the parametrized tests [here](https://github.com/kids-first/kf-api-dataservice/blob/master/tests/test_api.py#L14-L59)

tests/demographic/test_demographic_resources.py
znatty22: Good catch, will add 👍

parimalak (approved):
👍

dankolbman (approved):
looks good!

Issue #63, opened 2018-01-22
Re-write serialization layer in Marshmallow


dankolbman:
Flask-RESTPlus gets in the way with regards to error handling and doesn't provide very powerful request parsing. Marshmallow is a popular (de)serialization library that supports inheriting schemas from sqlalchemy models. There is also apispec which will generate much of the swagger documentation using marshmallow schemas.
Replacing Flask-RESTPlus should be straightforward in terms of functionality and will give us much more power in parsing requests and responses as well as reduce the number of hacky work-arounds needed.

PR #64, opened 2018-01-22
✨ Add Demographic db model

znatty22:
Adds the sqlalchemy demographic model and relevant tests 

Note that this was part of #62, however that PR is currently blocked by #63. Thus, to keep moving forward with development this PR covers the necessary persistence functionality which does not depend on #63.


dankolbman (approved):
lgtm 👍

dankolbman:
Could you clean up the commit history before merging?

PR #65, opened 2018-01-23
💥 Nuke and pave™

dankolbman:
Uses native Flask `MethodViews` for restful routes and registers the routes using flask's url rules. Replaces RestPlus' models with Marshmallow's schemas, which can inherit straight from the SQLAlchemy models.

Still need to address how documentation will be generated.

Addresses #63 

- [x] Set up routing
- [x] Add serializers
- [x] Participant resource
- [x] Status resource
- [ ] Swagger docs
- [ ] Refactor


dankolbman:
Tests passing 🎉

dataservice/api/__init__.py
```diff
@@ -1,37 +1,41 @@
...
+                     methods=['GET', 'PUT', 'DELETE'])
+
+api = Blueprint('api', __name__, url_prefix='')
+
+# Status resource
+status_view = StatusAPI.as_view('status')
+api.add_url_rule('/', view_func=status_view, methods=['GET'])
+api.add_url_rule('/status', view_func=status_view, methods=['GET'])
+# Participant resource
+register_crud_resource(api, ParticipantAPI, 'participants', '/participants/')
```
dankolbman, line 41: Might want to move all these rule definitions somewhere else.

dataservice/api/common/schemas.py
```diff
@@ -0,0 +1,49 @@
...
+        else:
+            _links = {}
+
+        return {'results': data,
+                '_links': _links,
+                '_status': {'message': self.status_message,
+                            'code': self.status_code}}
+
+
+class ErrorSchema(Schema):
```
dankolbman, line 30: Should prolly make this the base status schema and figure out how to inherit it in the model schemas.

dataservice/__init__.py
```diff
@@ -72,9 +72,12 @@ def register_error_handlers(app):
     """
     Register error handlers
     """
-    pass
+    from dataservice.api import errors
+    from werkzeug.exceptions import HTTPException
+    app.register_error_handler(HTTPException, errors.http_error)
```
dankolbman, line 79: This doesn't actually catch anything. Was supposed to be fixed recently in Flask, but couldn't get it to work. Might want to investigate more.

dataservice/__init__.py
```diff
@@ -32,8 +32,7 @@ def register_shellcontext(app):
 
     def shell_context():
         """Shell context objects."""
-        return {'db': db,
-                'Participant': Participant}
+        return {'db': db}
```
znatty22: Why remove Participant?

dataservice/__init__.py
znatty22: Hmm interesting.. Well I like this idea of having a catch all error handler for anything we don't explicitly catch. 

I found this [post](https://stackoverflow.com/questions/29332056/global-error-handler-for-any-exception) that provides some options but maybe you already saw it?

dataservice/api/__init__.py
```diff
@@ -1,37 +1,41 @@
...
 
-api_v1 = Blueprint('api', __name__, url_prefix='')
 
-api = Api(api_v1,
-          title='Kids First Data Service',
-          description=open('dataservice/api/README.md').read(),
-          version=_get_version(),
-          default='',
-          default_label='')
+def register_crud_resource(app, view, endpoint, url,
```
znatty22, line 6: Nice, I like it. I agree we could probably move this chunk of code (method and registering of routes) somewhere later, maybe a view factory or something. But for now it's probably fine.

dataservice/api/participant/resources.py
```diff
@@ -1,114 +1,74 @@
...
+        Gets a participant given a Kids First id
         """
-        participants = models.Participant.query.all()
-        return kf_response(participants, 200, '{} participants'.
-                           format(len(participants)))
+        if kf_id is None:
+            return (ParticipantSchema(many=True)
+                    .jsonify(Participant.query.all()))
+        else:
+            try:
```
znatty22, line 21: Why not use one_or_none here and abort if None? In general, I think its best to avoid the catch all type try/excepts (even though there is only one line of code in the try block). Plus I think the code would look a little cleaner

dataservice/api/participant/resources.py
```diff
@@ -1,114 +1,74 @@
...
     def post(self):
         """
         Create a new participant
-
-        Creates a new participant and assigns a Kids First id
         """
         body = request.json
-        participant = models.Participant(**body)
-        db.session.add(participant)
+        p = Participant(external_id=body.get('external_id'))
```
znatty22, line 33: Isn't this where you would use the marshmallow schema to load from dict to your model object?

dataservice/api/participant/resources.py
```diff
@@ -1,114 +1,74 @@
...
-                                    404: 'participant not found'})
-    @participant_api.expect(participant_fields)
     def put(self, kf_id):
         """
         Update an existing participant
         """
         body = request.json
-        participant = models.Participant.query.\
-            filter_by(kf_id=kf_id).one_or_none()
+        try:
```
znatty22, line 46: Same comment as above

dataservice/api/common/schemas.py
```diff
@@ -0,0 +1,49 @@
...
+
+class BaseSchema(ma.ModelSchema):
+
+    def __init__(self, code=200, message='success', *args, **kwargs):
+        self.status_code = code
+        self.status_message = message
+        super(BaseSchema, self).__init__(*args, **kwargs)
+
+    class Meta:
+        fields = ('kf_id',)
```
znatty22: What about the other fields we had before like created_at and modified_at?

znatty22 (commented):
See file comments

dataservice/__init__.py
dankolbman: Yeah, I tried catching on `Exception` too, but something was getting in the way of the handler. Will need to investigate more.

dataservice/api/participant/resources.py
dankolbman: Yeah, better not to do general try/catch all. I've added the actual exception type to the exception. As for `one_or_none`, I'd really like to handle errors explicitly, in this case

Example:
```
def get_toppings(cone_id):
    icecream_cone = Cone.query.filter_by(cone_id=cone_id).one_or_none()
    if icecream_cone is None:
        abort('That cone hasnt been ordered yet!')
    if icecream_cone.toppings is None:
        return 'no fun'
    return icecream_cone.toppings
```

Compare with:
```
def get_toppings(cone_id):
    try:
        icecream_cone = Cone.query.filter_by(cone_id=cone_id).one()
    except NoResultFound:
        abort('That cone hasnt been ordered yet!')
    if icecream_cone.toppings is None:
        return 'no fun'
    return icecream_cone.toppings
```

I like the second because
1) it's easy to tell at a glance when you're handling an error vs implementing functionality
2) Exceptions are nice. We could have a general exception handler for `NoResultFound` and not even bother with the try/catch block here

dataservice/api/participant/resources.py
dankolbman: Yup, getting to it :)

dataservice/api/participant/resources.py
znatty22: Cool, I like it 👍

znatty22 (approved):
Good enough to merge and then we can make updates as necessary 👍

parimalak (approved):

PR #66, opened 2018-01-23
✨ Sample db model

parimalak:


dataservice/api/participant/models.py
```diff
@@ -14,3 +14,5 @@ class Participant(db.Model, Base):
     """
     __tablename__ = "participant"
     external_id = db.Column(db.String(32))
+    sample = db.relationship('Sample', backref='participant', uselist=False,
```
znatty22: I think you want to use `uselist=True` here right? Since Participant --> Sample is a one to many relationship

znatty22: And then if that is true, change the variable name `sample` to `samples`

dataservice/api/sample/models.py
```diff
@@ -0,0 +1,19 @@
...
+    :param kf_id: Unique id given by the Kid's First DCC
+    :param external_id: Name given to sample by contributor
+    :param composition : composition of the sample
+    :param tissue_type: Either Normal or Tumor
+    """
+    __tablename__ = "sample"
+    external_id = db.Column(db.String(32))
+    tissue_type = db.Column(db.String(32))
+    composition = db.Column(db.String(32))
+    participant_id = db.Column(db.Integer, db.ForeignKey('participant.kf_id'),
```
dankolbman: should this refer to `participant._id`?

tests/sample/test_sample_models.py
```diff
@@ -0,0 +1,149 @@
...
+import uuid
+
+from dataservice.extensions import db
+from dataservice.api.participant.models import Participant
+from dataservice.api.sample.models import Sample
+from tests.utils import FlaskTestCase
+
+from sqlalchemy.exc import IntegrityError
+
+class ModelTest(FlaskTestCase):
```
znatty22, line 11: If the relationship between Participant and Sample is actually one to many then these tests will need to account for that. Should test adding multiple samples to a Participant, updating a single sample in the list of Participant's samples, deleting one sample from the list of Participant's samples, etc

dataservice/api/sample/models.py
```diff
@@ -0,0 +1,19 @@
...
+    Sample entity.
+    :param _id: Unique id assigned by RDBMS
+    :param kf_id: Unique id given by the Kid's First DCC
+    :param external_id: Name given to sample by contributor
+    :param composition : composition of the sample
+    :param tissue_type: Either Normal or Tumor
+    """
+    __tablename__ = "sample"
+    external_id = db.Column(db.String(32))
+    tissue_type = db.Column(db.String(32))
```
dankolbman: Maybe these should be longer? Not sure on the requirements and what kinds of headaches it could cause if we get something longer than 32 chars some time.

tests/sample/test_sample_models.py
```diff
@@ -0,0 +1,149 @@
...
+        s = Sample(external_id='Test_Sample_0',tissue_type='Normal',composition='Test_comp_0')
+
+        #Add Sample to db
+        self.assertRaises(IntegrityError, db.session.add(s))
+
+    def test_foreign_key_constraint(self):
+
+    
+        """
+        Test sample cannot be created with out required parameters such as participant_id
```
znatty22: Maybe change the docstring to `Test sample cannot be created with an existing/valid Participant`

znatty22 (changes requested):
See comments on files

tests/sample/test_sample_models.py
```diff
@@ -0,0 +1,149 @@
...
+    def create_participant_sample(self):
+        participant_id ="Test_Subject_0"
+        sample_id ="Test_Sample_0"
+        data={
+        'external_id': sample_id,
+        'tissue_type': 'Normal',
+        'composition': 'Test_comp_0',
+        }
+        sample_0 = Sample(**data)
+        participant_0 = Participant(external_id = participant_id,sample =sample_0)
```
dankolbman: Style nit: no spaces around kwarg assignments, spaces after the comma in arguments.

tests/sample/test_sample_models.py
```diff
@@ -0,0 +1,149 @@
...
+        db.session.delete(p)
+        db.session.commit()
+
+        s= Sample.query.filter_by(external_id=sample_id).one_or_none()
+        self.assertIs(s,None)
+ 
+        self.assertIsNone(Participant.query.filter_by(external_id=sample_id).one_or_none())
+
+    def test_not_null_constraint(self):
+
```
dankolbman: Shoudn't have any whitespace before the docstring

dankolbman (changes requested):

dataservice/api/participant/models.py
parimalak: yep one to many relationship between Participant and Sample.

dataservice/api/sample/models.py
znatty22: The kf-dictionary tissue_type has values like "Normal", "Abnormal" etc. So maybe its ok? But @allisonheath would know best.

https://github.com/kids-first/kf-dictionary/blob/514fee43d816e98e6bfae2427c4207c53785710f/gdcdictionary/schemas/sample.yaml#L483

allisonheath: In general with postgres there's no longer any reason to limit the character fields unless you want to for QC purposes: https://stackoverflow.com/questions/4848964/postgresql-difference-between-text-and-varchar-character-varying

So I'd suggest using the `Text` type here in the spirit of not backing ourselves into a corner with variable data.

dataservice/api/participant/models.py
```diff
@@ -14,3 +14,5 @@ class Participant(db.Model, Base):
     """
     __tablename__ = "participant"
     external_id = db.Column(db.String(32))
+    samples = db.relationship('Sample', backref='participant',
```
znatty22: Sorry I didn't think of this before, but I think you may want to add the `lazy="joined"` parameter to db.relationship here. I did this for Diagnosis and I think we will want the same for Sample. See explanation [here](https://github.com/kids-first/kf-api-dataservice/pull/67#discussion_r163430583)

tests/sample/test_sample_models.py
```diff
@@ -0,0 +1,205 @@
...
+from dataservice.extensions import db
+from dataservice.api.participant.models import Participant
+from dataservice.api.sample.models import Sample
+from tests.utils import FlaskTestCase
+
+from sqlalchemy.exc import IntegrityError
+
+class ModelTest(FlaskTestCase):
+    """
```
znatty22, line 12: I know this is nitpicking, but could you add docstrings to all of the methods 😬

znatty22 (changes requested):
See file comments

tests/sample/test_sample_models.py
parimalak: 👍

znatty22 (approved):

tests/sample/test_sample_models.py
```diff
@@ -0,0 +1,229 @@
...
+                    composition='Test_comp_1',participant_id =p.kf_id)
+        
+        db.session.add(s)
+        db.session.commit()
+
+        #Delete Sample
+        db.session.delete(s)
+        db.session.commit()
+        self.assertEqual(Sample.query.count(), 1)
+
```
dankolbman, line 229: Trailing line at EOF

dankolbman (approved):
LGTM 💯 
Could you get rid of the merge commits before merging to master?

Issue #69, opened 2018-01-24
Use Marshmallow schemas to deserialize objects


dankolbman:
Objects are still being created by manually parsing request bodies. We should move to using marshmallow's `load` to do input validation and construction of orm objects.

PR #72, opened 2018-01-24
✨ Deserialize objects

dankolbman:
Validate requests to create objects using Marshmallow's validation and deserialization. Now we get pretty feedback when the user does something stupid:

```
{
  "_status":{
    "code":400,
    "message":"could not create participant: {'blah': ['Unknown field']}"
  }
}
```

Addresses #69

### DIFF
--- a/dataservice/api/common/schemas.py
+++ b/dataservice/api/common/schemas.py
@@ -1,5 +1,5 @@
 from dataservice.extensions import ma
-from marshmallow import post_dump
+from marshmallow import post_dump, post_load, validates_schema, ValidationError
 from flask_marshmallow import Schema
 
 
@@ -12,6 +12,7 @@ class BaseSchema(ma.ModelSchema):
 
     class Meta:
         exclude = ('_id', 'uuid')
+        dump_only = ('created_at', 'modified_at')
 
     @post_dump(pass_many=True)
     def wrap_envelope(self, data, many):
@@ -25,6 +26,12 @@ class BaseSchema(ma.ModelSchema):
                 '_links': _links,
                 '_status': {'message': self.status_message,
                             'code': self.status_code}}
+
+    @validates_schema(pass_original=True)
+    def check_unknown_fields(self, data, original_data):
+        unknown = set(original_data) - set(self.fields)
+        if unknown:
+            raise ValidationError('Unknown field', unknown)
 
 
 class ErrorSchema(Schema):
--- a/dataservice/api/common/schemas.py
+++ b/dataservice/api/common/schemas.py
@@ -1,5 +1,5 @@
 from dataservice.extensions import ma
-from marshmallow import post_dump, post_load, validates_schema, ValidationError
+from marshmallow import post_dump, validates_schema, ValidationError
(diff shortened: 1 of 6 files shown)


dataservice/api/common/schemas.py
```diff
@@ -12,6 +12,7 @@ def __init__(self, code=200, message='success', *args, **kwargs):
 
     class Meta:
         exclude = ('_id', 'uuid')
+        dump_only = ('created_at', 'modified_at')
 
     @post_dump(pass_many=True)
     def wrap_envelope(self, data, many):
```
znatty22, line 18: Just curious, won't the many flag tell you if data is a list or not?

dataservice/api/participant/schemas.py
```diff
@@ -1,13 +1,14 @@
...
+from marshmallow_sqlalchemy import field_for
 
 
 class ParticipantSchema(BaseSchema):
+
     class Meta(BaseSchema.Meta):
         model = Participant
-
-    date_created = ma.DateTime()
+        dump_only = ('created_at', 'modified_at')
```
znatty22: Do you need this since the same dump_only fields are inherited from BaseSchema.Meta?

dataservice/api/participant/schemas.py
```diff
@@ -1,13 +1,14 @@
 from dataservice.api.participant.models import Participant
 from dataservice.api.common.schemas import BaseSchema
 from dataservice.extensions import ma
+from marshmallow_sqlalchemy import field_for
```
znatty22: Is this used anywhere?

dataservice/api/common/schemas.py
dankolbman: The `pass_many` in the decorator makes it pass the raw data and sets the `many` arg to `True`. Otherwise, it will pass each object independently, and I don't want that because I want all entities to be wrapped in the envelope.

znatty22: Maybe I'm misunderstanding `pass_many`? But I thought `pass_many` says that the arg `data` could be a single item or many, and if it is many, then the `many` flag will be set to true. Then you would just check `if not many:` instead of `if type(data) is not list:`. I was playing around with this before and it seemed to work

dataservice/api/common/schemas.py
```diff
@@ -1,5 +1,5 @@
 from dataservice.extensions import ma
-from marshmallow import post_dump
+from marshmallow import post_dump, post_load, validates_schema, ValidationError
```
znatty22: I don't think post_load is used anywhere here

dataservice/api/common/schemas.py
dankolbman: Yeah, that could be done too, just need to make sure it's always specified as `many` when it's being dumped.

dankolbman: Updated to check many and also check for presence of `_links`

znatty22 (approved):
👍

parimalak (approved):

PR #74, opened 2018-01-26
✨ Aliquot model with Tests

parimalak:
Added Aliquot model with tests.


dataservice/api/aliquot/models.py
```diff
@@ -25,7 +25,8 @@ class Aliquot(db.Model, Base):
     analyte_type = db.Column(db.Text(), nullable=False)
     concentration = db.Column(db.Integer())
     volume = db.Column(db.Float())
-    # '2017-06-01'
     shipment_date = db.Column(db.DateTime())
+    samples = db.relationship('Sample', back_populates='aliquots',
```
parimalak: switched to back_populates to delete parent(Sample) when empty child(Aliquot). There by have option to delete orphan reciprocally.

dataservice/api/aliquot/models.py
```diff
@@ -0,0 +1,32 @@
...
+    external_id = db.Column(db.Text())
+    shipment_origin = db.Column(db.Text())
+    shipment_destination = db.Column(db.Text())
+    analyte_type = db.Column(db.Text(), nullable=False)
+    concentration = db.Column(db.Integer())
+    volume = db.Column(db.Float())
+    shipment_date = db.Column(db.DateTime())
+    samples = db.relationship('Sample', back_populates='aliquots',
+                              cascade="all, delete-orphan", single_parent=True)
+    sample_id = db.Column(db.Integer, db.ForeignKey('sample.kf_id'),
```
dankolbman: This should be `String(8)` to match with `kf_id`

dataservice/api/aliquot/models.py
```diff
@@ -0,0 +1,32 @@
...
+    :param volume: The volume in microliters (ml) of the aliquots derived from
+     the analyte(s) shipped for sequencing and characterization
+    :param shipment_date: The date item was shipped in YYYY-MM-DD format
+    """
+    __tablename__ = "aliquot"
+    external_id = db.Column(db.Text())
+    shipment_origin = db.Column(db.Text())
+    shipment_destination = db.Column(db.Text())
+    analyte_type = db.Column(db.Text(), nullable=False)
+    concentration = db.Column(db.Integer())
```
znatty22, line 26: @allisonheath  Should this be a float too?

dataservice/api/sample/models.py
```diff
@@ -8,16 +9,23 @@ class Sample(db.Model, Base):
...
     :param external_id: Name given to sample by contributor
-    :param composition : composition of the sample
-    :param tissue_type: Either Normal or Tumor
-    :param anatomical_site : location of the collected sample
+    :param composition : The cellular composition of the sample.
+    :param tissue_type: description of the kind of tissue collected
+           with respect to disease status or proximity to tumor tissue
+    :param anatomical_site : The name of the primary disease site of the
+           submitted tumor sample
     :param age_at_event_days: age of participant when sample was collected
```
znatty22: I think the param comment should be: Age at the time sample was acquired, expressed in number of days since birth

tests/aliquot/test_aliquot_models.py
```diff
@@ -0,0 +1,383 @@
...
+        """
+        create a participant, sample, and aliquot save to db
+        returns participant_id, sample_id, and aliquot_id
+        """
+        dt = datetime.now()
+        participant_id ="Test_Subject_0"
+        sample_id ="Test_Sample_0"
+        aliquot_id ="Test_Aliquot_0"
+        sample_data={
+        'external_id':sample_id,
```
dankolbman: Can we indent the dicts?

dataservice/api/aliquot/models.py
```diff
@@ -0,0 +1,32 @@
...
+     from the sample or sample portion, measured in milligrams per milliliter
+    :param volume: The volume in microliters (ml) of the aliquots derived from
+     the analyte(s) shipped for sequencing and characterization
+    :param shipment_date: The date item was shipped in YYYY-MM-DD format
+    """
+    __tablename__ = "aliquot"
+    external_id = db.Column(db.Text())
+    shipment_origin = db.Column(db.Text())
+    shipment_destination = db.Column(db.Text())
+    analyte_type = db.Column(db.Text(), nullable=False)
```
dankolbman, line 25: Why is the analyte type required?

tests/aliquot/test_aliquot_models.py
```diff
@@ -0,0 +1,383 @@
...
+        s1 = Sample.query.filter_by(external_id=sample_id).all()
+        s2 = Sample.query.filter_by(external_id=s.external_id).all()
+        
+        a1 = Aliquot.query.filter_by(external_id=aliquot_id).all()
+        a2 = Aliquot.query.filter_by(external_id=a.external_id).all()
+        
+        #update a1 and a2
+        a1[0].analyte_type = 'RNA'
+        a2[0].volume = 13.89
+
```
dankolbman: I think you probably want to commit the session here to make sure the below queries don't just pull from the session.

tests/aliquot/test_aliquot_models.py
```diff
@@ -0,0 +1,383 @@
...
+        
+        a1 = Aliquot.query.filter_by(external_id=aliquot_id).one_or_none()
+        a2 = Aliquot.query.filter_by(external_id=a.external_id).one_or_none()
+        
+        #delete a1
+        db.session.delete(a1)
+        db.session.commit()
+        
+        self.assertEqual(Participant.query.count(), 1)
+        self.assertEqual(Sample.query.count(), 1)
```
dankolbman: Hmm so deleting the aliquot will also delete the its parent sample?

dankolbman (commented):
looks good, just some small nits

dataservice/api/aliquot/models.py
parimalak: analyte_type represents the kind of molecular specimen analyte.
in the kids first dictionary https://gen3.kids-first.io/dd/aliquot#/ ; the field is specified as required.

tests/aliquot/test_aliquot_models.py
parimalak: its because I have one aliquot for the sample. This is the example I tested ;a participant has two samples , each sample has one aliquot.

dankolbman: right, but is that sample is still connected to a participant? It seems that even though that sample no longer has any aliquots, it's still not an orphan, so it doesn't make sense to me for it to be deleted.

parimalak: I think aliquot ID is representative of the sample. so if we don't have an aliquot ID, they woudn't be any sample.aliquot is like the base unit (so to speak) of a sample.This is what Elizabeth told me. @allisonheath does a sample exists with out an aliquot?

tests/aliquot/test_aliquot_models.py
parimalak: yes. I am gonna make these changes.

tests/aliquot/test_aliquot_models.py
allisonheath: Yes a sample can exist without an aliquot. In reality, a sample always exists and an aliquot is either a piece of that sample or a processed piece of that sample. So I don't think they should be tightly coupled like this in general. In our case, the "sample" here is the top-level aliquot and the "aliquot" is the aliquot that's actually shipped to the sequencing center. We track a lot more "aliquots" in the intermediary, but that's doesn't generalize to other biospecimen processes. 

One issue for KF from most cohorts we often only have the aliquot information, nothing about the original sample taken from the participant. So there's a question of do we want the aliquot to go back directly to the participant as well in this case? Or create an artificial sample because it should have existed?

parimalak: To make myself clear , these are the cases right ?
1)If there is no sample information, but there is aliquot information then sample entity is just like a pointer that connects to participant
2)If Sample exists, no aliquots are extracted from it. Sample entity exists ;if any aliquot is extracted from this sample; those aliquot_ids are attached to this sample ; For now, in either case sample exists.

dataservice/api/sample/models.py
```diff
@@ -8,16 +9,24 @@ class Sample(db.Model, Base):
...
     """
     __tablename__ = "sample"
     external_id = db.Column(db.Text())
     tissue_type = db.Column(db.Text())
     composition = db.Column(db.Text())
     anatomical_site = db.Column(db.Text())
     age_at_event_days = db.Column(db.Integer())
-    participant_id = db.Column(db.Integer, db.ForeignKey('participant.kf_id'),
+    tumor_descriptor = db.Column(db.Text())
+    aliquots = db.relationship(Aliquot, backref='samples',
```
znatty22, line 29: @parimalak It looks like you made the changes to the models already. The way you have the relationships configured for Sample and Aliquot are correct now. This will achieve the one-to-many relationship going from Sample to Aliquots and also when you delete a Sample, it will delete all associated Aliquots. When you delete all the Aliquots for a given Sample, the Sample will still exist which is what we want. 👍

tests/aliquot/test_aliquot_models.py
```diff
@@ -0,0 +1,384 @@
...
+                    'concentration':100,
+                    'volume':12.67,
+                    'shipment_date':dt
+                    }
+        a = Aliquot(**aliquot_data,sample_id = '')
+
+        #Add aliquot to db
+        self.assertRaises(IntegrityError, db.session.add(a))
+
+    def test_one_to_many_relationship_create(self):
```
znatty22: Not sure that this is truly testing the one to many relationship between Sample and Aliquot. You are creating Sample s1 and adding Aliquot a1 to it. Then you are creating Sample s2 and adding aliquot a2 to it. I think you might want to try adding Aliquot a2 to Sample s1 so that you can verify that one sample has many aliquots.

znatty22 (changes requested):

dataservice/api/sample/models.py
parimalak: 👍

tests/aliquot/test_aliquot_models.py
parimalak: yes I could do that too. My initial thought was to test multi relationship between participant to sample and finally to aliquot. I will add another aliquot to one of the sample.

znatty22 (approved):

tests/aliquot/test_aliquot_models.py
```diff
@@ -0,0 +1,461 @@
...
+
+        a1 = Aliquot.query.filter_by(external_id=aliquot_id).one_or_none()
+        a2 = Aliquot.query.filter_by(external_id=a.external_id).one_or_none()
+
+        # delete a2
+        db.session.delete(a2)
+        db.session.commit()
+
+        self.assertEqual(Participant.query.count(), 1)
+        self.assertEqual(Sample.query.count(), 2)
```
dankolbman, line 460: 👍

dankolbman (approved):
Looks good! When you squash and rebase, could you also run `flask erd` to generate a new erd and commit it?

PR #75, opened 2018-01-29
✨ Diagnosis resource and tests

znatty22:
Add resources module for Diagnosis entity and unit tests


dataservice/api/diagnosis/resources.py
```diff
@@ -0,0 +1,117 @@
...
+
+        Update an existing diagnosis given a Kids First id
+        """
+        body = request.json
+        try:
+            # Check if diagnosis exists
+            d1 = Diagnosis.query.filter_by(kf_id=kf_id).one()
+            # For validation only
+            d = DiagnosisSchema(strict=True).load(body).data
+            # Deserialize
```
znatty22: As is the case with Demographic, this can probably be done more dynamically/better. But since its not trivial, will look into this in the future

dataservice/api/diagnosis/schemas.py
```diff
@@ -0,0 +1,36 @@
...
+    # still a bug in marshmallow_sqlalchemy. The bug is that ma sets
+    # dump_only=True for foreign keys by default. See link below
+    # https://github.com/marshmallow-code/marshmallow-sqlalchemy/issues/20
+    participant_id = field_for(Diagnosis, 'participant_id', required=True,
+                               load_only=True)
+
+    class Meta(BaseSchema.Meta):
+        model = Diagnosis
+
+    @validates('age_at_event_days')
```
znatty22: Wasn't sure what to use for the age range. Looked at kids first dictionary and used the min, max range for age_at_diagnosis_days. See [link](https://github.com/kids-first/kf-dictionary/blob/72aab588f06469ab74abaf406e52ef2d35e50289/gdcdictionary/schemas/diagnosis.yaml#L45)

dataservice/api/errors.py
```diff
@@ -1,6 +1,45 @@
 from dataservice.api.common.schemas import ErrorSchema
 
+FOREIGN_KEY_ERROR_PREFIX = 'foreign key constraint failed'
```
znatty22: I know this is duplicate code with demographic-entity branch. Just needed to include this since that branch has not been merged in and it is required to implement the diagnosis resource module.

dataservice/api/diagnosis/resources.py
```diff
@@ -0,0 +1,117 @@
...
+        # Not found in database
+        except NoResultFound:
+            abort(404, 'could not find {} `{}`'.format('diagnosis', kf_id))
+        # Request body not valid
+        except ValidationError as e:
+            abort(400, 'could not update diagnosis: {}'.format(e.messages))
+        # Database error
+        except IntegrityError as e:
+            db.session.rollback()
+            print(e)
```
dankolbman: 💣

dataservice/api/diagnosis/schemas.py
```diff
@@ -0,0 +1,36 @@
+from dataservice.api.diagnosis.models import Diagnosis
+from dataservice.api.common.schemas import BaseSchema
+from marshmallow_sqlalchemy import field_for
+from marshmallow import validates, ValidationError
+from dataservice.extensions import ma
+
+MIN_AGE_DAYS = 0
+MAX_AGE_DAYS = 32872
```
dankolbman: Isn't `INTEGER` type 4 bytes?

dataservice/api/errors.py
```diff
@@ -1,6 +1,45 @@
...
+
+DEFAULT_INVALID_INPUT_TEMPLATE = 'Client error: invalid {} provided'
+
 
 def http_error(e):
     """ Handles all HTTPExceptions """
     return ErrorSchema().jsonify(e), e.code
+
+
+def handle_integrity_error(**kwargs):
```
dankolbman: All of this shouldn't be necessary when we move to postgres. It returns much more detailed exceptions that can be parsed out in the native handler rather than an intermediate parser.

dataservice/api/diagnosis/resources.py
znatty22: Woops 😬, will remove

dataservice/api/diagnosis/schemas.py
znatty22: Yes, but I set the max based on the kids first max for age_at_diagnosis_days which is roughly 90 years.

dataservice/api/errors.py
znatty22: 👍

dataservice/api/diagnosis/schemas.py
dankolbman: where is this defined?

znatty22: https://github.com/kids-first/kf-dictionary/blob/72aab588f06469ab74abaf406e52ef2d35e50289/gdcdictionary/schemas/diagnosis.yaml#L45

dankolbman: hmm ok. But I think we can ignore that because we'll probably be taking that field out of the kf dictionary.

znatty22: 👍 I'll change the max to something that makes more sense for us (i.e. maybe 125 years or something like that?). @allisonheath - what do you think a reasonable max should be for age_at_event_days?

allisonheath: HIPAA safe harbor guidelines is that everyone above the age of 90 should be considered one group: https://www.hhs.gov/hipaa/for-professionals/privacy/special-topics/de-identification/index.html#standard

This is where 32872 (90 years in days) comes from and we should definitely keep it everywhere to ensure deidentification.

dankolbman: ooh forgot about that one 👍 
Should probably include a comment here about that.

dataservice/api/diagnosis/schemas.py
```diff
@@ -0,0 +1,36 @@
...
+        if value < MIN_AGE_DAYS:
+            raise ValidationError('Age must be an integer greater than {}.'
+                                  .format(MIN_AGE_DAYS))
+        if value > MAX_AGE_DAYS:
+            raise ValidationError('Age must be an integer less than {}.'
+                                  .format(MAX_AGE_DAYS))
+
+    _links = ma.Hyperlinks({
+        'self': ma.URLFor('api.diagnoses', kf_id='<kf_id>'),
+        'collection': ma.URLFor('api.diagnoses'),
```
dankolbman: will this be the same as `self` when getting `/diagnoses`?

dataservice/api/diagnosis/schemas.py
znatty22: 👍

dataservice/api/diagnosis/schemas.py
znatty22: self will be /diagnoses/<kf_id> where as collection will be just /diagnoses

dankolbman: Right, but what about when `<kf_id>` is `None`, like when I would do `GET /diagnoses/`

dataservice/api/diagnosis/resources.py
```diff
@@ -0,0 +1,117 @@
...
+                abort(404, 'could not find {} `{}`'
+                      .format('diagnosis', kf_id))
+            return DiagnosisSchema().jsonify(d)
+
+    def post(self):
+        """
+        Create a new diagnosis
+        """
+        body = request.json
+        try:
```
dankolbman: This should be two `try/catch` blocks, one for the validation, one for the transaction.

dataservice/api/diagnosis/resources.py
```diff
@@ -0,0 +1,117 @@
...
+                               .format(d.kf_id)).jsonify(d), 201
+
+    def put(self, kf_id):
+        """
+        Update existing diagnosis
+
+        Update an existing diagnosis given a Kids First id
+        """
+        body = request.json
+        try:
```
dankolbman, line 77: Should also be split up

dataservice/api/diagnosis/resources.py
dankolbman: 💥

dataservice/api/errors.py
dankolbman: should be taken care of during rebase

dankolbman (changes requested):
Just need to clean up the `try/catch`

dankolbman (approved):

parimalak (approved):
:ok_hand:

PR #77, opened 2018-01-29
✨ Add pagination

### LINKED ISSUE #12: Add pagination to participant resource
The top level resources that return lists should return items in a paginated format using a configurable number of results.

This should be implemented using good practices for pagination against the database such as use of a cursor.

Information about the page number, total number of results, etc. should also be returned either through the header or in a standardized envelope.

Examples:

```
GET /persons

{
  "_links": {
    "prev": "/persons?page=1",
    "self": "/persons?page=2",
    "next": "/persons?page=3"
  },
  "total": 23,
  "limit": 10,
  "results": {
    {"kf_id": "001"},
    {"kf_id": "002"},
    ...
    {"kf_id": "010"}
  }
}
```

dankolbman:
Uses the `created_at` timestamp to paginate through participants.
Can also use anything that may be parsed into a `datetime` by `dateutil.parser`.
Eg: `/participants?after=01-01-2017`

Other additions that could be made:

- Allow reverse iteration with a `?before=` (from newest to oldest)
- Add an index to the `created_at` column in pg to speed up filtering

Resolves #12

### DIFF
--- a/dataservice/api/common/pagination.py
+++ b/dataservice/api/common/pagination.py
@@ -1,5 +1,7 @@
 from flask import request, current_app
 from functools import wraps
+from dateutil import parser
+from datetime import datetime
 
 
 def paginated(f):
@@ -9,7 +11,76 @@ def paginated(f):
         def_limit = current_app.config['DEFAULT_PAGE_LIMIT']
         max_limit = current_app.config['MAX_PAGE_LIMIT']
         limit = min(request.args.get('limit', def_limit, type=int), max_limit)
-        page = request.args.get('page', 1, type=int)
-        return f(*args, **kwargs, limit=limit, page=page)
+        after = request.args.get('after', None)
+
+        if type(after) is str:
+            # Parser won't recognize the timestamp with fractional seconds
+            if after.replace('.', '').isdigit():
+                after = float(after)
+                after = datetime.fromtimestamp(after)
+            else:
+                try:
+                    after = parser.parse(after)
+                # Parser couldn't derive a datetime from the string
+                except ValueError:
+                    after = None
+
+        # Default to the unix epoch
+        if after is None:
+            after = datetime.fromtimestamp(0)
+
+        return f(*args, **kwargs, after=after, limit=limit)
 
     return paginated_wrapper
+
+
+class Pagination(object):
+    """
+    Object to help paginate through endpoints using the created_at field
+    """
+
+    def __init__(self, query, after, limit):
+        assert type(after) is datetime
+        self.query = query
+        self.after = after
+        self.limit = limit
+        self.total = query.count()
+        # Assumes that we only provide queries for one entity
+        # This is safe as pagination only accesses one entity at a time
+        model = query._entities[0].mapper.entity
+        assert hasattr(model, 'created_at')
+        self.items = (query.order_by(model.created_at.asc())
+                           .filter(model.created_at > after)
+                           .limit(limit).all())
+
+    @property
+    def prev_num(self):
+        """ Returns the timestamp of the first item """
+        if len(self.items) > 0:
+            return self._to_timestamp(self.items[0].created_at)
+        return datetime.fromtimestamp(0)
+
+    @property
+    def curr_num(self):
+        """
+        Returns the timestamp of the first item minus some epsilon
+        so that the current timestamp will be included in the range
+        """
+        if len(self.items) > 0:
+            return self._to_timestamp(self.items[0].created_at) - 1/1000
+        return self._to_timestamp(datetime.utcnow())
+
+    @property
+    def next_num(self):
+        """ Returns the timestamp of the last item"""
+        if self.has_next:
+            return self._to_timestamp(self.items[-1].created_at)
+
+    @property
+    def has_next(self):
+        """ True if there are more than `limit` results """
+        return len(self.items) >= self.limit
+
+    def _to_timestamp(self, dt):
+        """ Converts a datetime object to milliseconds since epoch """
+        return dt.timestamp()
--- a/dataservice/api/common/pagination.py
+++ b/dataservice/api/common/pagination.py
@@ -6,9 +6,9 @@ def paginated(f):
 
     @wraps(f)
     def paginated_wrapper(*args, **kwargs):
-        def_size = current_app.config['DEFAULT_PAGE_SIZE']
-        max_size = current_app.config['MAX_PAGE_SIZE']
-        limit = min(request.args.get('size', def_size, type=int), max_size)
+        def_limit = current_app.config['DEFAULT_PAGE_LIMIT']
+        max_limit = current_app.config['MAX_PAGE_LIMIT']
+        limit = min(request.args.get('limit', def_limit, type=int), max_limit)
         page = request.args.get('page', 1, type=int)
         return f(*args, **kwargs, limit=limit, page=page)
 
--- a/dataservice/api/common/pagination.py
+++ b/dataservice/api/common/pagination.py
@@ -1,0 +1,15 @@
+from flask import request, current_app
+from functools import wraps
+
+
+def paginated(f):
+
+    @wraps(f)
+    def paginated_wrapper(*args, **kwargs):
+        def_size = current_app.config['DEFAULT_PAGE_SIZE']
+        max_size = current_app.config['MAX_PAGE_SIZE']
+        limit = min(request.args.get('size', def_size, type=int), max_size)
+        page = request.args.get('page', 1, type=int)
+        return f(*args, **kwargs, limit=limit, page=page)
+
+    return paginated_wrapper
--- a/dataservice/api/common/schemas.py
+++ b/dataservice/api/common/schemas.py
@@ -6,8 +6,8 @@ from marshmallow import (
         validates_schema,
         ValidationError
 )
-from flask import url_for
-from flask_sqlalchemy import Pagination
+from flask import url_for, request
+from dataservice.api.common.pagination import Pagination
 from flask_marshmallow import Schema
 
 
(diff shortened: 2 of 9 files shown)


dataservice/api/participant/resources.py
```diff
@@ -16,8 +16,15 @@ def get(self, kf_id):
         Gets a participant given a Kids First id
         """
         if kf_id is None:
+            def_size = current_app.config['DEFAULT_PAGE_SIZE']
+            max_size = current_app.config['MAX_PAGE_SIZE']
+            size = min(request.args.get('size', def_size, type=int), max_size)
+            page = request.args.get('page', 1, type=int)
+
+            q = Participant.query.order_by(Participant.created_at.desc())
```
dankolbman: Uses `limit() offset()`, so not ideal. Will have to implement our own pagination later using `limit() filter()` on the created_at field.

dataservice/api/participant/schemas.py
```diff
@@ -7,7 +7,8 @@ class ParticipantSchema(BaseSchema):
 
     class Meta(BaseSchema.Meta):
         model = Participant
+        resource_url = 'api.participants'
```
znatty22: Hmm this is making me think that later on we might want to define the url strings in one place and reference those through out the application

dataservice/api/participant/resources.py
```diff
@@ -16,8 +16,15 @@ def get(self, kf_id):
         Gets a participant given a Kids First id
         """
         if kf_id is None:
+            def_size = current_app.config['DEFAULT_PAGE_SIZE']
+            max_size = current_app.config['MAX_PAGE_SIZE']
+            size = min(request.args.get('size', def_size, type=int), max_size)
```
znatty22: I think we should be consistent with naming. The query param in the request url is called `size` but in the response its called `limit`.

dataservice/api/common/schemas.py
```diff
@@ -14,18 +18,55 @@ class Meta:
         exclude = ('_id', 'uuid')
         dump_only = ('created_at', 'modified_at')
 
+    @pre_dump(pass_many=True)
+    def wrap_pre(self, data, many):
+        if type(data)is Pagination:
```
znatty22: Probably a good idea to use `isinstance`

dataservice/api/participant/resources.py
```diff
@@ -16,8 +16,15 @@ def get(self, kf_id):
         Gets a participant given a Kids First id
         """
         if kf_id is None:
+            def_size = current_app.config['DEFAULT_PAGE_SIZE']
```
znatty22: This bit of code will be common for all the resources. Can we move it somewhere in common so all resources can use it? Maybe you could even put this in a PaginationParams marshmallow schema class and do some validation on the pagination params like limit, offest (if we support it), etc.

dataservice/api/common/pagination.py
```diff
@@ -0,0 +1,15 @@
...
+from functools import wraps
+
+
+def paginated(f):
+
+    @wraps(f)
+    def paginated_wrapper(*args, **kwargs):
+        def_size = current_app.config['DEFAULT_PAGE_SIZE']
+        max_size = current_app.config['MAX_PAGE_SIZE']
+        limit = min(request.args.get('size', def_size, type=int), max_size)
```
znatty22: I think we should name the results per page parameter consistently. In the query params its named `size` but in the responses that come back its called `limit`. Lets choose one or the other....but personally I like `limit_per_page` or `results_limit`, even though those are not conventional. I like them bc their meanings are clearer to me than `limit` or `size`

dataservice/api/common/schemas.py
```diff
@@ -14,18 +18,55 @@ class Meta:
         exclude = ('_id', 'uuid')
         dump_only = ('created_at', 'modified_at')
 
+    @pre_dump(pass_many=True)
+    def wrap_pre(self, data, many):
+        if isinstance(data, Pagination):
+            self.__pagination__ = data
+            if self.Meta.resource_url:
```
znatty22: I'm not seeing how/where self._links is used? Do you still need this bit of code?

dataservice/api/participant/resources.py
```diff
@@ -1,23 +1,28 @@
...
 from dataservice.extensions import db
+from dataservice.api.common.pagination import paginated
 from dataservice.api.participant.models import Participant
 from dataservice.api.participant.schemas import ParticipantSchema
 
 
 class ParticipantAPI(MethodView):
-    def get(self, kf_id):
+
+    @paginated
```
znatty22: 👍  nice

znatty22 (commented):
Looks good, see minor comments on files

dataservice/api/common/pagination.py
```diff
@@ -0,0 +1,86 @@
...
+    @property
+    def next_num(self):
+        """ Returns the timestamp of the last item"""
+        if self.has_next:
+            return self._to_timestamp(self.items[-1].created_at)
+
+    @property
+    def has_next(self):
+        """ True if there are more than `limit` results """
+        return len(self.items) == self.limit
```
znatty22: Should it be >?

dataservice/api/common/pagination.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @property
+    def curr_num(self):
+        """
+        Returns the timestamp of the first item minus some epsilon
+        so that the current timestamp will be included in the range
+        """
+        if len(self.items) > 0:
+            return self._to_timestamp(self.items[0].created_at) - 1/1000
+        return datetime.fromtimestamp(0)
```
znatty22: Do we want to default to unix epoch here? Should it just return the request date `after` with no results? Or maybe just default to the first page.

Right now when you make a request with a future date that should return no results: 
`localhost:5000/participants?after=2018-02-28`

I get back:
```
{
    "_links": {
        "self": "/participants?after=1969-12-31+19%3A00%3A00"
    },
    "_status": {
        "code": 200,
        "message": "success"
    },
    "limit": 10,
    "results": [],
    "total": 1203
}
``` 
I think the self link should be set to the request `after` date in the original request or return first page of results by default with the `after` date kept as is (unix epoch)

dataservice/api/common/schemas.py
```diff
@@ -19,18 +24,47 @@ class Meta:
...
+        in the response will be populated with pagination details.
+        """
+        resp = {'_status': {'message': self.status_message,
+                            'code': self.status_code}}
+        # Move links to the envelope links if just a single object
         if not many and '_links' in data:
             _links = data['_links']
             del data['_links']
+        # Insert pagination object, if there is one
+        elif many and self.__pagination__ is not None:
```
znatty22, line 52: Should a link to prev be in here too?

dataservice/api/common/schemas.py
```diff
@@ -19,18 +24,47 @@ class Meta:
...
+            p = self.__pagination__
+            _links = {}
+            _links['self'] = url_for(self.Meta.resource_url,
+                                     after=p.curr_num)
```
znatty22: Can we render the `self` `next` and `prev` links as datetime strings instead of seconds? 

It's a little hard to tell what these numbers are or get a sense of dates without scrolling through the list of results to see what the created_at fields are for each result.
```
        "next": "/participants?after=1519430653.953975",
        "self": "/participants?after=1519430653.946914"
```

dataservice/api/common/pagination.py
dankolbman: It can't be more than `limit` number of items, but I've made it `>=` to be safe.

dataservice/api/common/pagination.py
dankolbman: Yeah, doesn't make sense to return epoch. We would expect some number of results then. Likewise, it doesn't make sense to return the first page, either.

I'm going to default to whatever the requested url was using request.full_path, and default to the current timestamp for the this property.

dataservice/api/common/schemas.py
dankolbman: Only supports iteration in one direction, at the moment.

dataservice/api/common/schemas.py
dankolbman: Maybe. I wanted to obscure the fact that those numbers are actually timestamps and let the user think of them more as a cursor.

dataservice/api/common/schemas.py
```diff
@@ -19,18 +24,46 @@ class Meta:
...
+                            'code': self.status_code}}
+        # Move links to the envelope links if just a single object
         if not many and '_links' in data:
             _links = data['_links']
             del data['_links']
+        # Insert pagination object, if there is one
+        elif many and self.__pagination__ is not None:
+            p = self.__pagination__
+            _links = {}
+            _links['self'] = request.full_path
```
znatty22: Hmm, I think you might want something different here. If I send this request:  
`localhost:5000/participants?after=adasdf`, I get back:
```
{
    "_links": {
        "next": "/participants?after=1519762761.405353",
        "self": "/participants?after=adasdf"
    },
    "_status": {
        "code": 200,
        "message": "success"
    },
    "limit": 10,
    "results": [
      actual results here ...
    ],
    "total": 300
}
```
Maybe you could do something like: 
```python
if p.after.timestamp() == 0.0:
    after = None
else:
    after = p.after

_links['self'] = url_for(self.Meta.resource_url,
                         after=after)
```
Which would return: 
```
{
    "_links": {
        "next": "/participants?after=1519762761.405353",
        "self": "/participants
    },
    "_status": {
        "code": 200,
        "message": "success"
    },
    "limit": 10,
    "results": [
        actual results here ...
    ],
    "total": 300
}
```
Or maybe you could do:
```python
_links['self'] = url_for(self.Meta.resource_url,
                         after=p.after)
```
Which would keep the default of unix epoch. However, if I was the user and didn't know that this timestamp was epoch I might be a little confused as to where the random date came from 😛 . It'd be less confusing to me if i just got back `self: /participants`

dankolbman: Made it so that it will default to just `/participants` if the `after` param cannot be parsed.

znatty22 (approved):
👍

dataservice/api/participant/resources.py
```diff
@@ -27,8 +31,15 @@ def get(self):
             resource:
               Participant
         """
+        q = (Participant.query
+                        .options(joinedload(Participant.diagnoses))
+                        .options(joinedload(Participant.samples))
```
parimalak, line 36: are we writing similar lines in samples resources as well?

dankolbman: Yeah, we'll want to do some basic benchmarks as we implement different resources to make sure we're doing things effectively.

PR #78, opened 2018-01-30
🐘 Postgres

### LINKED ISSUE #28: Switch to use Postgres as primary backend
Although SQLite reduces dependency requirements for development, we'll need to move to Postgres for our deployments. We should at least support this in the `ProductionConfig` `SQLALCHEMY_DATABASE_URI`.

### LINKED ISSUE #29: Secret management with vault
To be able to connect to Postgres in our deployment environments, we will need to maintain a password to Postgres. We are using Vault, so will need to incorporate the retrieval of our password with the Vault cluster, probably using the [hvac](https://github.com/ianunruh/hvac) package.

### LINKED ISSUE #76: Change particpant_id type in sample model
The `participant_id` foreign key constraint should be changed to match the `kf_id` format, `String(8)`, otherwise postgres will fail to create the tables.

dankolbman:
Reworks configuration, Jenkins pipeline, CircleCI, error handling, and secret management for operation with Postgres.

Fixes #28,
Fixes #29
Fixes #76 

- [x]  Use postgres in local environments
- [x]  Change CicrleCI to use postgres
- [x]  Change Jenkins to use postgres
- [x]  Integrate vault to use postgres on aws


README.md
```diff
@@ -58,6 +72,9 @@ Unit tests and pep8 linting is run via `flask test`
 ```
 # Install test dependencies
 pip install -r dev-requirements.txt
+# Setup test database
+docker run --name dataservice-pg -p 5432:5432 -d postgres
+docker exec dataservice-pg psql -U postgres -c "CREATE DATABASE dev;"
```
znatty22: This should be `"CREATE DATABASE test;"` right?

README.md
```diff
@@ -21,12 +21,18 @@ pip install -r requirements.txt
 # Configure the flask application
 export FLASK_APP=manage
 # Setup the database
-flask db init
+docker run --name dataservice-pg -p 5432:5432 -d postgres:9.5
+docker exec dataservice-pg psql -U postgres -c "CREATE DATABASE dev;"
 flask db upgrade
```
znatty22: Need to add a line above here with `flask db migrate`, otherwise the relations don't exist on the first time you try to run the app

tests/test_errors.py
```diff
@@ -0,0 +1,63 @@
...
+        # Check error message
+        if len(kf_id) <= 8:
+            message = '"{}" does not exist'.format(kwargs['participant_id'])
+            assert message in response['_status']['message']
+        else:
+            message = 'Longer than maximum length 8'
+            assert message in response['_status']['message']
+
+
+    def test_unique(self, client):
```
znatty22: Maybe this should be called test_unique_demographic since its specific to the demographic and we could have other one-to-one entity tests in this class.

tests/test_errors.py
```diff
@@ -0,0 +1,63 @@
...
+        """ Test integrity errors trying to break one-many with many-many """
```
znatty22: Shouldn't this be
```python 
""" Test integrity errors trying to break one-to-one with one-to-many """
``` 
since a participant can only have on demographic and a demographic can only be related to one participant?

znatty22 (commented):
Looks good! Just a couple minor comments on test_errors.py and the README.md

bin/initdb.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+flask db upgrade
+flask db migrate
```
znatty22: I think it should be migrate first then upgrade?

bin/run.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+source venv/bin/activate
+flask db upgrade
+flask db migrate
```
znatty22: Same here, I think migrate first then upgrade

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
+version: '3'
+
+services:
+  pg:
```
znatty22, line 4: Can we expose the pg port so that people can use other apps like pgadmin to connect

dankolbman: I'd rather not to avoid any conflict with local instances. Someone familiar enough to be connecting to the db should be able to figure out how to expose the port here.

znatty22: 👍

znatty22 (approved):
Looks great!

parimalak (approved):

PR #82, opened 2018-01-31
✨ Sample resource and tests

znatty22:
Add sample resource with crud methods and unit tests

* Note that nested de/serialization does not work right now. For example, cannot create/update/read a sample with embedded aliquots. This will implemented in near future.

### DIFF
--- a/dataservice/api/sample/resources.py
+++ b/dataservice/api/sample/resources.py
@@ -25,6 +25,7 @@ class SampleAPI(MethodView):
         Get a sample by Kids First id or get all samples if
         Kids First id is None
         """
+
         # Get all
         if kf_id is None:
             s = Sample.query.all()
@@ -43,16 +44,20 @@ class SampleAPI(MethodView):
         """
         Create a new sample
         """
+
         body = request.json
+
+        # Deserialize
         try:
-            # Deserialize
             s = SampleSchema(strict=True).load(body).data
-            # Add to and save in database
-            db.session.add(s)
-            db.session.commit()
         # Request body not valid
         except ValidationError as e:
             abort(400, 'could not create sample: {}'.format(e.messages))
+
+        # Add to and save in database
+        try:
+            db.session.add(s)
+            db.session.commit()
         # Database error
         except IntegrityError as e:
             db.session.rollback()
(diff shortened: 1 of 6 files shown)


dataservice/api/sample/schemas.py
```diff
@@ -0,0 +1,27 @@
+from marshmallow_sqlalchemy import field_for
+
+from dataservice.api.sample.models import Sample
+from dataservice.api.common.schemas import BaseSchema
+from dataservice.api.common.validation import validate_age
```
dankolbman, line 5: Why extract validation from the Schema?

dataservice/api/sample/resources.py
```diff
@@ -0,0 +1,120 @@
...
+            # Deserialize
+            s1.external_id = body.get('external_id')
+            s1.tissue_type = body.get('tissue_type')
+            s1.composition = body.get('composition')
+            s1.anatomical_site = body.get('anatomical_site')
+            s1.tumor_descriptor = body.get('tumor_descriptor')
+            s1.aliquots = body.get('aliquots', [])
+            s1.age_at_event_days = body.get('age_at_event_days')
+            s1.participant_id = body.get('participant_id')
+            db.session.commit()
```
dankolbman: probably best to limit the try block to just the actions that will throw.

dankolbman (commented):
minor comments, looks good otherwise

dataservice/api/sample/resources.py
```diff
@@ -0,0 +1,120 @@
...
+                abort(404, 'could not find {} `{}`'
+                      .format('sample', kf_id))
+            return SampleSchema().jsonify(s)
+
+    def post(self):
+        """
+        Create a new sample
+        """
+        body = request.json
+        try:
```
dankolbman: The schema validation and session transaction should be two different try/catches.

dataservice/api/sample/schemas.py
dankolbman: Saw the other use of the validator 👍

dataservice/api/sample/resources.py
znatty22: Ah sorry I meant to move it out. Will do

dataservice/api/sample/resources.py
znatty22: Why? It would essentially be the same thing. Each line in the try except throws a very specific exception

dataservice/api/sample/resources.py
znatty22: Actually sorry I was looking at this on my phone and didn't see everything. I retract my prev comment. I think everything in the try block is fine bc we are catching the specific exceptions (and nothing more) that we expect can be thrown.

dankolbman (approved):
👍 👍

parimalak (approved):
:ok_hand:

PR #85, opened 2018-02-02
♻️ Refactor pg exceptions to flask error handler

dankolbman:
Old error handling for sqlite breaks when we change backends. This implements a general flask exception handler for `IntegrityError` that processes error responses from postgres.

### DIFF
--- a/dataservice/api/errors.py
+++ b/dataservice/api/errors.py
@@ -1,14 +1,14 @@
+import re
 from dataservice.api.common.schemas import ErrorSchema
+from dataservice.extensions import db
 
-FOREIGN_KEY_ERROR_PREFIX = 'foreign key constraint failed'
-FOREIGN_KEY_ERROR_MSG_TEMPLATE = \
-    'Cannot {} {} without an existing {} entity'
 
-UNIQUE_CONSTRAINT_ERROR_PREFIX = 'unique constraint failed'
-UNIQUE_CONSTRAINT_ERROR_MSG_TEMPLATE = \
-    'Cannot {0} {1}, {2} already has a {1}'
+FOREIGN_KEY_RE = re.compile('.*\nDETAIL:.*\((?P<kf_id>.*)\)' +
+                            ' is not present in table "(?P<table>\w+)"\.')
 
-DEFAULT_INVALID_INPUT_TEMPLATE = 'Client error: invalid {} provided'
+UNIQUE_RE = re.compile('^.*"(?P<key>.*)_key"\n' +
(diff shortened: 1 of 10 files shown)


dataservice/api/errors.py
```diff
@@ -1,45 +1,46 @@
...
-FOREIGN_KEY_ERROR_PREFIX = 'foreign key constraint failed'
-FOREIGN_KEY_ERROR_MSG_TEMPLATE = \
-    'Cannot {} {} without an existing {} entity'
 
-UNIQUE_CONSTRAINT_ERROR_PREFIX = 'unique constraint failed'
-UNIQUE_CONSTRAINT_ERROR_MSG_TEMPLATE = \
-    'Cannot {0} {1}, {2} already has a {1}'
-
-DEFAULT_INVALID_INPUT_TEMPLATE = 'Client error: invalid {} provided'
+FOREIGN_KEY_RE = re.compile(r'DETAIL:.*\((?P<kf_id>\w{8})\)' +
```
znatty22: I think you may want to change this regex to be a bit more general (i.e don't limit to 8 chars, etc). Right now requests like 
```
POST /demographics
{
"participant_id": ""
}
``` 
result in the default error message of `error saving changes`.

dataservice/api/errors.py
```diff
@@ -1,45 +1,46 @@
...
 
 
 def http_error(e):
     """ Handles all HTTPExceptions """
     return ErrorSchema().jsonify(e), e.code
 
 
-def handle_integrity_error(**kwargs):
+def integrity_error(e):
     """
```
znatty22, line 20: Update docstring

dataservice/api/errors.py
```diff
@@ -1,45 +1,46 @@
...
+    m = FOREIGN_KEY_RE.match(error)
+    # Tried to create entity, but another entity must be created first
+    if m and 'kf_id' in m.groups():
+        message = 'A {} must be created first'.format(m.group('table'))
+    # Tried to create an entity dependent on an entity that doesn't exist
+    elif m:
+        message = '{} "{}" does not exist'.format(m.group('table'),
+                                                  m.group('kf_id'))
+
+    m = UNIQUE_RE.match(error)
```
znatty22, line 39: Hmm the error message is a little confusing here. If someone tries to create a demographic for a participant that already has a demographic the error message is:  `participant "144WDN78" already exists`

Can we make it something like: `participant "144WDN78" already has demographic`.
You could get the primary entity (demographic) by parsing e.statement (i.e. INSERT INTO demographic, UPDATE demographic etc).

PR #89, opened 2018-02-03
✨ Dummy data generator

parimalak:


dataservice/util/data_gen/data_generator.py
```diff
@@ -0,0 +1,183 @@
+from datetime import datetime
+from dateutil import relativedelta
+import uuid
+import random
+
+from dataservice.util.data_gen.data import *
```
dankolbman: `*` should be avoided.
You can look up python start imports for all the reasons it's bad.

dataservice/util/data_gen/data.py
```diff
@@ -0,0 +1,121 @@
...
+the entities
+"""
+import random
+import csv
+
+NUM_PARTICIPANTS = 10
+
+# NUM_DEMOGRAPHICS = 1
+
+NUM_DIAGNOSES = random.randint(0, 35)
```
dankolbman: Setting these to a random number on import seems counter intuitive.
This way, all samples will have the same number of aliquots, but that number will change between runs.
Using a random number of children for each parent may be a better idea. Perhaps this number could be `MAX_DIAGNOSES` then

dataservice/util/data_gen/data_generator.py
```diff
@@ -0,0 +1,183 @@
...
+from dataservice.api.participant.models import Participant
+from dataservice.api.demographic.models import Demographic
+from dataservice.api.diagnosis.models import Diagnosis
+from dataservice.api.sample.models import Sample
+from dataservice.api.aliquot.models import Aliquot
+from dataservice.api.sequencing_experiment.models import SequencingExperiment
+from dataservice.api.genomic_file.models import GenomicFile
+
+
+class DataGenerator(object):
```
dankolbman, line 19: Since you have a class to represent the generator's state, why no use class attributes instead of global variables?

dataservice/util/data_gen/data.py
```diff
@@ -0,0 +1,121 @@
...
+NUM_SAMPLES = random.randint(0, 65)
+
+NUM_ALIQUOTS = random.randint(0, 65)
+
+NUM_SEQ_EXP = 1
+
+NUM_GENOMIC_FILES = 5
+
+# Demographics data
+race_list = [
```
dankolbman: Might be nice to put these in a `txt` file to stay consistent and have less clutter here.

dankolbman (changes requested):
Looks like a lot of files are duplicated between `tests/` and `dataservice/utils/`?

dataservice/util/data_gen/data_generator.py
znatty22: Ditto

dataservice/util/data_gen/data_generator.py
```diff
@@ -0,0 +1,183 @@
+from datetime import datetime
+from dateutil import relativedelta
+import uuid
+import random
+
+from dataservice.util.data_gen.data import *
+from dataservice import create_app
+from dataservice.extensions import db
+from dataservice.api.participant.models import Participant
+from dataservice.api.participant.models import Participant
```
znatty22: Remove duplicate import

dataservice/util/data_gen/data_generator.py
znatty22: Agreed, I think you could have class attributes to represent the default totals for each entity and then use them as default args for each create entity method. 

And then probably the rest of the data/inputs driving the generator should all be outside of the class. I see that this class has some inputs provided from data.py and other inputs hardcoded in. I know it would be tedious to move all the inputs out 😑 ... but then the generator is totally agnostic of the inputs its using to build the data. And then its also nice bc you could have different data input files for the generator. Maybe something for the next version though?

tests/dummy_data_generator/test_data_generator.py
```diff
@@ -0,0 +1,251 @@
+from datetime import datetime
```
znatty22: I think you meant to remove the tests/dummy_data_generator folder right?

dataservice/util/data_gen/data.py
znatty22: Agreed. These numbers should either be constant totals for each entity type or maximums so that you can randomize the total (bounded by max) of that entity at runtime not during import

znatty22 (commented):
Looks pretty good! See comments on files

tests/dummy_data_generator/test_data_generator.py
znatty22: Remember to remove tests/dummy_data_generator

tests/dummy_data_generator/test_data_generator.py
```diff
@@ -0,0 +1,251 @@
...
+        for line in reader:
+            diagnosis_list.append(line[0])
+        progression_or_recurrence_list = ['yes', 'no', 'unknown']
+        diag_list = []
+        for i in range(total):
+
+            data = {
+                'external_id': 'diagnosis_{}'.format(i),
+                'diagnosis': random.choice(diagnosis_list),
+                'progression_or_recurrence':
```
znatty22: Remove progression_or_recurrence. This field is no longer in the Diagnosis model

znatty22:
I think in the future we may want to decouple the data input generation (creating/reading in choices, setting default maximums, etc) from the data generation/population (creating db objects and publishing to db). But this is probably fine for now in terms of an initial MVP

dataservice/util/data_gen/data_generator.py
```diff
@@ -0,0 +1,319 @@
...
+        """
+        creates diagnoses
+        """
+        self.diag_list = []
+        for i in range(total):
+
+            self.data = {
+                'external_id': 'diagnosis_{}'.format(i),
+                'diagnosis': random.choice(self.diagnosis_list),
+                'progression_or_recurrence':
```
znatty22: Remove progression_or_recurrence. This field is no longer in the Diagnosis model

tests/genomic_file/test_geonmic_file_models.py
```diff
@@ -22,6 +22,7 @@ def test_create_and_find(self):
         """
```
dankolbman, line 22: Looks like a typo in this file's name

dataservice/util/data_gen/data_generator.py
```diff
@@ -0,0 +1,316 @@
...
+    def _create_experiments(self, total):
+        """
+        Creates sequencing experiments
+        """
+
+        self.e_list = []
+        dt = datetime.now()
+        for i in range(total):
+
+            self.e_data = {
```
dankolbman: Could do with just assigning to `e_data` rather than a class property. I'm guessing you never actually need `self.e_data`?

dataservice/util/data_gen/data_generator.py
```diff
@@ -0,0 +1,316 @@
...
+                'tissue_type': random.choice(self.tissue_type_list),
+                'composition': random.choice(self.composition_list),
+                'anatomical_site': random.choice(self.anatomical_site),
+                'age_at_event_days': random.randint(0, 32872),
+                'tumor_descriptor': random.choice(self.tumor_descriptor_list)
+            }
+            self.s_list.append(
+                Sample(
+                    **self.sample_data,
+                    aliquots=self._create_aliquots(
```
dankolbman: This looks pretty messy and is hard to follow. Maybe try defining the aliquots first, then passing them in?

```
aliquots = self._create_aliquots()
Sample(**sample_data, aliquots=aliquots)
```

dataservice/util/data_gen/data_generator.py
```diff
@@ -0,0 +1,316 @@
...
+        # Create participants
+        self._create_participants(self.max_participants)
+        self.teardown()
+
+    def _create_participants(self, total):
+        """
+        creates participants with samples, demographics, and diagnoses
+        """
+        for i in range(total):
+            self.p = Participant(
```
dankolbman: p probably doesn't need to be a class property. Same comment as below about assigning samples and other relationships beforehand.

dataservice/util/data_gen/data_generator.py
```diff
@@ -0,0 +1,316 @@
...
+                external_id='participant_{}'.format(i),
+                samples=self._create_samples(
+                    random.randint(
+                        self.min_samples,
+                        self.max_samples)),
+                demographic=self._create_demographics(i),
```
dankolbman: I think you want `_create_demographics(1)` since you can only have one demographic per participant.

dataservice/util/data_gen/data_generator.py
```diff
@@ -0,0 +1,316 @@
...
+
+            self.e_data = {
+                'external_id': 'sequencing_experiment_{}'.format(i),
+                'experiment_date': dt - relativedelta.relativedelta(
+                    years=random.randint(1, 3)) + relativedelta.relativedelta(
+                    months=random.randint(1, 6)) +
+                relativedelta.relativedelta(days=random.randint(1, 30)),
+                'experiment_strategy':
+                random.choice
+                (self.experiment_strategy_list),
```
dankolbman: the function arguments should really be on the same line as the function

dankolbman:
Should be sufficient for a first pass, but there's some stuff that could be cleaned up.
I'm not too big a fan of being heavily reliant on the class properties that are assigned within functions, eg:

```
class DataMocker:
    def __init__(self):
        self._genomic_files()

    def _genomic_files(self):
        self.max_gen_files = 10
        self.min_gen_files = 1
        self.file_format_list = ['.txt', '.csv']
```

You now have a bunch of class properties that follow different naming schemes all over the file, and you'd have to grep through to find the one you're looking for. You could perhaps create some sort of `MockEntity` class with `entity_class`, `min`, `max`, `sample_values` attributes for each entity, but that may be over abstracting in this case. You may be better off trying to use some sort of `dict` to do it instead:

```
entities = [ Participant, GenomicFile, ... ]
self.entity_ranges =  {c: {
    'min': 5,
    'max': 25,
    'choices': open(c.__class__.__name__+'.txt').read().split('\n')
} for c in entities }
```

Now you could do `self.entity_ranges[Participant]` and get back the min, max, and choices in a consistent manner.

Perhaps worth considering adding a `fake()` static function on each of the entities that returns a new entity with appropriate fields set.

[Here's](https://github.com/klen/mixer#flask-flask-sqlalchemy) a cool library that may give some inspiration, too.

dataservice/util/data_gen/data_generator.py
parimalak: i am just trying to have same suffix for creating demographic. 
external_id': 'demo_id_{}'.format(i)

dataservice/util/data_gen/data_generator.py
```diff
@@ -0,0 +1,310 @@
...
+from dataservice.api.diagnosis.models import Diagnosis
+from dataservice.api.sample.models import Sample
+from dataservice.api.aliquot.models import Aliquot
+from dataservice.api.sequencing_experiment.models import SequencingExperiment
+from dataservice.api.genomic_file.models import GenomicFile
+
+
+class DataGenerator(object):
+
+    def __init__(self, config_name='testing'):
```
dankolbman: Can you set the config name to default to `FLASK_CONFIG` in the environment?

dataservice/util/data_gen/data_generator.py
```diff
@@ -0,0 +1,310 @@
...
+from dataservice.api.demographic.models import Demographic
+from dataservice.api.diagnosis.models import Diagnosis
+from dataservice.api.sample.models import Sample
+from dataservice.api.aliquot.models import Aliquot
+from dataservice.api.sequencing_experiment.models import SequencingExperiment
+from dataservice.api.genomic_file.models import GenomicFile
+
+
+class DataGenerator(object):
+    def __init__(self, config_name=os.environ.get('FLASK_CONFIG', 'default')):
```
dankolbman: Probably not a good idea to do any logic like this in a function handle. Probably more appropriate to put this in the `__init__.py` where the `DataGenerator` object get's created?

znatty22: You could get rid of config_name as a param and assume it will be read in from the environ, something like: 
```python
class DataGenerator(object):
    def __init__(self):
        self.setup(os.environ.get('FLASK_CONFIG', 'default'))
```
OR leave the config_name as an optional parameter and later on it could be passed in as a cmd line arg to the flask cmd:
 
```python
class DataGenerator(object):
    def __init__(self, config_name=None):
        if not config_name:
            config_name = os.environ.get('FLASK_CONFIG','default')
        self.setup(config_name)
```

dankolbman (approved):
Nice!

znatty22 (approved):

PR #107, opened 2018-02-07
🗃 Change uuid field to Postgres UUID

dankolbman:


znatty22 (approved):
👍

parimalak (approved):
👍

dankolbman:
Needed to write a custom migration to cast the text field into uuid. The downgrade works without custom migration because it's trivial to cast the uuid back to text.

PR #114, opened 2018-02-08
✨ Family relationship model

znatty22:
Adds the family relationship model with updates to participant model
Adds unit tests for the models

**Some Notes**
Based on discussions with team, I've modeled family relationships as an directed graph with multiple edges allowed between nodes. 

Nodes in the graph are participants and edges are family relationships
Each pair of participants can have one or more family relationships, either
from p1 --> p2 or p2 --> p1. This allows for a lot of flexibility and a wide
range of relationships between participants.

For immediate family members, the reverse relationship is automatically added (immediate family relatives of a participant are considered to be mother, father, brother, sister). For example when the following relationship is created relationship Fred -- father of -- Pebbles, the reverse relationship, Pebbles -- child of -- Fred will automatically be added to the family relationships table as well. 

There were a couple of different ways to model the family relationships and to implement the self-referential, many-to-many relationship here.

- I chose to follow sqlalchemy's directed graph example: http://docs.sqlalchemy.org/en/latest/_modules/examples/graphs/directed_graph.html


dataservice/api/participant/models.py
```diff
@@ -27,3 +27,36 @@ class Participant(db.Model, Base):
     demographic = db.relationship(Demographic, backref='participant',
                                   uselist=False, cascade="all, delete-orphan",
                                   lazy=True)
+
+    def upstream_relatives(self):
```
znatty22: I really wasn't sure how to name things here. I thought about having incoming edges to a participant be ancestors and then outgoing edges from a participant be descendants but then decided to keep things more general since we want to represent many types of familial relationships (i.e. father, mother, sister, cousin etc). 

Open to suggestions here!

dataservice/api/family_relationship/models.py
```diff
@@ -0,0 +1,83 @@
...
+
+class FamilyRelationship(db.Model, Base):
+    """
+    Represents a relationship between two family members.
+
+    The relationship table represents an undirected graph. Only one
+    relationship will exist between any two participants.
+    (P1 -> P2 is equal to P2 -> P1 and only one of these will be stored)
+
+    :param _id: Unique id assigned by RDBMS
```
dankolbman: can get rid of this now

dataservice/api/family_relationship/models.py
```diff
@@ -0,0 +1,83 @@
...
+        backref=db.backref('outgoing_family_relationships',
+                           cascade='all, delete-orphan'))
+
+    relative = db.relationship(
+        Participant,
+        primaryjoin=relative_id == Participant.kf_id,
+        backref=db.backref('incoming_family_relationships',
+                           cascade='all, delete-orphan'))
+
+    def _create_relationship_name(self, **kwargs):
```
dankolbman: It seems you're looking for a`participant_id` and a `relative_id` inside `kwargs`. It's probably better that that be stated explicitly in the handle using positional args so python will do the validation for you.

dataservice/api/family_relationship/models.py
```diff
@@ -0,0 +1,83 @@
...
+        if (participant_id is None) or (relative_id is None):
+            participant = kwargs.get('participant')
+            relative = kwargs.get('relative')
+            # Build from relationship objects in kwargs
+            if participant and relative:
+                participant_id = participant.kf_id
+                relative_id = relative.kf_id
+        # Create sorted list and concatenate using space
+        if participant_id and relative_id:
+            self._rel_name = ' '.join(sorted([participant_id, relative_id]))
```
dankolbman: So there may only be on relationship between two participants regardless of role?
What if my cousin in also my father? 
What if I later add (in your test example) p6 which is the sister of p2, and the mother of p5? Does the cousin relation become a mother relation from p2->p5?

dataservice/api/family_relationship/models.py
```diff
@@ -0,0 +1,83 @@
...
+
+    participant_id = db.Column(
+        db.String(8),
+        db.ForeignKey('participant.kf_id'), nullable=False)
+
+    relative_id = db.Column(
+        db.String(8),
+        db.ForeignKey('participant.kf_id'), nullable=False)
+
+    relationship_type = db.Column(db.Text, nullable=False)
```
dankolbman: Is there a controlled vocabulary around this?

Perhaps its easier to have a relative_role and a participant_role? Or have the roles on the participant? Maybe naming something like relation_to_participant is more appropriate?
It seems having a relationship_type on the relation itself implicitly defines this as a directed relation

dataservice/api/family_relationship/models.py
```diff
@@ -0,0 +1,83 @@
...
+        # Create sorted list and concatenate using space
+        if participant_id and relative_id:
+            self._rel_name = ' '.join(sorted([participant_id, relative_id]))
+
+    def __init__(self, **kwargs):
+        self._create_relationship_name(**kwargs)
+        super(FamilyRelationship, self).__init__(**kwargs)
+
+    def __repr__(self):
+        return '{} is {} of {}'.format(self.participant,
```
dankolbman: Can this be wrapped in `<>` to stay consistent with python?

dataservice/api/participant/models.py
```diff
@@ -27,3 +27,36 @@ class Participant(db.Model, Base):
...
+        """
+        Convenience method to get related participants upstream of self
+
+        An upstream relative is the participant pointing to self in the family
+        relationship.
+        For example for the family relationship: P1 --> P2,
+        P1 is the upstream relative of P2.
+        """
+        return [r.participant
```
dankolbman: should use yield here, otherwise you could potentially make a very large amount of queries on the db.

dataservice/api/participant/models.py
dankolbman: Is this just renaming `incoming_family_relationships`?
I suppose you probably want to capture the all of the upstream relatives too? Not just the immediate ones.

tests/family_relationship/test_family_relationship_models.py
```diff
@@ -0,0 +1,230 @@
...
+        # Get participant to delete
+        p = Participant.query.filter_by(external_id='Pebbles').one()
+
+        # Delete participant
+        db.session.delete(p)
+        db.session.commit()
+
+        # Check database
+        # No fam rels should exist, Pebbles was in every relationship
+        self.assertEqual(0, FamilyRelationship.query.count())
```
dankolbman, line 158: A check here to make sure all the participants weren't also deleted would be reassuring
```
 self.assertEqual(5, Participants.query.count())
```

znatty22: 👍

dataservice/api/participant/models.py
znatty22: 👍

dataservice/api/family_relationship/models.py
znatty22: 👍

dataservice/api/family_relationship/models.py
znatty22: No controlled vocab right now. I think in the future beyond MVP, we may have a separate role/relationship table with a controlled vocabulary.

Hmm what about `participant_to_relative_relation`? I like this because its one
field and provides clear directionality.

dataservice/api/family_relationship/models.py
znatty22: Lol, ok well I hope your cousin is not also your father, but yea I see your point there. We want to leave it flexible and open to all possibilities. I'm going to change the model to be a directed graph and allow any number of relationships/edges between two participants/nodes. 

This creates the possibility for somewhat duplicate relationships (i.e. P1 -- Father --> P2, P2 -- Son --> P1) to exist, but per our discussions earlier, we don't know how people will query or how they will input their data so it's probably best to leave it flexible.

dataservice/api/family_relationship/models.py
znatty22: This method is going to go away since I'll be changing the model to represent a directed graph. 

It was really just a helper method to help parse the kwargs from the db.Model
constructor. db.Model only allows you to create instances via kwargs. A FamilyRelationship can be created via passing in keyword args for the foreign keys (`participant_id`, `relative_id`) or by passing in keyword args for the model instances (`participant=Participant`, `relative=Participant`). So in this method I was just trying to account for this and extract the foreign keys from the various kwargs they could appear in.

dataservice/api/family_relationship/models.py
znatty22: 👍

dataservice/api/participant/models.py
znatty22: No, not exactly renaming here. `incoming_family_relationships` are incoming edges
to a participant node in the graph. An upstream relative is just the starting
node (or left) node on the incoming edge. And no this wasn't meant to be all
upstream relatives of the participant. I'm going to rename to `upstream_immediate_relatives` and `downstream_immediate_relatives`. And then maybe rename relatives to `immediate_relatives` or something like that.

And yes, I could have a method in the future that gets all relatives of a
participant as well.

znatty22:
WIP: I need to add updated ERD, mods to mock data generator, and migrations

dataservice/api/participant/models.py
```diff
@@ -27,3 +29,39 @@ class Participant(db.Model, Base):
     demographic = db.relationship(Demographic, backref='participant',
                                   uselist=False, cascade="all, delete-orphan",
                                   lazy=True)
+
+    def upstream_immediate_relatives(self):
```
dankolbman: Is this any different from `incoming_family_relationships`?

dataservice/api/participant/models.py
```diff
@@ -27,3 +29,39 @@ class Participant(db.Model, Base):
...
+
+        A downstream relative is the participant that self points to in the
+        family relationship.
+        For example for the family relationship: P1 --> P2 (where P1 is self),
+        P2 is the downstream relative of P1.
+        """
+        for r in self.outgoing_family_relationships:
+            yield r.relative
+
+    def immediate_relatives(self):
```
dankolbman, line 82: Could this be a class property?

dataservice/util/data_gen/data_generator.py
```diff
@@ -162,14 +164,38 @@ def drop_all(self):
...
 
+    def _pairwise(self, iterable):
+        """
+        Iterate over an iterable in consecutive pairs
+        """
+        a, b = tee(iterable)
+        next(b, None)
+        return zip(a, b)
+
+    def _create_family_relationships(self):
```
dankolbman, line 269: Might be nice to have more than one family with some participants having multiple first order relations.

migrations/versions/d2733490ad16_.py
```diff
@@ -0,0 +1,41 @@
+"""empty message
```
dankolbman: Can you add a message here? It makes it easier when looking at the migration history.

znatty22: 👍

dataservice/api/participant/models.py
znatty22: I think you're asking whether `immediate_relatives` could be an instance variable (list of Participant objects)? If so, I think this is typically done using association proxy, but I'm not quite sure how to implement it for a self-referential many to many. Even for a normal many-to-many, I'm still trying to figure out how to get association proxy to work. I tried using association proxy on the Study-Participant many-to-many and had issues so I abandoned it for now...

dataservice/api/participant/models.py
znatty22: Yes, an `incoming_family_relationship` is a FamilyRelationship obj representing an incoming edge to the participant (self). The upstream immediate relative is the Participant obj or starting node on the edge

dataservice/api/participant/models.py
dankolbman: I meant you could use `@property` here so it looks like a property rather than a function.

znatty22: Ah gotcha. Well this is exactly what association proxy is for and i figured out how to use it for the other many to many relationships yayyyy. So I'll see if i can get it to work with this one as well. If not then maybe I can use the property decorator as a temporary workaround

dankolbman: Hm, do you mean having this field be a relationship with secondary joins?

dankolbman:
To summarize decisions that are being made:
- relations *are* directed and we will assume that both directions will exist
- multiple relations may exist between two participants, even if they are the same
Does that sound right?

dataservice/api/family_relationship/models.py
```diff
@@ -0,0 +1,76 @@
...
+                                         self.relative)
+
+
+@event.listens_for(FamilyRelationship.participant_to_relative_relation, 'set')
+def set_reverse_relation(target, value, oldvalue, initiator):
+    """
+    Listen for set 'participant_to_relative_relation' events and
+    set the reverse relationship, 'relative_to_participant_relation' attribute
+    """
+    target.relative_to_participant_relation = REVERSE_RELS.get(value, None)
```
dankolbman, line 76: Should the  `participant_to_relative_relation` be an enum if there are strict role labels?

migrations/versions/d79a7986949e_.py
```diff
@@ -0,0 +1,46 @@
...
+    sa.Column('participant_to_relative_relation', sa.Text(), nullable=False),
+    sa.Column('relative_to_participant_relation', sa.Text(), nullable=True),
+    sa.ForeignKeyConstraint(['participant_id'], ['participant.kf_id'], ),
+    sa.ForeignKeyConstraint(['relative_id'], ['participant.kf_id'], ),
+    sa.PrimaryKeyConstraint('kf_id'),
+    sa.UniqueConstraint('kf_id'),
+    sa.UniqueConstraint('participant_id', 'relative_id', 'participant_to_relative_relation', 'relative_to_participant_relation'),
+    sa.UniqueConstraint('uuid')
+    )
+    op.create_unique_constraint(None, 'outcome', ['kf_id'])
```
dankolbman: Shouldn't be here

dataservice/api/family_relationship/models.py
znatty22: Probably eventually but for can we leave as free text? @allisonheath thoughts?

allisonheath: I would say free text for now, if it's not something recognized the reverse relationship will be null and then we can examine what relationships we have that aren't immediate family easier?

dankolbman (approved):
Looks good. I do think that there needs to be more investigation on the convenience accessors, they could likely be done in a more efficient manner.
It would also be nice to consider normalizing roles, or at least using enums.

dataservice/util/data_gen/data_generator.py
```diff
@@ -214,6 +216,9 @@ def drop_all(self):
         self.app_context.pop()
 
     def create_and_publish_all(self):
+
+
+<< << << < HEAD
```
parimalak: I think you got some merging comments

parimalak (approved):
👍

PR #117, opened 2018-02-09
📄 Update license with organization

dankolbman:
Adds the Kids first org to the license document.


allisonheath:
You actually don't change it here, it's just instructions on what to put in the separate notice file like so: https://github.com/NCI-GDC/gdcdictionary/blob/develop/NOTICE

dankolbman:
ohh TIL. I guess we should add a notice?

allisonheath:
Yup

PR #126, opened 2018-02-14
✨ Workflow model

znatty22:
Add workflow model (workflow table with workflow-genomic_file association)
Add workflow models unit tests
Add updated entity relationship diagram
Add updated mock data generator


dataservice/api/workflow/models.py
```diff
@@ -0,0 +1,71 @@
...
+    :param name: Name of workflow
+    :param version: Version of workflow
+    :param github_url: URL to repository hosted on GitHub
+    """
+    __tablename__ = 'workflow'
+
+    task_id = db.Column(db.Text())
+    name = db.Column(db.Text())
+    version = db.Column(db.Text())
+    github_url = db.Column(db.Text())
```
znatty22: This library looks pretty cool: http://sqlalchemy-utils.readthedocs.io/en/latest/_modules/sqlalchemy_utils/types/url.html

We could probably use it for custom types like URLs and maybe even do some validation at the model level. Something to look into later

dankolbman: That library seems nice if you want advanced url parsing (it uses furl objects). I think we can do some more specific validation here without the deps though:

```
@validates('github_url')
def do_validation(self, key, url):
    assert url.startswith('http{,s}://github.com/')
    return url
```

dataservice/api/workflow/models.py
```diff
@@ -0,0 +1,71 @@
...
+    :param task_id: Id of executed task
+    :param name: Name of workflow
+    :param version: Version of workflow
+    :param github_url: URL to repository hosted on GitHub
+    """
+    __tablename__ = 'workflow'
+
+    task_id = db.Column(db.Text())
+    name = db.Column(db.Text())
+    version = db.Column(db.Text())
```
dankolbman: I'd like to see this be a `github_commit`, to be as specific as possible, and to let it be better understood that we're talking about a git version. @allisonheath ?

dataservice/api/workflow/models.py
```diff
@@ -0,0 +1,71 @@
...
+
+    :param kf_id: Unique id given by the Kid's First DCC
+    :param created_at: Time of object creation
+    :param modified_at: Last time of object modification
+    :param is_input: Denotes whether the genomic file was an input to the
+    executed workflow. True = Input, False = Output
+    """
+
+    __tablename__ = 'workflow_genomic_file'
+    __table_args__ = (db.UniqueConstraint('genomic_file_id', 'workflow_id',
```
dankolbman, line 45: Can we use `unique=True` on the columns to stay consistent with how we do it in other models?

dataservice/api/workflow/models.py
znatty22: I like that better too

dataservice/api/workflow/models.py
znatty22: Isn't that for single column unique constraints though? I'm putting a unique constraint on the row essentially. Normally in a many to many you would just make each id column in the association table, a primary key, which creates a composite primary key. But in this case we already have a primary key for the table (kf_id) and I want the tuple (genomic_file_id, workflow_id, is_input) to be unique.

dataservice/util/data_gen/data_generator.py
```diff
@@ -330,9 +340,42 @@ def _create_genomic_files(self, total):
             gf_list.append(GenomicFile(**kwargs))
         return gf_list
 
+    def _create_workflows(self, total=2):
```
parimalak: can you make the total to take random number?

dataservice/api/workflow/models.py
dankolbman: Ah, ok. That's good then. Is the `is_input` necessary for the constraint? Would you want to allow a file to be both an input and an output?

dataservice/api/workflow/models.py
allisonheath: Agreed

dataservice/api/workflow/models.py
znatty22: A file can still be both an input and output with the constraint as is. You can have: 
file1, workflow1, is_input=True 
file1, workflow1, is_input=False

I think is_input is necessary for the constraint bc you wouldn't want duplicates like this: 
file1, workflow1, is_input=True 
file1, workflow1, is_input=True

dataservice/util/data_gen/data_generator.py
znatty22: 👍

dataservice/api/workflow/models.py
dankolbman: Do we need this field now?

znatty22: Ah shoot, forgot to remove it

dankolbman (approved):
lgtm 💯

parimalak (approved):
👍

PR #131, opened 2018-02-15
✨ Add fields to diagnosis Entity

parimalak:
Added diagnosis_category and tumor_location to diagnosis entity
Updated resource, tests, dummy data generation and migrations


dataservice/api/diagnosis/models.py
```diff
@@ -12,6 +12,8 @@ class Diagnosis(db.Model, Base):
     :param modified_at: Last time of object modification
     :param external_id: Name given to diagnosis by contributor
     :param diagnosis: Diagnosis of the participant
+    :param diagnosis_category: Contain very high level categorization
```
znatty22: Simplify to "High level diagnosis categorization"

dataservice/api/diagnosis/models.py
```diff
@@ -21,6 +23,11 @@ class Diagnosis(db.Model, Base):
                             doc='external id used by contributor')
     diagnosis = db.Column(db.Text(),
                           doc='the pathological diagnosis')
+    diagnosis_category = db.Column(db.Text(),
+                                   doc='contain very high level categorization'
```
znatty22: Simplify to "High level diagnosis categorization"

dataservice/util/data_gen/data_generator.py
```diff
@@ -179,6 +180,14 @@ def _outcomes_choices(self):
         self.vital_status_list = ['Alive', 'Dead', 'Not Reported']
         self.disease_related_list = [True, False, 'Not Reported']
 
+    def _get_unique_sites(self, diagnoses):
```
znatty22, line 201: Probably no need for "unique" here unless you really do intend to return the set of anatomical sites

migrations/versions/7c474ccc336b_.py
```diff
@@ -0,0 +1,33 @@
+"""
```
znatty22: This might need to change if we merge the Study model first

migrations/versions/7c474ccc336b_.py
```diff
@@ -0,0 +1,33 @@
...
+down_revision = 'c1b8fdd17637'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('diagnosis', sa.Column('diagnosis_category', sa.Text(), nullable=True))
+    op.add_column('diagnosis', sa.Column('tumor_location', sa.Text(), nullable=True))
+    op.create_unique_constraint(None, 'outcome', ['kf_id'])
```
dankolbman: I don't think this should be in here?

dataservice/util/data_gen/data_generator.py
```diff
@@ -208,11 +217,12 @@ def _create_participants(self, total):
         creates participants with samples, demographics, and diagnoses
         """
         for i in range(total):
+            diagnoses = self._create_diagnoses(
+                random.randint(self.min_samples, self.max_diagnoses))
+            self._get_unique_sites(diagnoses)
```
dankolbman: It's really difficult to know what this is doing without going back and reading through the entire code. Using global state to store intermediate results leads to a lot of headaches and counter-intuitive behavior. Looking at this scope, I cannot tell what depends on the value being calculated here and I would (wrongly) think that I could reposition this line anywhere ins the function.

Why not determine the unique sites, then return them, and pass them into the functions that need them explicitly?

Also, you could probably get away with doing this in one line so the reader doesn't have to trace through another function:
```
unique_sites = { d.tumor_location for d in diagnoses }
```

dankolbman (commented):
Model looks good. I think the migration might have gotten mixed in with some work you were doing with outcome, though.

dankolbman (approved):
Looks good!

znatty22 (approved):
👍

PR #132, opened 2018-02-15
✨ Add fields to Partcipant Entity

parimalak:
Added family_id, is_proband, consent_type to participant entity
Updated participant model, resource tests
Updated Mock data generation and migrations


dataservice/api/participant/models.py
```diff
@@ -13,12 +13,18 @@ class Participant(db.Model, Base):
 
     :param kf_id: Unique id given by the Kid's First DCC
     :param external_id: Name given to participant by contributor
+    :param family_id: Id for the participants grouped by family
+    :param is_proband: Is participant is proband of study
```
znatty22: Maybe reword to "Denotes whether participant is proband of study"

dataservice/api/participant/models.py
```diff
@@ -13,12 +13,18 @@ class Participant(db.Model, Base):
...
+    :param consent_type: Type of the consent participant belongs to
```
znatty22: Maybe reword to "Follows standard categorizations from dbGaP"

migrations/versions/f62d630e7bd7_.py
```diff
@@ -0,0 +1,35 @@
+"""
```
znatty22: This might need to change if we merge Study model first

tests/participant/test_resources.py
```diff
@@ -106,6 +112,7 @@ def test_put_participant(self):
         participant = resp['results']
         self.assertEqual(p.kf_id, participant['kf_id'])
```
znatty22, line 35: This is probably my fault from before, but these asserts need to change to:
```python
self.assertEqual(p.kf_id, participant['kf_id'])
self.assertEqual(p.external_id, body['external_id'])		          
self.assertEqual(p.consent_type, body['consent_type'])
```
The reason why is bc this test should not actually be passing, since the participant resource put method hasn't been updated with the new fields. Its only passing because the participant.<new field> and p.<new field> are probably both None.

So change these assert statements and then also make sure to update the participant resource put method. We should make that deserialization dynamic in the near future so that we don't have to manually deserialize fields into the put methods of resources.

migrations/versions/f62d630e7bd7_.py
```diff
@@ -0,0 +1,35 @@
...
+# revision identifiers, used by Alembic.
+revision = 'f62d630e7bd7'
+down_revision = 'c1b8fdd17637'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_unique_constraint(None, 'outcome', ['kf_id'])
```
dankolbman: I don't think this belongs here?

dankolbman (commented):
Probably ok for now, but we'll likely want to pull out families and consents to their own tables.

dataservice/api/participant/models.py
```diff
@@ -13,12 +13,18 @@ class Participant(db.Model, Base):
...
     :param created_at: Time of object creation
     :param modified_at: Last time of object modification
     """
     __tablename__ = "participant"
 
     external_id = db.Column(db.Text(), doc='ID used by external study')
+    family_id = db.Column(db.Text())
+    is_proband = db.Column(db.Boolean())
```
dankolbman: Hmm, I wonder if this should have a default, or at least be non-nullable?

znatty22: Yea I think it should have a default. When I had it on the StudyParticipant table, I had set the default to false because I assumed there would be more non-probands then probands? Idk if that's the correct assumption though. @allisonheath thoughts?

allisonheath: I think non-nullable makes sense. I would be a bit wary about setting a default - there are downstream consequences to it being set wrong as there's several search use cases that focus specifically on probands and their families. But if you really want a default, I think there's slightly less impact if it's set to false when it's supposed to be true versus the other way around

znatty22 (approved):

dankolbman (approved):
🎉

Issue #133, opened 2018-02-15
Implement GenomicFile resource


dankolbman:
The GenomicFile's resource needs to be syncronized with Gen3's indexd service.

The GenomicFile resource will mostly be a proxy around indexd, aside from registering the file, its relations, and any additional properties not captured by indexd.

PR #145, opened 2018-02-20
✨ Genomic file resource

dankolbman:
- [x] `GET /files` - return files from dataservice
- [x] `GET /files/{kf_id}` - Get a file by id
- [x] `POST /files` - Creates a new file in the dataservice and registers it in indexd
- [x] `PATCH /files/{kf_id}` - Updates a file and makes a new revision in indexd
- [x] `DELETE /files/{kf_id}` - Deletes a file in the dataservice and indexd
- [x]  Remove redundant fields

Addresses #133


baileyckelly:
@dankolbman   Can I connect this PR to issue #133?

dataservice/api/genomic_file/models.py
```diff
@@ -30,9 +49,166 @@ class GenomicFile(db.Model, Base):
...
+          "acls": "phs000000",
+          "kf_id": "PT_00000000"
+        }
+      }
+
+    The response upon successful registry will contain a `did` which will
+    be used as the target's uuid so that it may be joined with the indexd
+    data.
+    """
+    if current_app.config['INDEXD_URL'] is None:
```
dankolbman: Bit of a hack for running without an indexd service.

dataservice/api/genomic_file/models.py
```diff
@@ -30,9 +49,166 @@ class GenomicFile(db.Model, Base):
...
+        return
+
+    req_body = {
+        "file_name": target.file_name,
+        "size": target.size,
+        "form": "object",
+        "hashes": {
+            "md5": str(target.md5sum).replace('-', '')
```
dankolbman: Marshmallow deserialization of UUIDs has the hyphens in there. Likely a better way to do this.

dataservice/api/genomic_file/models.py
```diff
@@ -30,9 +49,166 @@ class GenomicFile(db.Model, Base):
...
+        },
+        "urls": [
+            target.file_url
```
dankolbman: Uses the `file_url` for the [urls] field in indexd. The file_url in the model should probably be removed.

dataservice/api/genomic_file/models.py
```diff
@@ -30,9 +49,166 @@ class GenomicFile(db.Model, Base):
     file_url = db.Column(db.Text())
     # See link for why md5sum should use uuid type
     # https://dba.stackexchange.com/questions/115271/what-is-the-optimal-data-type-for-an-md5-field
-    md5sum = db.Column(UUID(), unique=True)
+    md5sum = db.Column(UUID(as_uuid=True), unique=True)
```
dankolbman: Should be removed in favor of the `hashes` field stored in indexd.

dataservice/api/genomic_file/models.py
```diff
@@ -30,9 +49,166 @@ class GenomicFile(db.Model, Base):
     file_url = db.Column(db.Text())
```
dankolbman: should be removed in favor of the `urls` field in indexd.

dataservice/api/genomic_file/resources.py
```diff
@@ -0,0 +1,151 @@
...
+            get_list.yml
+          properties:
+            resource:
+              GenomicFile
+        """
+        # Get a page of the data from the model first
+        q = GenomicFile.query
+        pager = Pagination(q, after, limit)
+        for gf in pager.items:
+            gf.merge_indexd()
```
dankolbman: needs to be done async

dataservice/api/genomic_file/models.py
```diff
@@ -1,12 +1,31 @@
...
 from dataservice.api.common.model import Base, KfId
-from sqlalchemy.dialects.postgresql import UUID
 
 
 class GenomicFile(db.Model, Base):
     """
     GenomicFile entity.
 
+    A GenomicFile has fields that are stored in both the datamodel, and in
+    the indexd service. The two are linked through a common uuid.
```
znatty22: Can we change to "in the Gen3 indexd service"

dataservice/api/genomic_file/models.py
```diff
@@ -1,12 +1,31 @@
...
+
+    ## Lifetime
+
+    ### Creation
+
+    When a genomic file is created, an instance of the orm model here is
+    created, and when persisted to the database, a request is sent to gen3
+    to register the file in the indexd service. Upon successful registry
+    of the file, a response containing a did will be recieved. The GenomicFile
```
znatty22: Can we change to "a did (digital identifier)" (or baseid once we verify that is the correct identifier to use)

dataservice/api/genomic_file/models.py
```diff
@@ -30,9 +49,167 @@ class GenomicFile(db.Model, Base):
     file_url = db.Column(db.Text())
     # See link for why md5sum should use uuid type
```
znatty22: Should we remove this comment now that we're using UUID?

dataservice/api/genomic_file/models.py
```diff
@@ -30,9 +49,167 @@ class GenomicFile(db.Model, Base):
...
+@event.listens_for(GenomicFile, 'before_update')
+def update_indexd(mapper, connection, target):
+    """
+    Updates a document in indexd
+    """
+    if current_app.config['INDEXD_URL'] is None:
+        return
+
+    # Get the current revision if not already loaded
+    if target.rev is None:
```
znatty22: Seems like we would want to update the latest version if a version is not provided? Looked at the indexd Github page and wondering if GET /index/UUID returns the same thing as /index/UUID/latest?

https://github.com/uc-cdis/indexd#retrieve-the-lastest-version

dataservice/api/genomic_file/models.py
```diff
@@ -30,9 +49,167 @@ class GenomicFile(db.Model, Base):
...
+        resp = requests.get(current_app.config['INDEXD_URL']+target.uuid)
+        resp.raise_for_status()
```
znatty22: Should this be in a try/except HTTPError?

dataservice/api/genomic_file/models.py
```diff
@@ -30,9 +49,167 @@ class GenomicFile(db.Model, Base):
...
+                               target.rev)
+    resp = requests.put(url,
+                        auth=(current_app.config['INDEXD_USER'],
+                              current_app.config['INDEXD_PASS']),
+                        json=req_body)
+
+    try:
+        resp.raise_for_status()
+    except HTTPError:
+        abort(500, 'could not update genomic_file')
```
znatty22: Should we make this message more specific and say something like: `could not update genomic_file: Gen3 indexd update failed` or grab the error message from indexd if there is one (like you did in the `register_indexd`):
```
if 'error' in resp.json():
        message = '{}: {}'.format(message, resp.json()['error'])
        abort(500, message)
```

Is possible to pass back the original status_code from the indexd request?

dataservice/api/genomic_file/models.py
```diff
@@ -30,9 +49,167 @@ class GenomicFile(db.Model, Base):
...
+                               target.uuid,
+                               target.rev)
+    resp = requests.delete(url,
+                           auth=(current_app.config['INDEXD_USER'],
+                                 current_app.config['INDEXD_PASS']))
+
+    try:
+        resp.raise_for_status()
+    except HTTPError:
+        abort(500, 'could not delete genomic_file')
```
znatty22: Same comment as above in `update_indexd()`
Is possible to pass back the original status_code from the indexd request?

dataservice/api/genomic_file/models.py
```diff
@@ -30,9 +49,167 @@ class GenomicFile(db.Model, Base):
...
+                         json=req_body)
+
+    try:
+        resp.raise_for_status()
+    except HTTPError:
+        message = 'could not register genomic_file'
+        # Attach indexd error message, if there is one
+        if 'error' in resp.json():
+            message = '{}: {}'.format(message, resp.json()['error'])
+        abort(500, message)
```
znatty22: Is possible to pass back the original `status_code` from the indexd request?

dataservice/api/genomic_file/models.py
```diff
@@ -30,9 +49,167 @@ class GenomicFile(db.Model, Base):
...
+
+    resp = resp.json()
+    target.uuid = resp['did']
```
znatty22: Should this be `baseid` rather than `did`? Seems like `did` is the uuid of the version and you want the identifier of the original file object.

https://github.com/uc-cdis/indexd#create-an-index

dataservice/api/genomic_file/resources.py
```diff
@@ -0,0 +1,151 @@
...
+        # Deserialization will require this field and won't merge automatically
+        if 'sequencing_experiment_id' not in body:
+            body['sequencing_experiment_id'] = gf.sequencing_experiment_id
+        gf = GenomicFileSchema(strict=True).load(body, instance=gf).data
+
+        db.session.add(gf)
+        db.session.commit()
+
+        return GenomicFileSchema(
+            201, 'genomic_file {} updated'.format(gf.kf_id)
```
znatty22: Return `200` instead of `201` since we're not creating a new resource

dataservice/api/genomic_file/resources.py
```diff
@@ -0,0 +1,151 @@
...
+        if gf is None:
+            abort(404, 'could not find {} `{}`'
+                  .format('GenomicFile', kf_id))
+
+        # Fetch fields from indexd first
+        gf.merge_indexd()
+        # Deserialization will require this field and won't merge automatically
+        if 'sequencing_experiment_id' not in body:
+            body['sequencing_experiment_id'] = gf.sequencing_experiment_id
+        gf = GenomicFileSchema(strict=True).load(body, instance=gf).data
```
znatty22: Add `partial=True` so that a genomic file can be updated without all required fields needing to be supplied

dataservice/api/genomic_file/resources.py
```diff
@@ -0,0 +1,151 @@
...
+        ---
+        template:
+          path:
+            delete_by_id.yml
+          properties:
+            resource:
+              GenomicFile
+        """
+        try:
+            gf = GenomicFile.query.filter_by(kf_id=kf_id).one()
```
znatty22: Change to use get()

dataservice/api/genomic_file/resources.py
```diff
@@ -0,0 +1,151 @@
...
+        ---
+        template:
+          path:
+            get_by_id.yml
+          properties:
+            resource:
+              GenomicFile
+        """
+        try:
+            genomic_file = GenomicFile.query.filter_by(kf_id=kf_id).one()
```
znatty22: Change to use get()

dataservice/api/genomic_file/resources.py
znatty22: Surround with try/except ValidationError

dataservice/api/genomic_file/resources.py
```diff
@@ -0,0 +1,151 @@
...
+        gf = GenomicFile.query.get(kf_id)
+        if gf is None:
+            abort(404, 'could not find {} `{}`'
+                  .format('GenomicFile', kf_id))
+
+        # Fetch fields from indexd first
+        gf.merge_indexd()
+        # Deserialization will require this field and won't merge automatically
+        if 'sequencing_experiment_id' not in body:
+            body['sequencing_experiment_id'] = gf.sequencing_experiment_id
```
znatty22: What happens if someone wants to change the genomic file's `sequencing_experiment_id` to change its linked experiment?

dataservice/api/genomic_file/models.py
```diff
@@ -1,12 +1,31 @@
...
+from requests.exceptions import HTTPError
+from sqlalchemy import event
+from sqlalchemy.dialects.postgresql import UUID
+
 from dataservice.extensions import db
 from dataservice.api.common.model import Base, KfId
-from sqlalchemy.dialects.postgresql import UUID
 
 
 class GenomicFile(db.Model, Base):
```
znatty22: Could we move all the code that interfaces with the indexd service out of the ORM class? Maybe we could create a new package `dataservice/external` and create an `indexd` or `gen3` module or something. All of the code to talk  to indexd could go in that module. Then you would call that module in the GenomicFileAPI class. Keeps our persistence layer/code clean and completely separate from the http layer.

dataservice/api/genomic_file/models.py
dankolbman: I think there has always been a uuid here and the comment has always been around. Should probably get rid of it anyway.

dataservice/api/genomic_file/models.py
dankolbman: The revision isn't supposed to have anything to do with the version.
GET /index/UUID is different from /index/UUID/latest if there are multiple versions.

dataservice/api/genomic_file/models.py
dankolbman: I would like to, but that endpoint doesn't return any messages or json when it errors. I could throw something in there in hopes that it will one day return something.

dankolbman: We shouldn't send back anything about gen3, user's shouldn't be aware of this concept. We should log it internally, though, but I'm not yet sure how we want to take care of that.

dataservice/api/genomic_file/models.py
dankolbman: We should do that, but the problem is that you can't actually find anything about that document with only its baseid. You need to provide one of the uuids to indexd.

dataservice/api/genomic_file/models.py
dankolbman: Yeah, I wanted to do that, but wasn't sure where to put it. It turns out that their [Client](https://github.com/uc-cdis/indexclient) is open which is really what I ought to be using.

dataservice/api/genomic_file/resources.py
dankolbman: Should be ok given the if statement.

tests/conftest.py
```diff
@@ -1,29 +1,34 @@
...
+pytest_plugins = ['tests.mocks']
+
+
 @pytest.yield_fixture(scope='session')
 def app():
     yield create_app('testing')
 
 
-@pytest.yield_fixture(scope='module')
+@pytest.yield_fixture(scope='function')
```
parimalak, line 30: Why did we changed the fixture scope to `function` here? is this because app_context will be changed when we access indexd?

tests/conftest.py
```diff
@@ -77,7 +99,8 @@ def entities(client):
             'analyte_type': 'DNA',
             'concentration': 200,
             'volume': 13.99,
-            'shipment_date': str(datetime.utcnow())
+            'shipment_date': str(datetime.utcnow()),
+            'analyte_type': 'DNA'
```
parimalak: multiple `analyte_type` fields in here

tests/conftest.py
dankolbman: It's because some of the tests change the state of the db, so it needs to be rebuilt for every test.

dataservice/extensions/flask_indexd.py
```diff
@@ -0,0 +1,218 @@
...
+    def teardown(self, exception):
+        ctx = stack.top
+        if hasattr(ctx, 'indexd_session'):
+            ctx.indexd_session.close()
+
+    def get(self, record):
+        """
+        Retrieves a record from indexd
+
+        :param did: The did to fetch from indexd
```
znatty22: Should it be `:param record: The record to fetch from indexd`?

dataservice/api/genomic_file/resources.py
```diff
@@ -0,0 +1,163 @@
...
+          path:
+            get_list.yml
+          properties:
+            resource:
+              GenomicFile
+        """
+        # Get a page of the data from the model first
+        q = GenomicFile.query
+        refresh = True
+        while refresh:
```
znatty22: Why do you have to get the entire page of items again if one gf gets deleted during pagination? Could you keep another list, append the successfully merged gfs to it, and after the loop set pager.items equal to that list?

dataservice/api/genomic_file/models.py
```diff
@@ -1,37 +1,45 @@
...
 
-    :param _id: Unique id assigned by RDBMS
+    A GenomicFile has fields that are stored in both the datamodel, and in
+    the Gen3 indexd service. The two are linked through a common uuid.
+
+
     :param kf_id: Unique id given by the Kid's First DCC
     :param created_at: Time of object creation
     :param modified_at: Last time of object modification
     :param uuid: UUID assigned to file from Gen3
```
znatty22: Can we add to the docstring that this the `baseid` field?

dataservice/extensions/flask_indexd.py
```diff
@@ -0,0 +1,218 @@
...
+
+        :param record: A file-like object
+        """
+        # If running in dev mode, dont call indexd
+        if self.url is None:
+            record.latest_did = str(uuid.uuid4())
+            return record
+
+        # Fetch rev for the did
+        if record.rev is None:
```
znatty22: Don't you always want to get the latest revision # (regardless of whether its None) before you do the update?

dataservice/extensions/flask_indexd.py
```diff
@@ -0,0 +1,218 @@
...
+            "form": "object",
+            "file_name": record.file_name,
+            "size": record.size,
+            "hashes": record.hashes,
+            "urls": record.urls,
+            "metadata": record._metadata
+        }
+
+        # Update the file on indexd
+        url = '{}{}?rev={}'.format(self.url, record.latest_did, record.rev)
```
znatty22, line 163: Might be that I don't understand the indexd API docs, but I'm wondering which uuid is supposed to be used here in the url. It seems that it should be POST /indexd/{baseid} bc you're creating a new version (new did) for the original record identified by the baseid.

dataservice/api/common/model.py
```diff
@@ -34,6 +38,111 @@ def kf_id(cls):
...
+
+        If the document matching this object's latest_did cannot be found in
+        indexd, remove the object from the database
+
+        :returns: This object, if merge was successful, otherwise None
+        """
+        try:
+            return indexd.get(self)
+        except RecordNotFound as err:
+            self.was_deleted = True
```
znatty22, line 97: Are you keeping this flag so that you don't always have to send a request to check if an indexd file was deleted?

dataservice/api/genomic_file/resources.py
```diff
@@ -0,0 +1,163 @@
...
+              GenomicFile
+        """
+        body = request.json or {}
+        gf = GenomicFile.query.get(kf_id)
+        if gf is None:
+            abort(404, 'could not find {} `{}`'
+                  .format('genomic_file', kf_id))
+
+        # Fetch fields from indexd first
+        gf.merge_indexd()
```
znatty22: Should we return a 404 if the the merge results in a non-existent genomic file in indexd?

dataservice/extensions/flask_indexd.py
dankolbman: Yes, it's confusing. All UUIDs referenced in the docs are the `did`. The `baseid` cannot be used anywhere. I've put in a request to make it so that you can look up documents by `baseid`

dataservice/api/common/model.py
dankolbman: This is a hack so we don't get stuck in an endless loop in the before_delete event.

dataservice/api/genomic_file/resources.py
dankolbman: I don't. It's just easier to do it this way.

znatty22: But isn't that doing a lot more work unnecessarily? Pagination will take a lot longer if we always restart the getting of gfs for that page from indexd (when a gf is deleted in indexd)

dataservice/api/errors.py
```diff
@@ -13,6 +14,7 @@
 
 def http_error(e):
     """ Handles all HTTPExceptions """
+    db.session.rollback()
```
znatty22, line 17: What if there is an http error that didn't involve a db transaction?

dataservice/api/genomic_file/resources.py
```diff
@@ -0,0 +1,191 @@
...
+        ---
+        template:
+          path:
+            get_list.yml
+          properties:
+            resource:
+              GenomicFile
+        """
+        # Get a page of the data from the model first
+        q = GenomicFile.query
```
znatty22, line 35: How about something like this to condense a bit: 
```python
# Get a page of the data from the model first
        q = GenomicFile.query
        pager = Pagination(q, after, limit)

        keep = []
        while (pager.total > 0) and (len(keep) < limit):
            # Move the cursor ahead to the last valid file
            after = keep[-1].created_at if len(keep) > 0 else after
            # Number of results needed to fulfill the original limit
            remain = limit - len(keep)
            # Paginate
            pager = Pagination(q, after, remain)
            for gf in pager.items:
                merged = gf.merge_indexd()
                if merged is not None:
                    keep.append(gf)

        # Replace original page's items with new list of valid files
        pager.items = keep

        return (GenomicFileSchema(many=True)
                .jsonify(pager))
```

dataservice/api/errors.py
dankolbman: Nothing will happen

dataservice/api/genomic_file/resources.py
dankolbman: Better, but goes into an infinite loop if there are < 10 files

znatty22 (approved):
💯 🥇 Nice job!

parimalak (approved):

PR #149, opened 2018-02-21
✨ Study file model

parimalak:
Added StudyFile model with tests
Added Migrations and mock up data
Updated ERD with StudyFile model


dankolbman (approved):
✨ ✨ ✨

znatty22 (approved):
👍

PR #158, opened 2018-03-01
🐛 Fix self links

dankolbman:
- Fixes formatting of the `self` link where a datetime was being used instead of a timestamp.
- Fixes issue where the timestamp being returned for the current page did not have a small enough delta causing earlier results to be  returned.


znatty22 (approved):
👍

parimalak (approved):

PR #173, opened 2018-03-07
✨ ⚡️Change PUT to PATCH in all resource APIs and minor updates

znatty22:
For participant, demographic, diagnosis, and sample - change the PUT http method in the resource class to PATCH http method. 

- Previous PUT implementation was actually a PATCH implementation. PATCH allows for partial update of resources. PUT replaces the entire resource and will be implemented in the near future. 

For participant, demographic, diagnosis, and sample - change method of retrieval of resource to use SQLAlchemy get() which is optimized for retrieval by primary key.

### DIFF
--- a/dataservice/__init__.py
+++ b/dataservice/__init__.py
@@ -121,6 +121,7 @@ def register_error_handlers(app):
     from dataservice.api import errors
     app.register_error_handler(HTTPException, errors.http_error)
     app.register_error_handler(IntegrityError, errors.integrity_error)
+    app.register_error_handler(405, errors.http_error)
     app.register_error_handler(404, errors.http_error)
     app.register_error_handler(400, errors.http_error)
 
--- a/dataservice/__init__.py
+++ b/dataservice/__init__.py
@@ -23,7 +23,6 @@ from dataservice.api.study_file.models import StudyFile
 from config import config
 
 from sqlalchemy.exc import IntegrityError
-from werkzeug.exceptions import HTTPException
 
 
 def create_app(config_name):
@@ -119,11 +118,10 @@ def register_error_handlers(app):
     NB: Exceptions to be handled must be imported in the head of this module
     """
     from dataservice.api import errors
-    app.register_error_handler(HTTPException, errors.http_error)
     app.register_error_handler(IntegrityError, errors.integrity_error)
-    app.register_error_handler(405, errors.http_error)
-    app.register_error_handler(404, errors.http_error)
-    app.register_error_handler(400, errors.http_error)
+    from werkzeug.exceptions import default_exceptions
+    for ex in default_exceptions:
+        app.register_error_handler(ex, errors.http_error)
 
 
 def register_blueprints(app):
--- a/dataservice/api/demographic/resources.py
+++ b/dataservice/api/demographic/resources.py
@@ -82,17 +82,12 @@ class DemographicAPI(CRUDView):
             resource:
               Demographic
         """
-        # Get all
-        if kf_id is None:
-            dm = Demographic.query.all()
-            return DemographicSchema(many=True).jsonify(dm)
-        # Get one
-        else:
-            dm = Demographic.query.get(kf_id)
-            if dm is None:
-                abort(404, 'could not find {} `{}`'
-                      .format('demographic', kf_id))
-            return DemographicSchema().jsonify(dm)
+        dm = Demographic.query.get(kf_id)
+        if dm is None:
+            abort(404, 'could not find {} `{}`'
+                  .format('demographic', kf_id))
+
+        return DemographicSchema().jsonify(dm)
 
     def patch(self, kf_id):
         """
--- a/dataservice/api/demographic/resources.py
+++ b/dataservice/api/demographic/resources.py
@@ -98,11 +98,9 @@ class DemographicAPI(CRUDView):
                       .format('demographic', kf_id))
             return DemographicSchema().jsonify(d)
 
-    def put(self, kf_id):
+    def patch(self, kf_id):
         """
-        Update existing demographic
-
-        Update an existing demographic given a Kids First id
+        Update an existing demographic. Allows partial update of resource
         ---
         template:
           path:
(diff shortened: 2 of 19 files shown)


tests/demographic/test_demographic_resources.py
```diff
@@ -190,117 +194,78 @@ def test_get_all(self):
...
+        for k, v in body.items():
+            self.assertEqual(v, getattr(dm, k))
+
+        # Content - Check remaining fields are unchanged
+        unchanged_keys = (set(demographic.keys()) -
+                          set(body.keys()))
+        for k in unchanged_keys:
+            val = getattr(dm, k)
+            if isinstance(val, datetime):
+                d = val.replace(tzinfo=tz.tzutc())
```
dankolbman, line 233: Does this add the timezone to the dt or does it change an existing one?

tests/demographic/test_demographic_resources.py
```diff
@@ -369,3 +334,12 @@ def _create_save_to_db(self):
         kwargs['kf_id'] = d.kf_id
 
         return kwargs
+
+    def _test_response_content(self, resp, status_code):
```
dankolbman: If we really want this, it could be part of the base test class. However, it's been removed from all the other tests and replaced by the general api tests.

tests/test_api.py
```diff
@@ -68,40 +75,61 @@ def test_status_format(self, client, endpoint, method):
...
-        assert (field not in body['results']
-                or body['results'][field] != 'test')
+        for field in fields:
+            assert (field not in body['results']
+                    or body['results'][field] != 'test')
 
-    @pytest.mark.parametrize('endpoint,field', [
-        ('/participants', 'blah'),
-        ('/samples', 'blah')
+    @pytest.mark.parametrize('endpoint, method, field', [
```
dankolbman: We can also double parametrize things to reduce duplication. We could probably just use 'blah' for all the tests too. I think something like this will work:

```
@pytest.mark.parametrize('method', ['POST', 'PATCH'])
@pytest.mark.parametrize('endopint,', ['/participants', '/demographics' ... ])
def test_unkown_field(self, client, entities, endpoint, method):
    pass
```

dataservice/__init__.py
```diff
@@ -119,6 +119,7 @@ def register_error_handlers(app):
     from dataservice.api import errors
     app.register_error_handler(HTTPException, errors.http_error)
     app.register_error_handler(IntegrityError, errors.integrity_error)
+    app.register_error_handler(405, errors.http_error)
```
dankolbman: when does this happen?

dataservice/api/demographic/resources.py
```diff
@@ -83,26 +82,21 @@ def get(self, kf_id):
             resource:
               Demographic
         """
-
         # Get all
         if kf_id is None:
```
dankolbman: This logic was supposed to be removed when we changed to the two-view method.

tests/demographic/test_demographic_resources.py
znatty22: It adds a timezone to `val` since `val` doesn't have one and then returns a new datetime obj. 

https://stackoverflow.com/questions/7065164/how-to-make-an-unaware-datetime-timezone-aware-in-python

tests/demographic/test_demographic_resources.py
znatty22: Mkay 👍

tests/test_api.py
znatty22: Mkay 👍

dataservice/__init__.py
znatty22: If you try sending a `PUT` request (or any other HTTP method we don't support). It looks like 
` app.register_error_handler(HTTPException, errors.http_error)` isn't catching all http exceptions. Maybe we need to investigate that further

znatty22: This seems to work though?
```
from werkzeug.exceptions import default_exceptions
for ex in default_exceptions:
        app.register_error_handler(ex, errors.http_error)
```

We could probably remove the other explicit HTTP error registrations (i.e. 400, 404, etc) if we use this. 

@dankolbman  I vaguely remember you said you tried this or something similar and it didn't work or you had issues with it?

dankolbman: Yeah, it has something to do with the scoping on the exception. I forget what exactly I changed to make it work, but it had to do with where the HTTPException was imported.

parimalak (approved):
👍

dataservice/__init__.py
```diff
@@ -119,10 +118,10 @@ def register_error_handlers(app):
     NB: Exceptions to be handled must be imported in the head of this module
     """
     from dataservice.api import errors
-    app.register_error_handler(HTTPException, errors.http_error)
     app.register_error_handler(IntegrityError, errors.integrity_error)
-    app.register_error_handler(404, errors.http_error)
-    app.register_error_handler(400, errors.http_error)
+    from werkzeug.exceptions import default_exceptions
+    for ex in default_exceptions:
+        app.register_error_handler(ex, errors.http_error)
```
dankolbman, line 124: No luck with HTTPException? It looks like default_exceptions is just all the subclasses of HTTPException, so you would think it would work.

tests/diagnosis/test_diagnosis_resources.py
```diff
@@ -12,6 +14,8 @@
 DIAGNOSES_URL = 'api.diagnoses'
 DIAGNOSES_LIST_URL = 'api.diagnoses_list'
 
+from pprint import pprint
```
dankolbman: unused?

dataservice/__init__.py
znatty22: I think we have to go with this approach for now. Doesn't look like the fix for the issue with HTTPException is in Flask 0.12. Its slated for Flask milestone 1.0 which isn't complete yet

https://github.com/pallets/flask/pull/2314

dankolbman (approved):
💯

Issue #161, opened 2018-03-05
Study endpoint


dankolbman:
Write an `/studies` endpoint that supports basic CRUD operations:

- [ ] `GET /studies` - return studies from dataservice
- [ ] `GET /studies/{kf_id}` - Get an study by id
- [ ] `POST /studies` - Creates a study in the dataservice
- [ ] `PATCH /studies/{kf_id}` - Partially updates a study by id
- [ ] `DELETE /studies/{kf_id}` - Deletes a study by id

PR #175, opened 2018-03-08
✨ Study resource

dankolbman:
Adds endpoints for the study resource.

Addresses #161

### DIFF
--- a/dataservice/api/study/resources.py
+++ b/dataservice/api/study/resources.py
@@ -1,0 +1,134 @@
+from flask import abort, request
+from marshmallow import ValidationError
+
+from dataservice.extensions import db
+from dataservice.api.common.pagination import paginated, Pagination
+from dataservice.api.study.models import Study
+from dataservice.api.study.schemas import StudySchema
+from dataservice.api.common.views import CRUDView
+
+
+class StudyListAPI(CRUDView):
+    """
+    Study API
+    """
+    endpoint = 'studies_list'
+    rule = '/studies'
+    schemas = {'Study': StudySchema}
+
+    @paginated
+    def get(self, after, limit):
+        """
+        Get a paginated studies
+        ---
+        template:
+          path:
+            get_list.yml
+          properties:
+            resource:
+              Study
+        """
+        q = Study.query
+
+        return (StudySchema(many=True)
+                .jsonify(Pagination(q, after, limit)))
+
+    def post(self):
+        """
+        Create a new study
+        ---
+        template:
+          path:
+            new_resource.yml
+          properties:
+            resource:
+              Study
+        """
+        try:
+            st = StudySchema(strict=True).load(request.json).data
+        except ValidationError as err:
+            abort(400, 'could not create study: {}'.format(err.messages))
+
+        db.session.add(st)
+        db.session.commit()
+        return StudySchema(
+            201, 'study {} created'.format(st.kf_id)
+        ).jsonify(st), 201
+
+
+class StudyAPI(CRUDView):
+    """
+    Study API
+    """
+    endpoint = 'studies'
+    rule = '/studies/<string:kf_id>'
+    schemas = {'Study': StudySchema}
+
+    def get(self, kf_id):
+        """
+        Get a study by id
+        ---
+        template:
+          path:
+            get_by_id.yml
+          properties:
+            resource:
+              Study
+        """
+        st = Study.query.get(kf_id)
+        if st is None:
+            abort(404, 'could not find {} `{}`'
+                  .format('study', kf_id))
+        return StudySchema().jsonify(st)
+
+    def patch(self, kf_id):
+        """
+        Update an existing study. Allows partial update of resource
+        ---
+        template:
+          path:
+            update_by_id.yml
+          properties:
+            resource:
+              Study
+        """
+        body = request.json
+        st = Study.query.get(kf_id)
+        if st is None:
+            abort(404, 'could not find {} `{}`'
+                  .format('study', kf_id))
+
+        try:
+            st = (StudySchema(strict=True).load(body, instance=st,
+                                                partial=True).data)
+        except ValidationError as err:
+            abort(400, 'could not update study: {}'.format(err.messages))
+
+        db.session.add(st)
+        db.session.commit()
+
+        return StudySchema(
+            200, 'study {} updated'.format(st.kf_id)
+        ).jsonify(st), 200
+
+    def delete(self, kf_id):
+        """
+        Delete study by id
+        ---
+        template:
+          path:
+            delete_by_id.yml
+          properties:
+            resource:
+              Study
+        """
+        st = Study.query.get(kf_id)
+        if st is None:
+            abort(404, 'could not find {} `{}`'.format('study', kf_id))
+
+        db.session.delete(st)
+        db.session.commit()
+
+        return StudySchema(
+            200, 'study {} deleted'.format(st.kf_id)
+        ).jsonify(st), 200
--- a/tests/study/test_study_resources.py
+++ b/tests/study/test_study_resources.py
@@ -1,0 +1,147 @@
+import json
+
+from flask import url_for
+
+from dataservice.api.study.models import Study
+from tests.utils import FlaskTestCase
+
+STUDY_URL = 'api.studies'
+STUDY_LIST_URL = 'api.studies_list'
+
+
+class StudyTest(FlaskTestCase):
+    '''
+    Test study api endopoints
+    '''
+
+    def test_post_study(self):
+        '''
+        Test creating a new study
+        '''
+        response = self._make_study(external_id='TEST')
+        resp = json.loads(response.data.decode('utf-8'))
+
+        self.assertEqual(response.status_code, 201)
+
+        self.assertIn('study', resp['_status']['message'])
+        self.assertIn('created', resp['_status']['message'])
+        self.assertNotIn('_id', resp['results'])
+
+        s = Study.query.first()
+        study = resp['results']
+        self.assertEqual(s.kf_id, study['kf_id'])
+        self.assertEqual(s.external_id, study['external_id'])
+
+    def test_get_study(self):
+        '''
+        Test retrieving a study by id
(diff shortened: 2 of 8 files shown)


dataservice/api/study/resources.py
```diff
@@ -0,0 +1,132 @@
...
+              Study
+        """
+        body = request.json
+        try:
+            st = Study.query.filter_by(kf_id=kf_id).one()
+        except NoResultFound:
+            abort(404, 'could not find {} `{}`'
+                  .format('Study', kf_id))
+
+        st = StudySchema(strict=True).load(body, instance=st).data
```
znatty22: Couple requested changes here.. 
1. Add the `partial=True` kw arg to the load so that marshmallow allows you to deserialize without all required fields. With PATCH we want to allow someone to update one or more fields without having to include all required fields.
2. Surround the load with a try/except ValidationError. 

This is what i did for participant: 

```
       # Partial update - validate but allow missing required fields
        try:
            p = ParticipantSchema(strict=True).load(body, instance=p,
                                                    partial=True).data
        except ValidationError as err:
            abort(400, 'could not update participant: {}'.format(err.messages))
```

dataservice/api/study/resources.py
```diff
@@ -0,0 +1,132 @@
...
+        try:
+            st = Study.query.filter_by(kf_id=kf_id).one()
+        except NoResultFound:
+            abort(404, 'could not find {} `{}`'
+                  .format('Study', kf_id))
+        return StudySchema().jsonify(st)
+
+    def patch(self, kf_id):
+        """
+        Update an existing study
```
znatty22: Can we update the comment to something like: 

`Update an existing study. Allows for partial update of resource.`

tests/study/test_study_resources.py
```diff
@@ -0,0 +1,129 @@
...
+                                           kf_id=kf_id),
+                                   headers=self._api_headers())
+        resp = json.loads(response.data.decode('utf-8'))
+        self.assertEqual(response.status_code, 200)
+        self._test_response_content(resp, 200)
+
+        study = resp['results']
+        self.assertEqual(kf_id, study['kf_id'])
+
+    def test_patch_study(self):
```
znatty22, line 52: Probably a good idea to check the database as well. Make sure that only the fields in the body were modified and others remained the same.

dataservice/api/study/resources.py
```diff
@@ -0,0 +1,132 @@
...
+        ---
+        template:
+          path:
+            get_by_id.yml
+          properties:
+            resource:
+              Study
+        """
+        try:
+            st = Study.query.filter_by(kf_id=kf_id).one()
```
znatty22: Should we change this to use `.get` now that `kf_id` is the primary key? If we decide to make that change, we'll also have to change the try/except NoResultFound to a if else statement to check if `st` is None. I think get() returns None if obj is not found

dataservice/api/study/resources.py
```diff
@@ -0,0 +1,132 @@
...
+        template:
+          path:
+            update_by_id.yml
+          properties:
+            resource:
+              Study
+        """
+        body = request.json
+        try:
+            st = Study.query.filter_by(kf_id=kf_id).one()
```
znatty22: Similar to above comment, should we change this to use `.get` now that `kf_id` is the primary key?

dataservice/api/study/resources.py
```diff
@@ -0,0 +1,132 @@
...
+        ---
+        template:
+          path:
+            delete_by_id.yml
+          properties:
+            resource:
+              Study
+        """
+        try:
+            st = Study.query.filter_by(kf_id=kf_id).one()
```
znatty22: Similar to above comment, should we change this to use `.get` now that `kf_id` is the primary key?

tests/test_api.py
```diff
@@ -85,7 +94,8 @@ def test_read_only(self, client, entities, endpoint, field):
 
     @pytest.mark.parametrize('endpoint,field', [
         ('/participants', 'blah'),
-        ('/samples', 'blah')
+        ('/samples', 'blah'),
+        ('/studies', 'blah'),
     ])
     def test_unknown_field(self, client, entities, endpoint, field):
```
znatty22: I changed this method in #173 to provide an HTTP method as input so that we could perform this test for both POST and PATCH (and later PUT). We may want to merge #173 before this one and then you can update this test for study

dataservice/api/study/resources.py
```diff
@@ -0,0 +1,132 @@
...
+        except NoResultFound:
+            abort(404, 'could not find {} `{}`'
+                  .format('Study', kf_id))
+
+        st = StudySchema(strict=True).load(body, instance=st).data
+        db.session.commit()
+
+        return StudySchema(
```
znatty22: Should return `200` instead of `201` since we are not creating a new resource

tests/study/test_study_resources.py
```diff
@@ -0,0 +1,129 @@
...
+        db.session.commit()
+
+        body = {
+            'external_id': 'new_id'
+        }
+        response = self.client.patch(url_for(STUDY_URL,
+                                           kf_id=kf_id),
+                                   headers=self._api_headers(),
+                                   data=json.dumps(body))
+        self.assertEqual(response.status_code, 201)
```
znatty22: Rember to change to check for `200` after updating patch method to return `200` instead of `201`

dataservice/api/study/resources.py
dankolbman: yeah, I'd like to use `get_or_404()`, but you're not able to pass a custom message to it. If we could do that, these 5 lines would be reduced down to one tidy line. We could consider hacking `get_or_404()` to take a message arg or not having a specific entity type returned in the not found response.

dataservice/api/study/resources.py
dankolbman: hmm I thought we had a general `ValidationError` handler, maybe we should add one?

dataservice/api/study/resources.py
```diff
@@ -0,0 +1,130 @@
...
+          path:
+            get_by_id.yml
+          properties:
+            resource:
+              Study
+        """
+        st = Study.query.get(kf_id)
+        if st is None:
+            abort(404, 'could not find {} `{}`'
+                  .format('Study', kf_id))
```
znatty22: Change 'Study' to 'study' to keep all error msgs standard

dataservice/api/study/resources.py
```diff
@@ -0,0 +1,130 @@
...
+            update_by_id.yml
+          properties:
+            resource:
+              Study
+        """
+        body = request.json
+        st = Study.query.get(kf_id)
+        if st is None:
+            abort(404, 'could not find {} `{}`'
+                  .format('Study', kf_id))
```
znatty22: Change Study to study to keep all error msgs standard

dataservice/api/study/resources.py
```diff
@@ -0,0 +1,130 @@
...
+        template:
+          path:
+            delete_by_id.yml
+          properties:
+            resource:
+              Study
+        """
+        st = Study.query.get(kf_id)
+        if st is None:
+            abort(404, 'could not find {} `{}`'.format('Study', kf_id))
```
znatty22: Change Study to study to keep all error msgs standard

tests/test_api.py
```diff
@@ -41,7 +43,11 @@ def test_status_codes(self, client, endpoint, method, status_code):
         ('/participants', 'GET', 'success'),
         ('/participants/123', 'GET', 'could not find Participant `123`'),
         ('/participants/123', 'PUT', 'could not find Participant `123`'),
-        ('/participants/123', 'DELETE', 'could not find Participant `123`')
+        ('/participants/123', 'DELETE', 'could not find Participant `123`'),
+        ('/studies', 'GET', 'success'),
+        ('/studies/123', 'GET', 'could not find Study `123`'),
```
znatty22: Remember to change from Study to study after updating resource class

tests/study/test_study_resources.py
```diff
@@ -0,0 +1,161 @@
...
+        body = {
+            'external_id': external_id,
+            'version': '1.0'
+        }
+        response = self.client.post(url_for(STUDY_LIST_URL),
+                                    headers=self._api_headers(),
+                                    data=json.dumps(body))
+        return response
+
+    def _test_response_content(self, resp, status_code):
```
znatty22: Remove method and calls to it

tests/study/test_study_resources.py
```diff
@@ -0,0 +1,161 @@
+import json
+
+from flask import url_for
+
+from dataservice.extensions import db
```
znatty22: Remove unused import: `dataservice.extensions.db`

dataservice/api/study/resources.py
```diff
@@ -0,0 +1,134 @@
...
+        st = Study.query.get(kf_id)
+        if st is None:
+            abort(404, 'could not find {} `{}`'
+                  .format('study', kf_id))
+
+        try:
+            st = (StudySchema(strict=True).load(body, instance=st,
+                                                partial=True).data)
+        except ValidationError as err:
+            abort(400, 'could not update participant: {}'.format(err.messages))
```
znatty22: Prob copy/paste error, change `could not update participant` to `could not update study`

dataservice/api/study/resources.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        st = Study.query.get(kf_id)
+        if st is None:
+            abort(404, 'could not find {} `{}`'
+                  .format('study', kf_id))
+        return StudySchema().jsonify(st)
+
+    def patch(self, kf_id):
+        """
+        Update fields in a study and preserve existing fields.
```
znatty22: Nitpicking but can we be consistent w the docstrings for crud resources? All the other resources have patch docstring like: `Update an existing demographic. Allows partial update of resource`. If you don't like above, then lets at least create an issue to standardize docstrings for the resource crud methods

znatty22 (approved):
👍

PR #179, opened 2018-03-12
🔥 Remove SQLite configuration

dankolbman:
Seeing as we are more invested in postgres specific features now, the sqlite code is no longer needed.


znatty22 (approved):
👍

parimalak (approved):
👍

Issue #163, opened 2018-03-05
Phenotype endpoint


parimalak:
Write Phenotype endpoint that supports basic CRUD operations:
- [x] - GET /phenotypes - return phenotype from dataservice
- [x] -  GET /phenotype/{kf_id} - Get an phenotype by id
- [x] -  POST /phenotypes - Creates a phenotype in the dataservice
- [x] -  PATCH /phenotypes/{kf_id} - Partially updates an phenotype by id
- [x] -  DELETE /phenotypes/{kf_id} - Deletes an phenotype

PR #180, opened 2018-03-12
✨ Add phenotype endpoint

parimalak:


dataservice/api/phenotype/resources.py
```diff
@@ -0,0 +1,162 @@
...
+            p = Phenotype.query.filter_by(kf_id=kf_id).one()
+        # Not found in database
+        except NoResultFound:
+            abort(404, 'could not find {} `{}`'
+                  .format('phenotype', kf_id))
+        return PhenotypeSchema().jsonify(p)
+
+    def patch(self, kf_id):
+        """
+        Partially update existing phenotype
```
znatty22: Can we change it to: "Update an existing phenotype. Allows partial update of resource"

dataservice/api/phenotype/resources.py
```diff
@@ -0,0 +1,162 @@
...
+        template:
+          path:
+            get_by_id.yml
+          properties:
+            resource:
+              Phenotype
+        """
+        # Get one
+        try:
+            p = Phenotype.query.filter_by(kf_id=kf_id).one()
```
znatty22: Let's start using get() instead of filter_by now that kf_id is the primary key. See #173 for example

dataservice/api/phenotype/resources.py
```diff
@@ -0,0 +1,162 @@
...
+          path:
+            update_by_id.yml
+          properties:
+            resource:
+              Phenotype
+        """
+        body = request.json
+        try:
+            # Check if phenotype exists
+            p1 = Phenotype.query.filter_by(kf_id=kf_id).one()
```
znatty22: Let's start using get() instead of filter_by now that kf_id is the primary key. See #173 for example

dataservice/api/phenotype/resources.py
```diff
@@ -0,0 +1,162 @@
...
+            abort(404, 'could not find {} `{}`'.format('phenotype', kf_id))
+
+        # Validation only
+        try:
+            p = PhenotypeSchema(strict=True).load(body).data
+        # Request body not valid
+        except ValidationError as e:
+            abort(400, 'could not update phenotype: {}'.format(e.messages))
+
+        # Deserialize
```
znatty22: I think you want to remove this block here. Looks like this is left over from PUT implementation since this block of code will overwrite all of the properties on phenotype.

dataservice/api/phenotype/resources.py
```diff
@@ -0,0 +1,162 @@
...
+        try:
+            # Check if phenotype exists
+            p1 = Phenotype.query.filter_by(kf_id=kf_id).one()
+        # Not found in database
+        except NoResultFound:
+            abort(404, 'could not find {} `{}`'.format('phenotype', kf_id))
+
+        # Validation only
+        try:
+            p = PhenotypeSchema(strict=True).load(body).data
```
znatty22: Use `partial=True` here so that a user can update a phenotype without needing to supply all required fields. See #173 for example

dataservice/api/phenotype/resources.py
```diff
@@ -0,0 +1,162 @@
...
+          path:
+            delete_by_id.yml
+          properties:
+            resource:
+              Phenotype
+        """
+
+        # Check if phenotype exists
+        try:
+            p = Phenotype.query.filter_by(kf_id=kf_id).one()
```
znatty22: Let's start using get() instead of filter_by now that kf_id is the primary key. See #173 for example

tests/phenotype/test_phenotype_resources.py
```diff
@@ -0,0 +1,395 @@
...
+                         kwargs['participant_id'])
+        self.assertEqual(phenotype['phenotype'], kwargs['phenotype'])
+        self.assertEqual(kwargs['hpo_id'],
+                         phenotype['hpo_id'])
+        self.assertEqual(kwargs['observed'], phenotype['observed'])
+        self.assertEqual(phenotype['age_at_event_days'],
+                         kwargs['age_at_event_days'])
+        self.assertEqual(participant_id, kwargs['participant_id'])
+
+    def test_get_not_found(self):
```
znatty22: We can remove this test now that we test for it in test_api.py

tests/phenotype/test_phenotype_resources.py
```diff
@@ -0,0 +1,395 @@
...
+        kwargs = self._create_save_to_db()
+
+        response = self.client.get(url_for(PHENOTYPES_LIST_URL),
+                                   headers=self._api_headers())
+        self.assertEqual(response.status_code, 200)
+        response = json.loads(response.data.decode("utf-8"))
+        content = response.get('results')
+        self.assertEqual(len(content), 1)
+
+    def test_patch(self):
```
znatty22, line 120: After you update the patch() method in the phenotype resource, this test will need to be updated based on expected PATCH behavior. It should test that only the properties supplied in body were updated on the phenotype and the rest of the properties remained unchanged since PATCH allows for partial update

tests/phenotype/test_phenotype_resources.py
```diff
@@ -0,0 +1,395 @@
...
+        self.assertEqual(kwargs['kf_id'], phenotype['kf_id'])
+        # Fields that should be None since they were not in patch request body
+        self.assertIs(None, phenotype['age_at_event_days'])
+        self.assertIs(None, phenotype['observed'])
+        # Fields that should be updated w values
+        self.assertEqual(body['phenotype'], phenotype['phenotype'])
+        self.assertEqual(body['hpo_id'],
+                         phenotype['hpo_id'])
+
+    def test_patch_not_found(self):
```
znatty22: We can remove this test since this is tested in test_api.py

dataservice/api/phenotype/resources.py
parimalak: 👍

tests/phenotype/test_phenotype_resources.py
```diff
@@ -0,0 +1,395 @@
...
+        self.assertIn(message, response['_status']['message'])
+        # Check field values
+        ph = Phenotype.query.first()
+        self.assertIs(ph, None)
+
+    def test_post_invalid_age(self):
+        """
+        Test create phenotype with bad input data
+
+        Participant with participant_id does not exist in db
```
dankolbman: description needs to be updated?

tests/phenotype/test_phenotype_resources.py
```diff
@@ -0,0 +1,395 @@
...
+        # Check response body
+        response = json.loads(response.data.decode("utf-8"))
+        # Check error message
+        message = 'could not update phenotype'
+        self.assertIn(message, response['_status']['message'])
+        # Check field values
+        p = Phenotype.query.first()
+        self.assertEqual(p.phenotype, kwargs['phenotype'])
+
+    def test_delete(self):
```
dankolbman: This is basically the same test as the one below without the checking of response code? We could probably get rid of it.

tests/phenotype/test_phenotype_resources.py
```diff
@@ -0,0 +1,395 @@
...
+        # Check response body
+        response = json.loads(response.data.decode("utf-8"))
+        # Check error message
+        message = '"{}" does not exist'.format(kwargs['participant_id'])
+        self.assertIn(message, response['_status']['message'])
+        # Check field values
+        ph = Phenotype.query.first()
+        self.assertIs(ph, None)
+
+    def test_post_multiple(self):
```
dankolbman: Could be more concise by calling `_create_save_to_db()` twice.

dataservice/api/phenotype/schemas.py
```diff
@@ -0,0 +1,29 @@
...
+from marshmallow import validates, ValidationError
+
+from dataservice.api.phenotype.models import Phenotype
+from dataservice.api.common.schemas import BaseSchema
+from dataservice.api.common.validation import validate_age
+from dataservice.extensions import ma
+
+
+class PhenotypeSchema(BaseSchema):
+    # Should not have to do this, since participant_id is part of the
```
dankolbman: We should consider getting rid of this comment. If we want to do something about the problem, we should find a different way to track it.

tests/phenotype/test_phenotype_resources.py
parimalak: 👍

tests/phenotype/test_phenotype_resources.py
parimalak: using different parameters for creating the second phenotype

dataservice/api/phenotype/resources.py
```diff
@@ -0,0 +1,154 @@
...
+        Allows partial update of resource
+        ---
+        template:
+          path:
+            update_by_id.yml
+          properties:
+            resource:
+              Phenotype
+        """
+        # Partial update - validate but allow missing required fields
```
znatty22: This comment is confusing here. It originally was above line 114 (the try catch). You could move it there or remove it

dataservice/api/phenotype/resources.py
```diff
@@ -0,0 +1,154 @@
...
+        # Validation only
+        try:
+            p = PhenotypeSchema(strict=True).load(body, instance=p,
+                                                  partial=True).data
+        # Request body not valid
+        except ValidationError as e:
+            abort(400, 'could not update phenotype: {}'.format(e.messages))
+
+        # Save to database
+        db.session.add(p)
```
znatty22, line 122: Shouldn't need to add to the session. Can remove this line

tests/phenotype/test_phenotype_resources.py
dankolbman: Still needs updating?

dankolbman (approved):
💯

dataservice/api/phenotype/resources.py
znatty22: Sorry I was wrong about this!

tests/test_api.py
```diff
@@ -105,6 +112,7 @@ def test_status_format(self, client, endpoint, method):
         ('/samples', 'PATCH', ['created_at', 'modified_at']),
         ('/studies', 'POST', ['created_at', 'modified_at']),
         ('/studies', 'PATCH', ['created_at', 'modified_at']),
+        ('/phenotypes', 'PATCH', ['created_at', 'modified_at']),
```
znatty22: What about `('/phenotypes', 'POST', ['created_at', 'modified_at']),`

znatty22 (commented):
I think there are a couple missing tests in test_api.py:

```python
 def test_relations(self, client, entities, resource, field)
def test_bad_foreign_key(self, client, entities, endpoint, method, field)
```

znatty22 (commented):
Looks like you're still missing one test in test_api.py:
```python
def test_bad_foreign_key(self, client, entities, endpoint, method, field)
```

znatty22 (approved):
👍

parimalak:
closes #163 and resolves #163

PR #183, opened 2018-03-14
♻️ Fix self links

### LINKED ISSUE #172: Change self link format or implement query parameters
Currently all of the self links are endpoints which have `kf_id` as a query parameter like so: 
```
"self": "/participants?kf_id=PT_7MJ05MNZ"
```
The dataservice does not currently support query parameters except for pagination. This means none of the self links will retrieve the expected resource. In the example above, that endpoint will not retrieve participant PT_7MJ05MNZ. Instead the result will be the default, which is the first page of participants.

To solve this, we have a couple options: 

1) Change self links to this format: 
```
"self": "/participants/PT_7MJ05MNZ"
```
or 2) Keep the self links the way they are and implement query parameters

### LINKED ISSUE #182: Standardize _links
It appears that there is currently inconsistent formatting of what links are included in the `_links` parameter in the response bodies. Some entities return only `self` while others return `collection` and  also links to foreign relations.

We should standardize what links are being returned and make sure to add tests to ensure that we remain consistent.

dankolbman:
Added `resource_url` and `collection_url` for each schema.

We may be able to get rid of the manual `_link` definitions by putt logic into the common schema or by extending marshmallow-sqlalchemy.

Fixes #172, resolves #182

### DIFF
--- a/dataservice/api/demographic/schemas.py
+++ b/dataservice/api/demographic/schemas.py
@@ -14,11 +14,12 @@ class DemographicSchema(BaseSchema):
                                load_only=True)
 
     class Meta(BaseSchema.Meta):
-        resource_url = 'api.demographics_list'
+        resource_url = 'api.demographics'
+        collection_url = 'api.demographics_list'
         model = Demographic
 
     _links = ma.Hyperlinks({
-        'self': ma.URLFor('api.demographics', kf_id='<kf_id>'),
-        'collection': ma.URLFor('api.demographics_list'),
+        'self': ma.URLFor(Meta.resource_url, kf_id='<kf_id>'),
+        'collection': ma.URLFor(Meta.collection_url),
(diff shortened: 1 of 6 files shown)


dataservice/api/demographic/schemas.py
```diff
@@ -14,11 +14,12 @@ class DemographicSchema(BaseSchema):
                                load_only=True)
 
     class Meta(BaseSchema.Meta):
-        resource_url = 'api.demographics_list'
+        resource_url = 'api.demographics'
+        collection_url = 'api.demographics_list'
```
dankolbman, line 18: It would be nice if we could figure out how to define these as part of the resource registration process.

znatty22 (approved):
👍

parimalak (approved):
👍

PR #186, opened 2018-03-16
📝 Docs branding

dankolbman:
Adds shiny logo to the docs and info on how to connect to the api.
Adds docstrings to the models so that the fields have descriptions in the docs.

### DIFF
--- a/dataservice/api/README.md
+++ b/dataservice/api/README.md
@@ -1,4 +1,60 @@
-Primary API for interacting with the Kid's First Data Model.
+<center>
+![Kids First](/logo)
+</center>
+
+Welcome to the developer documentation for the Kids First Dataservice API!
+
+The dataservice is the primary hub for retrieval and creation of Kids First
+data. It exposes functionality using a familiar REST interface over HTTP
+requests.
+
+# Connect to the API
+
+The API is currently only available internally inside the Kids First cloud
+environments.
+
+## curl
+
+Curl is the quickest way to communicate with the api from a unix shell
+
+Get the system status:
+```bash
+curl -H "Content-Type: application/json" http://kf-api-dataservice-qa.kids-first.io/status
+```
+
+Create a new study with the name 'my study':
+```bash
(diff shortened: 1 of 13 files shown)


dataservice/api/README.md
```diff
@@ -1,4 +1,58 @@
...
+
+Curl is the quickest way to communicate with the api from a unix shell
+
+```bash
+curl -H "Content-Type: application/json" http://kf-api-dataservice-qa.kids-first.io/status
+```
+
+To give input to the API:
+```bash
+curl -H "Content-Type: application/json" http://kf-api-dataservice-qa.kids-first.io/status -d '{ "external_id": "my identifier" }'
```
znatty22: Since we can't post to /status how about something a little more realistic like :

```bash
curl -X POST -H "Content-Type: application/json" http://kf-api-dataservice-qa.kids-first.io/studies -d '{ "external_id": "my identifier" }'
```

dataservice/api/README.md
```diff
@@ -1,4 +1,58 @@
...
+
+## curl
+
+Curl is the quickest way to communicate with the api from a unix shell
+
+```bash
+curl -H "Content-Type: application/json" http://kf-api-dataservice-qa.kids-first.io/status
+```
+
+To give input to the API:
```
znatty22: `To give input to the API (i.e. create a new study)`

dataservice/api/README.md
```diff
@@ -1,4 +1,58 @@
...
+```python
+import requests
+
+body = {
+  "external_id": "my identifier"
+}
+
+resp = requests.post('http://kf-api-dataservice-qa.kids-first.io/participants',
```
znatty22: Following along with the above example, maybe change to `/studies`. Another reason to use studies endpoint is study entity is pretty simple and `external_id`  is the only required field right now so the POST studies example will work whereas this POST participants will break right now

znatty22 (commented):
For all of the doc params for `age_at_event_days`, could we use this phrasing: 
"Age at the time of event occurred in number of days since birth."
i.e. Age at the time of outcome occurred in number of days since birth.

znatty22 (approved):
👍

PR #187, opened 2018-03-16
♻️Suggested refactor of genomic_file resource

znatty22:
Take a look and let me know what you think! All tests pass except the delete ones that were failing before.

### DIFF
--- a/dataservice/api/genomic_file/resources.py
+++ b/dataservice/api/genomic_file/resources.py
@@ -1,8 +1,7 @@
 from flask import abort, request
-from sqlalchemy.orm.exc import NoResultFound
 from marshmallow import ValidationError
 
-from dataservice.extensions import db
+from dataservice.extensions import db, indexd
 from dataservice.api.common.pagination import paginated, Pagination
 from dataservice.api.genomic_file.models import GenomicFile
 from dataservice.api.genomic_file.schemas import GenomicFileSchema
@@ -35,8 +34,10 @@ class GenomicFileListAPI(CRUDView):
         # Get a page of the data from the model first
         q = GenomicFile.query
         pager = Pagination(q, after, limit)
+
+        # For each genomic_file, get indexd properties and merge into obj
         for gf in pager.items:
-            gf.merge_indexd()
+            indexd.merge_properties(gf.uuid, gf)
 
         return (GenomicFileSchema(many=True)
                 .jsonify(pager))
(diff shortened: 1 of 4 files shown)


dataservice/extensions/flask_indexd.py
```diff
@@ -35,6 +35,52 @@ def teardown(self, exception):
         if hasattr(ctx, 'indexd_session'):
             ctx.indexd_session.close()
 
+    def merge_properties(self, uuid, record):
```
dankolbman: Having this be a method of the indexd client doesn't make much sense to me. If you're going to mutate the state of an object's properties, it would seem more appropriate to have the object be responsible for doing it.

dataservice/api/genomic_file/resources.py
```diff
@@ -147,6 +155,9 @@ def delete(self, kf_id):
         if gf is None:
             abort(404, 'could not find {} `{}`'.format('GenomicFile', kf_id))
 
+        # Delete genomic file from indexd
+        indexd.delete_record(gf)
```
dankolbman: I'd like to see these operations masked behind the GenomicFile object so that we maintain the same interface among all of our entities. It also helps in the case that we ever want to work with GenomicFiles in another setting where we will likely expect to do the same operations and would likely need to duplicate this logic there as well.

znatty22: Hmm well I think the ORM layer shouldn’t really be making http calls. We’re probably going to have other instances where we’re integrating with 3rd party APIs to (i.e. Cavatica to get workflow info). I would think that we would want to encapsulate the code for interfacing with a 3rd party API in its own module. And keep it out of the the ORM class  even if there is a possibility of some duplicate code in the future

dataservice/extensions/flask_indexd.py
znatty22: Could we maybe break down the flask_indexd module and put the methods (merge_properties, register_record, update_record, delete_record) that really modify GenomicFile in the resource class? That class already seems to modify GenomicFile (i.e. in update method). I left them in flask_indexd because I saw the other methods already access the GenomicFile object quite a bit anyway.

dataservice/api/genomic_file/resources.py
dankolbman: This isn't really in the ORM layer. The indexd extension sort of creates it's own ORM which maps a subset of the object's fields to indexd and hooks into SQLalchemy's events to syncronize changes. If we want more functionality out of the indexd, it may be better to abstract it out into a client, but seeing as it's rather limited, it seems like it's not yet appropriate.

PR #84, opened 2018-02-02
✨ Aliquot resource

znatty22:
Add aliquot resource with crud methods and unit tests

* Note that nested de/serialization does not work right now. For example, cannot create/update/read a sample with embedded aliquots. Also cannot create/update/read aliquots with embedded sequencing experiments. This functionality will implemented in near future.

PR #189, opened 2018-03-16
✨Add outcome endpoint

parimalak:
Added outcome resources and schemas
Updated tests for outcome resource


tests/outcome/test_outcome_resources.py
```diff
@@ -0,0 +1,333 @@
...
+        self.assertEqual(kwargs['kf_id'], outcome['kf_id'])
+        
+        # Fields that should be updated w values
+        self.assertEqual(body['vital_status'],
+                         outcome['vital_status'])
+
+
+    def test_patch_bad_input(self):
+        """
+        Test updating an existing participant with invalid input
```
dankolbman: Needs to be updated.

tests/outcome/test_outcome_resources.py
```diff
@@ -0,0 +1,333 @@
...
+from dataservice.api.outcome.models import Outcome
+from dataservice.api.participant.models import Participant
+from dataservice.api.study.models import Study
+from tests.utils import FlaskTestCase
+
+OUTCOMES_URL = 'api.outcomes'
+OUTCOMES_LIST_URL = 'api.outcomes_list'
+
+
+class OutcomeTest(FlaskTestCase):
```
znatty22, line 16: We should merge #84 first and then you can probably remove a lot of these tests (bad input, missing req params) and put them in test_api.py.

dankolbman (approved):
Looks good!

tests/test_api.py
```diff
@@ -105,6 +112,7 @@ def test_status_format(self, client, endpoint, method):
         ('/samples', 'PATCH', ['created_at', 'modified_at']),
         ('/studies', 'POST', ['created_at', 'modified_at']),
         ('/studies', 'PATCH', ['created_at', 'modified_at']),
+        ('/outcomes', 'PATCH', ['created_at', 'modified_at']),
```
znatty22: What about `('/outcomes', 'POST', ['created_at', 'modified_at'])`

znatty22 (commented):
I think you're missing a relations test for `outcome` in test_api.py. In:

```python
def test_relations(self, client, entities, resource, field)
```

tests/test_api.py
parimalak: 👍

znatty22 (approved):
👍

Issue #128, opened 2018-02-14
Create initial participant_alias model


baileyckelly:
Per conversations in this morning's design meeting, we need to create a table called participant_alias and an ORM model to represent it called ParticipantAlias. 

This is an association table for the many-to-many relationship between Participant and Participant. It stores of pairs of participant ids to help track duplicate participants.

PR #196, opened 2018-03-22
✨ Alias Group Model (for participant aliases)

znatty22:
- [x] Add an alias group table to store aliases of participants
- [x] Add unit tests
- [x] Update erd
- [x] Update data generator
- [x] Update migrations

Addresses #128


dataservice/api/participant/models.py
```diff
@@ -50,6 +59,15 @@ class Participant(db.Model, Base):
...
                                  backref=db.backref('participants',
                                                     lazy=True))
+
+    aliases = db.relationship(
+        'Participant', secondary=participant_aliases,
+        primaryjoin=(
+            'participant_aliases.c.participant_id == Participant.kf_id'),
+        secondaryjoin=(
+            'participant_aliases.c.alias_id == Participant.kf_id'),
+        backref=db.backref('aliased'))
```
dankolbman: does it need a backref if both directions are added directly?

dataservice/api/participant/models.py
```diff
@@ -87,5 +105,30 @@ def immediate_relatives(self):
         return chain(self.upstream_immediate_relatives(),
                      self.downstream_immediate_relatives())
 
+    def add_alias(self, alias):
```
dankolbman: Seems like there could be a better way to do this, maybe through an association proxy creator? It would be very easy to wrongly append to the aliases attribute, as you would any other model.

tests/participant/test_alias_models.py
```diff
@@ -0,0 +1,97 @@
...
+            'Fred1': {
+                'external_id': 'Fred1',
+                'is_proband': False
+            },
+            'Fred2': {
+                'external_id': 'Fred2',
+                'is_proband': False
+            }
+        }
+        for _, kwargs in alias_data.items():
```
dankolbman: Is this an accurate portrayal of a use case? Fred1 and Fred2 are aliases of Fred, but Fred1 is not an alias of Fred2.

tests/participant/test_alias_models.py
```diff
@@ -0,0 +1,97 @@
...
+        db.session.commit()
+
+        # Check db
+        fred = Participant.query.filter_by(external_id='Fred').first()
+        ext_ids = {alias.external_id for alias in fred.aliases}
+        self.assertIn('SuperFred', ext_ids)
+        self.assertNotIn(external_id, ext_ids)
+
+    def test_delete(self):
+        """
```
dankolbman: Test removal of one of the participants?

dataservice/api/participant/models.py
znatty22: Probably not

dataservice/api/participant/models.py
znatty22: Yeeea, I initially tried to listen for append and remove events on aliases, and then add or remove the reverse alias depending on the fired event. I was getting recursion max out errors so I don't think those events work the way I thought. I'll try looking into them again and I can also look into assoc proxy

tests/participant/test_alias_models.py
znatty22: It would be if I implemented a method to retrieve all aliases for a participant through a dfs. Right now aliases are basically modeled as an undirected graph, so I could probably just change the way aliases are retrieved and return all aliases of a participant through dfs instead of just the immediate aliases. I don't think we'd want to store alias links for all participants that are aliased to each other bc that could quickly grow large.

dataservice/api/participant/models.py
```diff
@@ -87,5 +110,86 @@ def immediate_relatives(self):
...
+
+            # Merge smaller alias group with larger alias group
+            # aka, change all participants' alias_group_id in the smaller group
+            # to be the alias_group_id of the larger group
+            _criteria = Participant.alias_group_id == smaller_alias_group.kf_id
+            for p in db.session.query(Participant).filter(_criteria):
+                p.alias_group = larger_alias_group
+
+            # Delete old alias group
+            db.session.delete(smaller_alias_group)
```
znatty22: In the future we may want to consider doing a "soft delete" of an alias group and for each participant record the alias groups that the participant has belonged to (another column with a list of alias group ids). This would help us back track if we need to (i.e. if we roll back to a previous release or need to compare a previous release to the current release etc)

dataservice/api/participant/models.py
```diff
@@ -87,5 +110,86 @@ def immediate_relatives(self):
...
+
+    @property
+    def aliases(self):
```
dankolbman, line 187: I would think this could be replaced by a relationship:
```
 aliases = db.relationship("Participant",
                           primaryjoin='and_(participant.c.alias_group_id==foreign(participant.c.alias_group_id), participant.c.kf_id!=foreign(participant.c.kf_id))',
                           post_update=True,
                           uselist=True)
```

But it seems like deletes are cascading through participants somehow, or something the session is stale after deletes.

If that can't be made to work, it would be nearly as good to just do the same join here to make sure there's only one query going to the db.

dataservice/api/participant/models.py
```diff
@@ -87,5 +110,86 @@ def immediate_relatives(self):
         return chain(self.upstream_immediate_relatives(),
                      self.downstream_immediate_relatives())
 
+    def add_alias(self, pt):
```
dankolbman, line 122: Seems like this could be a constructor in an association proxy, though I wasn't able to get it working.

dataservice/api/participant/models.py
znatty22: Why do we need a join? This is just a simple select where query to the db:
`Participant.query.filter_by(alias_group_id=self.alias_group_id).all()`, results in: 

```sql
SELECT participant.uuid AS participant_uuid, participant.created_at AS participant_created_at, participant.modified_at AS participant_modified_at, participant.external_id AS participant_external_id, participant.family_id AS participant_family_id, participant.is_proband AS participant_is_proband, participant.consent_type AS participant_consent_type, participant.study_id AS participant_study_id, participant.alias_group_id AS participant_alias_group_id, participant.kf_id AS participant_kf_id
FROM participant
WHERE participant.alias_group_id = %(alias_group_id_1)s
```

dataservice/api/participant/models.py
znatty22: Hmm yea if we had an `aliases` association proxy, then this method would be the creator for it and then I migtht be able to provide a getset_factory to return the result of my @property aliases method. I don't know what function I would use for the setter in getset_factory though. 

I'd prefer to stick with what I have for now though. It seems pretty straightforward and I've found association proxies to be a little tricky sometimes

dataservice/api/participant/models.py
dankolbman: Sorry, not a join, but have the second condition where `self.kf_id != Participant.kf_id`. this way, it can be done in 1-2 lines.

Also, I believe I thought that using `yield` was more effective here, but it doesn't make a difference as `all()` will just load everything. So it's probably easier to just return the result rather than a generator so `len()` works out of the box.

znatty22: Ah I see, ok i'll change that query

dankolbman (approved):
Can you update erd before merge?

parimalak (approved):
Looks good. Can we rename the PR to Alias Group model?

PR #197, opened 2018-03-23
✨Study file resource

parimalak:
Added resources for study_file endpoint
Added tests for study_file endpoint

Resolves #197

### DIFF
--- a/dataservice/api/study_file/schemas.py
+++ b/dataservice/api/study_file/schemas.py
@@ -17,5 +17,5 @@ class StudyFileSchema(BaseSchema):
     _links = ma.Hyperlinks({
         'self': ma.URLFor(Meta.resource_url, kf_id='<kf_id>'),
         'collection': ma.URLFor(Meta.collection_url),
-        'study_files': ma.URLFor('api.studies', kf_id='<study_id>')
+        'study': ma.URLFor('api.studies', kf_id='<study_id>')
     })
--- a/dataservice/api/study_file/schemas.py
+++ b/dataservice/api/study_file/schemas.py
@@ -1,0 +1,21 @@
+from marshmallow_sqlalchemy import field_for
+
+from dataservice.api.study_file.models import StudyFile
+from dataservice.api.common.schemas import BaseSchema
+from dataservice.extensions import ma
+
+
+class StudyFileSchema(BaseSchema):
+    study_id = field_for(StudyFile, 'study_id', required=True,
+                         load_only=True)
+
+    class Meta(BaseSchema.Meta):
+        model = StudyFile
+        resource_url = 'api.study_files'
+        collection_url = 'api.study_files_list'
+
+    _links = ma.Hyperlinks({
+        'self': ma.URLFor(Meta.resource_url, kf_id='<kf_id>'),
+        'collection': ma.URLFor(Meta.collection_url),
+        'study_files': ma.URLFor('api.studies', kf_id='<study_id>')
+    })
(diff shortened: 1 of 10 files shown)


dataservice/api/study_file/schemas.py
```diff
@@ -0,0 +1,21 @@
...
+
+    class Meta(BaseSchema.Meta):
+        model = StudyFile
+        resource_url = 'api.study_files'
+        collection_url = 'api.study_files_list'
+
+    _links = ma.Hyperlinks({
+        'self': ma.URLFor(Meta.resource_url, kf_id='<kf_id>'),
+        'collection': ma.URLFor(Meta.collection_url),
+        'study_files': ma.URLFor('api.studies', kf_id='<study_id>')
```
dankolbman: I think this should just be `study`?

tests/study_file/test_study_file_resources.py
```diff
@@ -0,0 +1,153 @@
...
+from dataservice.api.study_file.models import StudyFile
+from dataservice.api.study.models import Study
+from tests.utils import FlaskTestCase
+
+STUDY_FILE_URL = 'api.study_files'
+STUDY_FILE_LIST_URL = 'api.study_files_list'
+
+
+class StudyFileTest(FlaskTestCase):
+    '''
```
dankolbman: Small nit, but we should stay consistent with the use of double quotes for docstrings.

dataservice/api/study_file/schemas.py
znatty22: Should be singular:
```python
'study_file': ma.URLFor('api.studies', kf_id='<study_id>')
```

tests/study_file/test_study_file_resources.py
```diff
@@ -0,0 +1,153 @@
...
+
+        resp = json.loads(response.data.decode('utf-8'))
+        self.assertIn('study_file', resp['_status']['message'])
+        self.assertIn('updated', resp['_status']['message'])
+
+        study_file = resp['results']
+        self.assertEqual(study_file['kf_id'], kf_id)
+        self.assertEqual(study_file['file_name'], body['file_name'])
+
+    def test_patch_study_file_no_required_field(self):
```
znatty22: Could probably remove this test since it's essentially the same as `test_patch`

tests/test_pagination.py
```diff
@@ -31,6 +32,12 @@ def participants(client):
             db.session.add(s)
         db.session.commit()
 
+        # Add study files for studies
+        for _ in range(102):
```
znatty22: Could probably create the study files and link the files to a study in the loop where you're creating each study

dataservice/api/study_file/schemas.py
parimalak: i dont know the usecase for having 'study' information in _links but, it is easier to get all the study_files from study as each study_file belong to one study. i am saying this because we can remove this line altogether from _links similar to what we did for investigator and study resources.

tests/study_file/test_study_file_resources.py
parimalak: yep i used study tests as reference i am gonna change for other resources as well👍

tests/study_file/test_study_file_resources.py
parimalak: 👍

dankolbman (approved):
💯

znatty22 (approved):
👍

PR #199, opened 2018-03-27
👷Added the verbose checkout declaration

alubneuski:
🌳The multibranch scm checkout modified the way it clone the repo. It ommits the tags. In order for tags to be discovered, Jenkinsfile needs to have more complicated checkout process. There is no way to do it in DSL during the project creation.


dankolbman:
No emoji in commit 👎

PR #213, opened 2018-03-29
⚡️ Only load relation's kf_id

dankolbman:
Increases aliquot and sample paginated endpoints by only loading the `kf_id` of related entities.
Also decreases of sequential page scanning by roughly the same amount.

Baseline:
```
wrk -t1 -c12 -d60s 'http://localhost:5000/samples?limit=100'
Running 1m test @ http://localhost:5000/samples?limit=100
  1 threads and 12 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.10s   556.88ms   1.84s    66.67%
    Req/Sec     3.09      0.49     5.00     85.15%
  202 requests in 1.00m, 12.68MB read
  Socket errors: connect 0, read 0, write 0, timeout 196
Requests/sec:      3.37
Transfer/sec:    216.35KB
```

With `load_only` kf_id on relations:
```
wrk -t1 -c12 -d60s 'http://localhost:5000/samples?limit=100'
Running 1m test @ http://localhost:5000/samples?limit=100
  1 threads and 12 connections
Bn  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   877.35ms   91.07ms   1.18s    87.10%
    Req/Sec    13.28      5.01    20.00     60.76%
  814 requests in 1.00m, 51.10MB read
Requests/sec:     13.55
Transfer/sec:      0.85MB
```

### DIFF
--- a/dataservice/api/aliquot/resources.py
+++ b/dataservice/api/aliquot/resources.py
@@ -1,4 +1,5 @@
 from flask import abort, request
+from sqlalchemy.orm import joinedload
 from marshmallow import ValidationError
 
 from dataservice.extensions import db
@@ -29,7 +30,11 @@ class AliquotListAPI(CRUDView):
             resource:
               Aliquot
         """
-        q = Aliquot.query
+        q = (Aliquot.query
+                    .options(
+                        joinedload('sequencing_experiments')
+                        .load_only('kf_id')
+                     ))
 
         return (AliquotSchema(many=True)
                 .jsonify(Pagination(q, after, limit)))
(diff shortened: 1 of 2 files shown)


znatty22 (commented):
This PR might become OBE pretty soon, since we're getting rid of sample and aliquot. Do we still want it? 
If its highly desired for the data tracker team since they're the ones using paginated samples/aliquots, then we can keep it.  

One other comment regardless of above is, should we also do this for paginated `/participants` as well? If so, might as well include it in this PR

dankolbman:
Yeah, this pr may not be relevant anymore for the aliquot+samples. We should go back and inspect all of the entities with fks, though, which is just about everything. I will close this, but we can still use it for reference.

Issue #192, opened 2018-03-19
Add 400+ Responses to kf-dataservice Documentation


znatty22:
Currently our API docs have example 200 responses included for each endpoint. We should add example 400/404 responses to each of the endpoints.

dankolbman:
Should be pretty easy to define new [templates](https://github.com/kids-first/kf-api-dataservice/tree/master/dataservice/api/templates) for each of the errors.
The templates use jinja so you can easily extend the base responses that are there now with other external files.

PR #223, opened 2018-04-05
📝 Update API docs to include 400, 404 responses

znatty22:
Addresses #192

### DIFF
--- a/dataservice/api/common/schemas.py
+++ b/dataservice/api/common/schemas.py
@@ -108,12 +108,27 @@ class ErrorSchema(Schema):
     """ Handles HTTPException marshalling """
 
     class Meta:
-        fields = ('description', 'code')
+        fields = ("description", "code")
 
     @post_dump(pass_many=False)
     def wrap_envelope(self, data):
         return {'_status': {'message': data['description'],
                             'code': data['code']}}
+
+
+def error_response_generator(resource_name, status_code):
+    class ErrorResponseSchema(Schema):
+        _examples = {
+            404: "could not find {} <kf_id>".format(resource_name),
+            400: ("could not update {}".format(resource_name) +
+                  " {'<field>': ['Not a valid <expected type>.']}")
+        }
+        description = fields.String(description='status message',
+                                    example=_examples.get(status_code))
+        code = fields.Integer(description='HTTP response code',
+                              example=status_code)
+
+    return ErrorResponseSchema
 
 
 def response_generator(schema):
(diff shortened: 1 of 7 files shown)


dataservice/utils.py
```diff
@@ -13,3 +14,14 @@ def iterate_pairwise(iterable):
     a, b = tee(iterable)
     next(b, None)
     return zip(a, b)
+
+
+def to_snake_case(camel_case_str):
```
dankolbman, line 19: Our resources are hyphenated, not underscored. eg `/genomic-files`, `/family-relationships`

znatty22: Yea our endpoints are hyphenated but the error messages contain our entity names which are snake cased. Example: 
```json
{
    "_status": {
        "code": 400,
        "message": "could not create family_relationship: {'participant_id': ['Missing data for required field.'], 'participant_to_relative_relation': ['Missing data for required field.'], 'relative_id': ['Missing data for required field.']}"
    }
}
```

dankolbman: Right, but the snake case is still being used for the resource names in the pagination

dataservice/api/common/schemas.py
```diff
@@ -97,12 +112,13 @@ class RespSchema(Schema):
...
 
-def paginated_generator(schema):
+def paginated_generator(resource_name, schema):
     class PaginatedSchema(Schema):
         _status = fields.Dict(example={'message': 'success', 'code': 200})
-        _links = fields.Dict(example={'next': '/participants?after=1519402953',
-                                      'self': '/participants?after=1519402952'
-                                      })
+        _links = fields.Dict(
+            example={'next': '/{}?after=1519402953'.format(resource_name),
```
dankolbman: might want to do a `resource_name.replace('_', '-')` here

znatty22: Oh i see, 👍

dataservice/api/common/schemas.py
```diff
@@ -81,14 +81,29 @@ class ErrorSchema(Schema):
...
-        fields = ('description', 'code')
+        fields = ("description", "code")
 
     @post_dump(pass_many=False)
     def wrap_envelope(self, data):
         return {'_status': {'message': data['description'],
                             'code': data['code']}}
 
 
+def error_response_generator(resource_name, status_code):
```
dankolbman, line 119: I'm not sure we should regenerate an identical schema for every error type. It seems to violate the idea of polymorphism, which swagger supports.

dataservice/api/common/views.py
```diff
@@ -38,12 +41,26 @@ def register_spec(cls, spec):
...
             for name, schema in c.schemas.items():
+                # Entity response schemas
                 spec.definition(name, schema=schema)
                 ResponseSchema = response_generator(schema)
                 spec.definition(name + 'Response', schema=ResponseSchema)
-                PaginatedSchema = paginated_generator(schema)
-                spec.definition(name + 'Paginated', schema=PaginatedSchema)
+
+                # Pagination schemas
+                if c.__name__.endswith('ListAPI'):
```
dankolbman: Maybe better to use the url rule?

znatty22: Wasn't sure if I should do that or use the ListAPI postfix bc the url rules for paginated endpoints might not always be /\<entity\>. For example we might have /genomic-files?sequence_center=<kf_id> soon. Figured looking for the 'ListAPI' postfix might be more explicit and a good assumption for paginated endpoints?

dankolbman: I don't think the query params show up in url rules?

dankolbman (approved):

parimalak (approved):
👍

Issue #202, opened 2018-03-28
Merging sample and aliquot as biospecimen


parimalak:
- [x]  Update `concentration` field to Float for aliquot(BioSpecimen) model
- [x]  Add `uberon_id`, `external_sample_id`, `external_aliquot_id` fields to sample(BioSpecimen)
- [x] Merge Sample and aliquot as BioSpecimen
- [x]  Link Biospecimen to GenomicFile
- [x] Remove connection between aliquot(BioSpecimen) and sequencing_experiment and link sequencing_experiment to genomic_file 
- [ ] update genomic-file link in sequencing_experiment schema

PR #228, opened 2018-04-10
✨Biospecimen model

parimalak:
- [x]  Update `concentration` field to float for aliquot(biospecimen) model
- [x]  Rename `concentration` and `volume` to  `concentration_mg_per_ml`, `volume_ml`
- [x]  Add `uberon_id`, `external_sample_id`, `external_aliquot_id` fields to sample(biospecimen)
- [x] Merge sample and aliquot as biospecimen
- [x]  Link biospecimen to genomic_file
- [x] Remove connection between aliquot(biospecimen) and sequencing_experiment and link sequencing_experiment to genomic_file 
- [ ] update genomic-file link in sequencing_experiment schema
- [x] update erd and migrations
- [x] update data generation script

Addresses #202


tests/biospecimen/test_biospecimen_resources.py
```diff
@@ -0,0 +1,394 @@
...
+from flask import url_for
+
+from dataservice.extensions import db
+from dataservice.api.common import id_service
+from dataservice.api.biospecimen.models import Biospecimen
+from dataservice.api.participant.models import Participant
+from dataservice.api.study.models import Study
+from tests.utils import FlaskTestCase
+
+SAMPLES_URL = 'api.biospecimens'
```
dankolbman: Needs to be renamed?

dataservice/api/biospecimen/models.py
```diff
@@ -0,0 +1,78 @@
...
+                          doc='The ID of the term from Uber-anatomy ontology'
+                          'which represents harmonized anatomical ontologies')
+    participant_id = db.Column(KfId(),
+                               db.ForeignKey('participant.kf_id'),
+                               nullable=False,
+                               doc='The kf_id of the biospecimen\'s donor')
+    genomic_files = db.relationship(GenomicFile,
+                                    cascade="all, delete-orphan",
+                                    backref=db.backref(
+                                        'genomic_files', lazy=True),
```
znatty22: The backref should be named `biospecimen` to represent the many to one going from genomic file to biospecimen (this will give us GenomicFile.biospecimen).

dataservice/api/biospecimen/models.py
```diff
@@ -0,0 +1,78 @@
...
+           submitted tumor biospecimen
+    :param age_at_event_days: Age at the time biospecimen was
+            acquired, expressed in number of days since birth
+    :param tumor_descriptor: The kind of disease present in the tumor
+           specimen as related to a specific timepoint
+    :param shipment_origin : The origin of the shipment
+    :param shipment_destination: The destination of the shipment
+    :param analyte_type: Text term that represents the kind of molecular
+           specimen analyte
+    :param concentration: The concentration of an analyte or aliquot extracted
```
znatty22: What did we decide in terms of standard units for concentration and volume? From the X01 data so far I see that concentration is usually in micrograms per microliter and volume is in microliters. 
@allisonheath

dataservice/api/biospecimen/models.py
```diff
@@ -0,0 +1,78 @@
...
+                                      'number of days since birth.')
+    tumor_descriptor = db.Column(db.Text(),
+                                 doc='Disease present in the biospecimen')
+    shipment_origin = db.Column(db.Text(),
+                                doc='The original site of the aliquot')
+    shipment_destination = db.Column(db.Text(),
+                                     doc='The site recieving the aliquot')
+    analyte_type = db.Column(db.Text(), nullable=False,
+                             doc='The molecular description of the aliquot')
+    concentration = db.Column(db.Float(),
```
znatty22: I think we decided that names of measured quantity fields should include the units. So I think we might need to change to the following: 

`concentration_ng_per_ul`
`volume_ul`

dataservice/api/sequencing_experiment/models.py
```diff
@@ -44,8 +44,6 @@ class SequencingExperiment(db.Model, Base):
     mean_depth = db.Column(db.Float())
     total_reads = db.Column(db.Integer())
     mean_read_length = db.Column(db.Float())
-    aliquot_id = db.Column(KfId(),
-                           db.ForeignKey('aliquot.kf_id'), nullable=False)
     genomic_files = db.relationship(GenomicFile,
                                     cascade="all, delete-orphan",
                                     backref=db.backref(
```
znatty22: The name of the backref should be `sequencing_experiment`.

tests/biospecimen/test_biospecimen_models.py
```diff
@@ -0,0 +1,308 @@
...
+        aliquot_id) = self.create_participant_biospecimen()
+        s = Biospecimen.query.filter_by(external_sample_id=sample_id).\
+            one_or_none()
+        p = Participant.query.filter_by(
+            external_id=participant_id).one_or_none()
+        self.assertEqual(p.kf_id, s.participant_id)
+        self.assertEqual(p.biospecimens[0].external_sample_id,
+                         s.external_sample_id)
+
+    def test_find_sampe(self):
```
znatty22: Rename to `test_find_biospecimen`

dataservice/api/biospecimen/models.py
parimalak: 👍

dataservice/api/biospecimen/models.py
```diff
@@ -0,0 +1,78 @@
+from dataservice.extensions import db
+from dataservice.api.common.model import Base, KfId
+from dataservice.api.genomic_file.models import GenomicFile
+
+
+class Biospecimen(db.Model, Base):
+    """
+    Biospecimen entity.
+    :param _id: Unique id assigned by RDBMS
```
dankolbman: should be `uuid` now

dataservice/api/biospecimen/models.py
```diff
@@ -0,0 +1,78 @@
...
+    external_sample_id = db.Column(db.Text(),
+                                   doc='Name given to sample by contributor')
+    external_aliquot_id = db.Column(db.Text(),
+                                    doc='Name given to aliquot by contributor')
+    tissue_type = db.Column(db.Text(),
+                            doc='Description of the kind of biospecimen '
+                            'collected')
+    composition = db.Column(db.Text(),
+                            doc='The cellular composition of the biospecimen')
+    anatomical_site = db.Column(db.Text(),
```
dankolbman, line 46: Shouldn't this be covered by `uberon_id`?

dataservice/api/sequencing_experiment/models.py
parimalak: 👍

dataservice/api/biospecimen/models.py
parimalak: it's for storing raw anatomical_site

dataservice/api/biospecimen/models.py
allisonheath: I'm actually wondering if we should add unit fields as a separate column here? For volume it can be different based on whether it's liquid (blood - typically microliters) or solid (tissue - could be real volume like cubic mm or a weigh like mg, sometimes that's called quantity instead of volume). 

Concentration, as long as the SI prefix is the same for both units it should be fine, e.g. GDC specifies milligram/milliliter which would be the same numerical values as micrograms/microliter. The problem is when you start seeing them mixed and ensuring people convert them correctly, I see ng/ul on a few of the manifests for example.

dataservice/api/biospecimen/models.py
allisonheath: This is the raw text field from which uberon_id is derived, right?

dankolbman: yeah, could we name it better?

dataservice/api/biospecimen/models.py
dankolbman: Where else do we have units? Maybe we should fk units to another unit table to ensure we have a strict vocab?

parimalak: @dankolbman we have units in biospecimen only

znatty22: @allisonheath  Oh I'm sorry, I meant that I have seen ng/ul for most of the x01 cohorts I've looked at not ug/ul. Hmm I don't know if we need a separate table for units just yet. I think the only fields that require units right now are these two in biospecimen, concentration and volume. Will we eventually start tracking weight and height in phenotypes?

dataservice/api/biospecimen/models.py
znatty22: What would we name it? `raw_anatomical_site`?

parimalak: ya. we can rename other ontology's raw texts too in this standard. for suppose adding `raw_` to original field name.

allisonheath: How about `source_`? and yeah would be nice to have this consistent across the tables.

allisonheath: How about `source_`? and yeah would be nice to have this consistent across the tables.

dankolbman: We already use `raw_text` for diagnosis+phenotype. Maybe `raw_` is better

allisonheath: We do? I thought it was still just `diagnosis` and `phenotype`? Not really too opinionated here, mainly that it's consistent.

dankolbman: yeah, I guess you weren't here for that discussion last week. Those names weren't very helpful, so we renamed `raw_text`.

dataservice/api/biospecimen/models.py
allisonheath: Weight and height are properties we're evaluating to ask for moving forward. We also use days as units for age. If we go with a specific unit for now then I'd say yes, let's explicitly name them like the comment below.

dataservice/api/biospecimen/models.py
allisonheath: Literally `raw_text` or `raw_phenotype` / `raw_diagnosis`? Also don't see issue/PR for it?

parimalak: we discussed but didn't change anything yet. it's still phenotype and diagnosis

dataservice/api/biospecimen/models.py
parimalak: Ok. Let's rename `concentration` to `concentarion_mg_per_ml` and volume to `volume_ml` and  make sure to convert values to these units while importing data.

dataservice/api/biospecimen/models.py
parimalak: Created a separate issue for changing the ontology raw fields #237

znatty22 (approved):
👍

dankolbman (approved):
💯

PR #230, opened 2018-04-10
✨ Filter Resources by Study 

znatty22:
Not a full implementation of resource query parameters/filters. Should look at using marshmallow schemas with [webargs](https://webargs.readthedocs.io/en/latest/advanced.html#advanced) for more complete implementation of query filters. Simple implementation for one query parameter (study_id):
- /\<resource name\>?study_id=\<study kids first id\>. 

- [x] Filter participants by study_id 
- [x] Filter study files by study_id 
- [x] Filter diagnoses by study_id 
- [x] Filter phenotypes by study_id 
- [x] Filter outcomes by study_id 
- [x] Filter family_relationships by study_id 
- [x] Filter familes by study_id 
- [x] Filter investigators by study_id 
- [x] Filter sequencing_experiment by study_id 

Implement after #145 and #228 merge
- [x] Filter genomic files by study_id
- [x] Filter biospecimens by study_id

### DIFF
--- a/dataservice/api/biospecimen/resources.py
+++ b/dataservice/api/biospecimen/resources.py
@@ -1,6 +1,6 @@
 from flask import abort, request
 from marshmallow import ValidationError
-from sqlalchemy.orm import Load, load_only
+from sqlalchemy.orm import joinedload
 
 from dataservice.extensions import db
 from dataservice.api.common.pagination import paginated, Pagination
@@ -30,14 +30,15 @@ class BiospecimenListAPI(CRUDView):
             resource:
               Biospecimen
         """
-        q = Biospecimen.query.options(load_only('kf_id'))
+        q = Biospecimen.query
 
         # Filter by study
         from dataservice.api.participant.models import Participant
         study_id = request.args.get('study_id')
         if study_id:
-            q = (q.join(Participant.biospecimens)
-                 .options(Load(Participant).load_only('kf_id', 'study_id'))
+            q = (q.options(joinedload(Biospecimen.genomic_files)
+                           .load_only('kf_id'))
+                 .join(Participant.biospecimens)
                  .filter(Participant.study_id == study_id))
 
         return (BiospecimenSchema(many=True)
--- a/dataservice/api/biospecimen/resources.py
+++ b/dataservice/api/biospecimen/resources.py
@@ -1,5 +1,6 @@
 from flask import abort, request
 from marshmallow import ValidationError
+from sqlalchemy.orm import Load
 
 from dataservice.extensions import db
 from dataservice.api.common.pagination import paginated, Pagination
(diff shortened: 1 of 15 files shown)


dankolbman:
It way be worth validating the `study_id` before attempting a query to make sure it's a kf_id. Maybe return the default results plus a warning in the message?

znatty22:
Yea I was wondering about that. Right you can put in any string for the `study_id` query parameter and 0 results will be returned which I think makes sense. You're searching for entities relating to a study that doesn't exist. Perhaps we could validate the `study_id` and throw a validation error and report to the user as we normally do. I don't think we want to return the default pagination results?

dataservice/api/biospecimen/resources.py
```diff
@@ -29,7 +30,15 @@ def get(self, after, limit):
             resource:
               Biospecimen
         """
-        q = Biospecimen.query
+        q = Biospecimen.query.options(load_only('kf_id'))
```
dankolbman: Does this not only return kf_id on the biospecimen?

dataservice/api/family/resources.py
```diff
@@ -28,7 +29,17 @@ def get(self, after, limit):
...
+        q = Family.query.options(load_only('kf_id'))
+
+        # Filter by study
+        from dataservice.api.participant.models import Participant
+        study_id = request.args.get('study_id')
+        if study_id:
+            q = (q.join(Family.participants)
+                 .options(Load(Participant).load_only('kf_id', 'study_id'))
+                 .filter(Participant.study_id == study_id)
+                 .distinct(Family.kf_id)
```
dankolbman: Could also do a left join? I believe I've seen that `count(distinct)` isn't often performant...

dataservice/api/family/resources.py
```diff
@@ -28,7 +29,17 @@ def get(self, after, limit):
...
+                 .order_by(Family.kf_id))
```
dankolbman: Doesn't this get re-ordered by the pagination?

tests/test_pagination.py
```diff
@@ -1,49 +1,68 @@
...
+
 
 class TestPagination:
     """
     Test that entities are iterated and returned properly
     """
 
     @pytest.fixture(scope='function')
-    def participants(client):
+    def participants(client, indexd):
```
dankolbman, line 32: May want to consider merging this with the fixture in conftest

dataservice/api/family/resources.py
znatty22: I have to include the order_by because of the call to distinct. If I don't, PG throws an exception. Its because the DISTINCT ON expression needs to know which row to pick out of the ones that matched on Family.kf_id.

dataservice/api/biospecimen/resources.py
znatty22: Eh nope thats not what I want. And it looks like all my queries are actually missing joinedloads for the related child entities so I need to fix that as well.

dankolbman (approved):
💯

parimalak (approved):
👍

PR #235, opened 2018-04-11
🐛 fix backref's for diagnoses, outcomes, phenotypes in participant model

parimalak:


znatty22 (approved):
💯

dankolbman (commented):
Should we set `uselist=False`?? on the backref?

parimalak:
isn't `uselist=False` lets only single value which is not desired for us as participant expects list of `phenotypes` or `outcomes` ?

dankolbman:
Not if it's inside the backref: `backref=db.backref('participant', uselist=False, lazy=True)`

znatty22:
@parimalak we do want the backref to be singular because phenotype reference one participant, outcome references one participant, etc. Going the other way, participant has many phenotypes, many outcomes, etc. 

@dankolbman I don't think you need to specify `uselist=False` here because sqlalchemy knows this is a one to many relationship (since there is a foreign key on the child) and the backref is referring to the many to one relationship. `uselist` is set to False by default.

dankolbman (approved):

PR #244, opened 2018-04-16
✨ Sequencing Center entity

parimalak:
- [x] Add sequencing_center model and resources
- [x] Add and update tests
- [x] Update erd and migrations
- [x] Update data generation script


dataservice/api/biospecimen/models.py
```diff
@@ -69,6 +68,10 @@ class Biospecimen(db.Model, Base):
                                db.ForeignKey('participant.kf_id'),
                                nullable=False,
                                doc='The kf_id of the biospecimen\'s donor')
+    sequencing_center_id = db.Column(KfId(),
+                                     db.ForeignKey('sequencing_center.kf_id'),
+                                     nullable=False,
```
znatty22, line 71: Hmmm, should we really make this a required field? Before it wasn't. @allisonheath can you verify

dataservice/api/sequencing_center/models.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    SequencingExperiment entity.
+    :param kf_id: Unique id given by the Kid's First DCC
+    :param name: Name given to sequencing center by contributor
+    """
+    __tablename__ = 'sequencing_center'
+    __prefix__ = 'SC'
+    name = db.Column(db.Text(), nullable=False,
+                     doc='Name given to sequencing center by contributor')
+    sequencing_experiment_id = db.Column(
```
znatty22: Hmm I don't think this makes sense here. Since sequencing center has a 1 to many relationship with sequencing experiment, sequencing experiment should have a foreign key sequencing_center_id. Otherwise, the relationship is backwards

dataservice/api/biospecimen/schemas.py
```diff
@@ -19,6 +19,8 @@ class BiospecimenSchema(BaseSchema):
                           validate=validate_positive_number)
 
     shipment_date = DateOrDatetime()
+    sequencing_center_id = field_for(Biospecimen, 'sequencing_center_id',
```
znatty22, line 22: Same comment here, do we really want to make this a required field? @allisonheath can you verify

dataservice/api/sequencing_center/models.py
```diff
@@ -0,0 +1,28 @@
...
+                                         KfId(),
+                                         db.ForeignKey(
+                                                       'sequencing'
+                                                       '_experiment.kf_id'),
+                                         nullable=False,
+                                         doc='The kf_id of the sequencing'
+                                         ' experiment')
```
znatty22: I think we're missing a relationship and backref here. Similar to comment above, since sequencing center has a 1 to many relationship with sequencing experiment you need to include: 
```python
sequencing_experiments = db.relationship(SequencingExperiment,
                                   backref='sequencing_center')
)
```
Also, I don't think you want the cascade delete here? If a sequencing center is deleted, we probably don't want to delete all the biospecimens and sequencing experiments that were linked to it. @allisonheath can you verify

dataservice/api/sequencing_experiment/models.py
```diff
@@ -68,3 +68,8 @@ class SequencingExperiment(db.Model, Base):
                                     backref=db.backref(
                                         'sequencing_experiment',
                                         lazy=True))
+    sequencing_centers = db.relationship(SequencingCenter,
```
znatty22: Remove this as per above comments

tests/sequencing_center/test_seq_center_models.py
```diff
@@ -0,0 +1,127 @@
...
+        sc = SequencingCenter.query.one_or_none()
+
+        sc.name = "Baylor"
+        db.session.commit()
+
+        # get updated sequencing_center
+        e = SequencingCenter.query.one_or_none()
+        self.assertEqual(e.name, "Baylor")
+
+    def test_delete_sequencing_center(self):
```
znatty22, line 75: This test should check whether relations were properly managed (if cascade delete enabled for orphans on a related entity, then those related entities should have been deleted. if cascade delete was not enabled for orphans - which I don't think it should be - then none of the related entities should be deleted when a seq center is deleted.)

dataservice/api/biospecimen/models.py
parimalak: Ya i also thought about it like in the cases of where the collected biospecimens are not sent for sequencing etc. in that case we need to remove nullable=False here. Only thing that concerns me is; we usually have sequencing_center information when creating biospecimen and sequencing experiment . Making foreign key nullable in 1:n relationship makes difficult to join if multiple null values exists in the sequencing_center table. For suppose, Biospecimen1 from one center has null in sequencing_center_id and and also Biospecimen2 from another center has null value. Now we get in to situation of dealing with foreign key issues.

znatty22: I'm not sure I understand... there should never be null values in the sequencing center table. The null values would be foreign keys to the sequencing center table from biospecimen or sequencing experiment. Also i don't think there would be issues with joins between sequencing center and biospecimen/sequencing experiment where the foreign keys are null. We would just need to make sure the join is not an inner join so that we still get back biospecimens/sequencing experiments from the join result.

dankolbman: `nullable=False` seems to make sense to me. AFAIK, we receive initial information about the sample/biospecimen from the shipment manifests, which are manifests provided to a specific sequencing center.

dataservice/api/biospecimen/models.py
```diff
@@ -52,8 +51,8 @@ class Biospecimen(db.Model, Base):
                                  doc='Disease present in the biospecimen')
     shipment_origin = db.Column(db.Text(),
                                 doc='The original site of the aliquot')
-    shipment_destination = db.Column(db.Text(),
-                                     doc='The site recieving the aliquot')
+    # shipment_destination = db.Column(db.Text(),
+    #                                  doc='The site recieving the aliquot')
```
dankolbman: Remove this?

dataservice/api/sequencing_center/resources.py
```diff
@@ -0,0 +1,157 @@
...
+          properties:
+            resource:
+              SequencingCenter
+        """
+        q = SequencingCenter.query
+
+        return (SequencingCenterSchema(many=True)
+                .jsonify(Pagination(q, after, limit)))
+
+    def post(self):
```
dankolbman, line 52: We may consider an upsert here by checking to see if there is already a sequencing center with the same name, and returning that instead of creating a duplicate center.

dataservice/api/biospecimen/models.py
znatty22: I think @parimalak has a good point though, sometimes biospecimens will exist and we won't have sequencing center info for them.

dataservice/api/sequencing_center/resources.py
znatty22: Should we put a unique constraint on the name?

dataservice/api/biospecimen/models.py
allisonheath: Would it make sense to make `nullable=False` now since all of the current use cases there is a sequencing center (as the info is coming from a shipping manifest) and relax it later if the case of biospecimens without an assigned center does come up?

dataservice/api/sequencing_center/resources.py
dankolbman: That would make sense if we go that route.

dataservice/api/sequencing_center/models.py
```diff
@@ -0,0 +1,24 @@
+from dataservice.extensions import db
+from dataservice.api.common.model import Base, KfId
```
znatty22: Remove unused import KfId

tests/sequencing_center/test_seq_center_models.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        ids = self.create_seqexp_seqcen()
+
+        # get sequencing_center
+        e = SequencingCenter.query.one_or_none()
+
+        # Delete sequencing_center
+        self.assertRaises(IntegrityError, db.session.delete(e))
+
+    def test_not_null_constraint(self):
```
znatty22: This doesn't seem to be testing anything..did you mean to test creation of seq center without a name? Also, I've learned that you have to do this when testing for exceptions:

```python
db.session.add(sc)
    db.session.commit()
```
Just doing `self.assertRaises()` always results in a pass

tests/sequencing_center/test_seq_center_models.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        Test sequencing_center can be created with out sequencing experiment
+        """
+        # Create sequencing_center without sequencing experiment kf_id
+        sc = SequencingCenter(name='Baylor')
+
+        # Add sequencing_center to db
+        self.assertRaises(IntegrityError, db.session.add(sc))
+
+    def test_unique_constraint(self):
```
znatty22: Again, I don't think this is actually testing anything. Looks like you need to do something like:

```python
ids = self.create_seqexp_seqcen()
sc = SequencingCenter(name='Baylor')
db.session.add(sc)
    db.session.commit()
```

tests/sequencing_center/test_seq_center_models.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        Test deleting sequencing_center with existing sequencing_experiment
+        """
+        ids = self.create_seqexp_seqcen()
+
+        # get sequencing_center
+        e = SequencingCenter.query.one_or_none()
+
+        # Delete sequencing_center
+        self.assertRaises(IntegrityError, db.session.delete(e))
```
znatty22: Similar comment here with testing for exceptions.

dataservice/api/sequencing_center/resources.py
```diff
@@ -0,0 +1,166 @@
...
+        ---
+        description: Get all sequencing_centers
+        template:
+          path:
+            get_list.yml
+          properties:
+            resource:
+              SequencingCenter
+        """
+        q = SequencingCenter.query
```
znatty22: Do the following so that you're loading the children kf_id's upfront and everything is in one query:
```python
SequencingCenter.query
             .options(joinedload(SequencingCenter.biospecimens).load_only('kf_id'))
             .options(joinedload(SequencingCenter.sequencing_experiments).load_only('kf_id'))
```

dataservice/api/__init__.py
```diff
@@ -20,12 +20,14 @@
 from dataservice.api.biospecimen import BiospecimenListAPI
 from dataservice.api.outcome import OutcomeAPI
 from dataservice.api.outcome import OutcomeListAPI
+from dataservice.api.phenotype.resources import PhenotypeAPI
```
znatty22: I think the phenotype ones should be:

```python
from dataservice.api.phenotype import PhenotypeAPI
from dataservice.api.phenotype import PhenotypeListAPI
```

migrations/versions/4a7f7cb8ac00_.py
```diff
@@ -0,0 +1,51 @@
...
+
+Revision ID: 4a7f7cb8ac00
+Revises: 243980a97e4c
+Create Date: 2018-04-19 10:42:46.065593
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
```
znatty22: Need an `import dataservice`

tests/test_api.py
```diff
@@ -259,6 +269,7 @@ def test_unknown_field(self, client, entities, endpoint, method):
         ('/participants', 'phenotypes'),
         ('/studies', 'study_files'),
         ('/families', 'participants'),
+        ('/sequencing-centers', 'biospecimens')
```
znatty22: What about `('/sequencing-centers', 'sequencing_experiments')`

tests/test_api.py
```diff
@@ -259,6 +269,7 @@ def test_unknown_field(self, client, entities, endpoint, method):
...
     ])
     def test_relations(self, client, entities, resource, field):
```
znatty22: Looks like you're also missing a couple tests for sequencing centers in `test_bad_input` and `test_missing_required_params`

znatty22 (commented):
Maybe I missed it, but I didn't see a test on the unique constraint for the sequencing_center.name. Probably want that test in the resource tests, so you can also test for the integrity error, error message content, etc

tests/sequencing_center/test_seq_center_models.py
```diff
@@ -0,0 +1,109 @@
...
+import random
+
+from dataservice.extensions import db
+from dataservice.api.sequencing_experiment.models import SequencingExperiment
+from dataservice.api.sequencing_center.models import SequencingCenter
+from tests.utils import FlaskTestCase
+
+from sqlalchemy.exc import IntegrityError
+
+MAX_SIZE_MB = 5000
```
dankolbman: Are these used anywhere?

znatty22 (approved):
👍

dankolbman (approved):
lgtm 🎉

PR #257, opened 2018-04-21
🐛 Install python deps in venv

dankolbman:
Was missing the setup of venv in the docker install, causing the service to fail if built in a directory without the venv already installed.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+Dockerfile
+venv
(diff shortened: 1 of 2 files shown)


znatty22 (approved):

parimalak (approved):

baileyckelly:
Moving to Backlog

baileyckelly:
Moving to backlog

dankolbman:
I don't deem this necessary any more as there is not really a reason to add another layer of encapsulation within virtualenv while running in docker.

Issue #248, opened 2018-04-18
Delete orphaned parent entities when children have been deleted 


znatty22:
Applies to family, sequencing experiments, and alias groups:

Currently when we delete all participants in a family, the family hangs around. We need to change this behavior in the model so that when a family is orphaned (no related participants exist), delete the family. 

Similarly with sequencing experiments, when we delete all genomic files in a seq exp, the seq exp hangs around. We need to change this behavior in the model so that when a seq exp is orphaned (no related genomic files exist), delete the seq exp.

Issue #122, opened 2018-02-11
Use postgres CASCADE for deletion


dankolbman:
We currently rely on the sqlalchemy ORM cascading for deletion of child relations which requires all children, (and their children) to be pulled into memory from the database and deleted. By using `ondelete='CASCADE'`, postgres will take care of the deletions without any further communication with the ORM.
A comparison I did using `cascades='all, delete-orphan'` and `ondelete='CASCADE'`, `passive_deletes=True` for a participant with `30 samples` and `900 aliquots`:

- ORM cascade: 44s
- pg cascade: 2.4s

PR #258, opened 2018-04-21
🐛Delete parent orphans

znatty22:
Address #248

### DIFF
--- a/dataservice/api/family/models.py
+++ b/dataservice/api/family/models.py
@@ -1,3 +1,5 @@
+from sqlalchemy import event
+
 from dataservice.extensions import db
 from dataservice.api.common.model import Base
 from dataservice.api.participant.models import Participant
@@ -18,3 +20,10 @@ class Family(db.Model, Base):
     external_id = db.Column(db.Text(), doc='ID used by external study')
 
     participants = db.relationship(Participant, backref='family')
+
+
+@event.listens_for(Participant, 'after_delete')
+def delete_orphans(mapper, connection, state):
+    q = (db.session.query(Family)
+         .filter(~Family.participants.any()))
+    q.delete(synchronize_session='fetch')
--- a/dataservice/api/participant/models.py
+++ b/dataservice/api/participant/models.py
@@ -1,6 +1,6 @@
 from itertools import chain
 
-from sqlalchemy import and_
+from sqlalchemy import and_, event
 
 from dataservice.extensions import db
 from dataservice.api.common.model import Base, KfId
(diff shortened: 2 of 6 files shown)


dataservice/api/family/models.py
```diff
@@ -18,3 +20,10 @@ class Family(db.Model, Base):
     external_id = db.Column(db.Text(), doc='ID used by external study')
 
     participants = db.relationship(Participant, backref='family')
+
```
dankolbman, line 23: If we change the above `participants` relation to have `backref=backref("family", cascade="all,delete")`, will that accomplish the same thing?

dankolbman:
Are we sure there isn't a way to do this without events?
Now might be a good time to look more closely at #122  too

dataservice/api/family/models.py
znatty22: `cascade = "all, delete"` is actually already doing the delete since the `all` translates to `save-update, merge, refresh-expire, expunge, delete`.

Maybe what you meant was `cascade = "all, delete-orphan"` which unfortunately does not work. It will delete the family when any one participant is deleted and then NULL out the other participant's foreign keys which refer to the deleted family.

znatty22:
I don't think there is a way to do this without events. Here's what sqlalchemy says about deleting parents when children are deleted: 

http://docs.sqlalchemy.org/en/latest/orm/cascades.html

> While database-level ON DELETE functionality works only on the “many” side of a relationship, SQLAlchemy’s “delete” cascade has limited ability to operate in the reverse direction as well, meaning it can be configured on the “many” side to delete an object on the “one” side when the reference on the “many” side is deleted. However this can easily result in constraint violations if there are other objects referring to this “one” side from the “many”, so it typically is only useful when a relationship is in fact a “one to one”. The single_parent flag should be used to establish an in-Python assertion for this case.

Its seems that a parent can indeed be deleted when a related child is deleted. However, the key in that statement is the "a related child". I think this is what they mean by "SQLAlchemy’s “delete” cascade has limited ability to operate in the reverse direction". What we want is to delete a parent only once all the children are deleted. 

As for configuring cascade deletes using Postgres, we can definitely do that for all of our foreign keys since its much more efficient. But again, even with postgres ON DELETE CASCADE we can't achieve what we want (only delete parent when all children are deleted). The reason why is bc cascading delete, whether you do it through the ORM or the database is always about cascading deletes from parent to children.

dankolbman (approved):
Should be alright since the clean-up queries should be run relatively infrequently and the number of families should remain reasonable.

parimalak (approved):

Issue #260, opened 2018-04-24
API /studies with POST method does not include the attribute "investigator_id" 


grugna:
{'_status': {'code': 400, 'message': "could not create study: {'investigator_id': ['Unknown field']}"}}

PR #261, opened 2018-04-24
✨ Relation links

### LINKED ISSUE #245: Implement/Verify all resource links
Verify that the `_links` element in any dataservice response has the correct related entity links. Every entity that has a parent should have a link to it's parent in `_links`. If a link isn't there implement it.

This is especially important for genomic file since Cavatica task service wants to collect all the clinical data for a particular genomic file. In order to do this, the service will need to follow the links back to the participant and collect the response data as it traverses.

znatty22:
Closes #245, #260


dataservice/api/participant/schemas.py
```diff
@@ -21,19 +20,29 @@ class Meta(BaseSchema.Meta):
     _links = ma.Hyperlinks({
         'self': ma.URLFor(Meta.resource_url, kf_id='<kf_id>'),
         'collection': ma.URLFor(Meta.collection_url),
-        'study': ma.URLFor('api.studies', kf_id='<study_id>'),
+        'study': ma.URLFor('api.studies', kf_id='<study_id>')
     })
 
-    def dump(self, p, *args, **kwargs):
+    def dump(self, obj, *args, **kwargs):
```
znatty22: I think there may be a more generic way to do this (and rendering of all of our relation links whether parent or child) on the base schema, but I think it would require some refactoring and I'd like to get this first version merged in soon since it's a blocker for consumers.

dataservice/api/common/custom_fields.py
```diff
@@ -1,10 +1,73 @@
...
+
+    def _serialize(self, value, key, obj):
+        """Output the URL for the endpoint, given the kwargs passed to
+        ``__init__``.
+        """
+        param_values = {}
+        for name, attr_tpl in iteritems(self.params):
+            attr_name = _tpl(str(attr_tpl))
+            if attr_name:
+                attribute_value = get_value(obj, attr_name, default=missing)
```
dankolbman, line 53: What is `missing`?

znatty22: `missing` is a marshmallow obj and has been imported from marshmallow (see top of file). Its just used as the default value when the attribute doesn't exist on the obj (ie. family_id doesn't exist on Participant).
 
I basically just copied and pasted the original `URLFor._serialize` method and made a couple minor tweaks to make it work for null values using the `allow_none` kwarg

dankolbman (approved):
lgtm 💯

parimalak (approved):

PR #263, opened 2018-04-24
✨ Add integration with bucket service

dankolbman:
Calls bucket service to create a new bucket for every new study added.

### DIFF
--- a/dataservice/api/study/models.py
+++ b/dataservice/api/study/models.py
@@ -1,3 +1,6 @@
+import requests
+from sqlalchemy import event
+from flask import current_app
 from dataservice.extensions import db
 from dataservice.api.common.model import Base, KfId
 from dataservice.api.participant.models import Participant
@@ -51,3 +54,13 @@ class Study(db.Model, Base):
 
     def __repr__(self):
         return '<Study {}>'.format(self.kf_id)
+
+
+@event.listens_for(Study, 'after_insert', propagate=True)
+def register_indexd(mapper, connection, target):
+    """
+    Invokes the new bucket service to create a bucket in s3 for this study
+    """
+    url = current_app.config['BUCKET_SERVICE_URL']
+    if url is not None:
+        resp = requests.post(url+'/buckets', json={'study_id': target.kf_id})
--- a/dataservice/api/study/models.py
+++ b/dataservice/api/study/models.py
@@ -57,10 +57,19 @@ class Study(db.Model, Base):
 
 
 @event.listens_for(Study, 'after_insert', propagate=True)
-def register_indexd(mapper, connection, target):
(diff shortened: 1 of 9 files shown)


dataservice/api/study/models.py
```diff
@@ -51,3 +54,19 @@ class Study(db.Model, Base):
 
     def __repr__(self):
         return '<Study {}>'.format(self.kf_id)
+
+
+@event.listens_for(Study, 'after_insert', propagate=True)
+def register_indexd(mapper, connection, target):
```
znatty22: Is this an unintentional copy/paste .. should the method be called something else?

tests/mocks.py
```diff
@@ -101,3 +101,18 @@ def indexd(mocker):
     mock.Session().get.side_effect = indexd_mock.get
     mock.Session().post.side_effect = indexd_mock.post
     yield mock.Session()
+
```
znatty22: Do you still need this since its now part of the FlaskTestCase class?

tests/study/test_study_models.py
```diff
@@ -29,6 +30,20 @@ def test_create_and_find(self):
         for p in participants:
             self.assertEqual(study.kf_id, p.study.kf_id)
 
+    def test_bucket_service(self):
```
znatty22, line 33: What happens if someone tries to create buckets with the same study_id? Will the old bucket be overwritten? Will bucket service send back a 400 or does the bucket service do an upsert?

tests/study/test_study_models.py
```diff
@@ -4,6 +4,7 @@
 from dataservice.api.participant.models import Participant
 from dataservice.api.study.models import Study
 from tests.utils import FlaskTestCase
+from unittest.mock import patch, Mock
 
 
 class ModelTest(FlaskTestCase):
```
znatty22, line 10: How does deleting a study affect the study's bucket?

tests/mocks.py
dankolbman: Yeah, it's used in the participants mock.

tests/study/test_study_models.py
dankolbman: Right now it will 500 on the bucket service. The dataservice, however, does  not perform any checks to see that the request was completed successfully (can't really do anything, unless we want to completely fail the request)

tests/study/test_study_models.py
dankolbman: Nothing happens. The bucket will remain so as to avoid any unintentional data loss. May be worth adding logic to the bucket service to remove buckets only if they are empty and don't exist in the dataservice.

znatty22 (approved):
👍

parimalak (approved):

Issue #48, opened 2018-01-19
List Desired Enums


baileyckelly:
Participant Fields
- [ ] Race
- [ ] Gender
- [ ] Ethnicity

Diagnosis Fields
- [ ] diagnosis_category

parimalak:
release_status of study

allisonheath:
Idea is to put this in the Marshmallow enums on the application layer so it's easier to adjust than on the database layer.

parimalak:
can we make these columns to enums?
- [ ] vital_status in outcome (Dead, alive or Not Reported)
- [ ] availability in genomic_file and study_file(available for immediate download, cold storage, unknown)

PR #270, opened 2018-04-26
✨ Add availabilty to study_file

parimalak:


znatty22 (approved):
💯

dankolbman:
Do we want to define the values of `availability`?
Maybe just `available` or `cold storage`?

znatty22:
I think like everything else, eventually this field will become an ENUM. We could define those values informally now and put it in the docstring tho

dataservice/api/study_file/models.py
```diff
@@ -16,7 +16,8 @@ class StudyFile(db.Model, Base, IndexdFile):
     :param urls: Locations of file
     :param hashes: A dict keyed by hash type containing hashes of the file
     :param _metadata: A dict with any additional information
-    :param controlled_access: whether or not the file is controlled access
+    :param availability: Indicates whether a file is available for immediate
```
parimalak, line 19: its in doc string

dankolbman: yeah, I see that. Just wondering if we need to define these enums.

znatty22: Will be figured out in #48 and implemented in #274

dankolbman (approved):

PR #281, opened 2018-04-30
🗃 Squash migrations

### LINKED ISSUE #273: Squash database migrations
We should squash our migrations one last time before we generate our next release.

dankolbman:
Squash migrations to single version for v1.0.0

Also adds the datamodel version and migration number to the `/status` endpoint.
Any future migrations will need to begin with `X.X.X -` semantic version number before the description.

closes #273


znatty22 (commented):
How are we assigning the semantic versions of the data model?

dankolbman:
Using [semver](https://semver.org/).
Basically:
- minor version if addition of columns/tables/relations
- major version if removal or renaming of columns/tables/relations

znatty22:
Ok I see. And jw, will these versions be manually updated by us as we generate new migrations or are we going to automate this in some way?

dankolbman:
It would be pretty difficult to do automatic versioning, but not impossible. We could at least have another check to make sure that version doesn't exist and it's > than any other version number. The migration process is already pretty manual, so hopefully we can keep this in check.

znatty22 (approved):
👍

parimalak (approved):
👍

PR #282, opened 2018-04-30
🚧 Draft enumerations for controlled fields

znatty22:
Wasn't sure where exactly to put this, but I wanted to get the list of entities and controlled fields out for review.

A lot of the values were taken from: https://github.com/kids-first/kf-dictionary which is a fork of the GDC dictionary. So some things may not make sense and I'd like us to weed those out!

Please take a look and feel free to comment or modify and submit PRs against this one.

### DIFF
--- a/dataservice/api/common/enumerations.yaml
+++ b/dataservice/api/common/enumerations.yaml
@@ -1,33 +1,40 @@
-## ------------ NOTES ----------- ##
-# 'Unknown' - contributor specified this in source data
-# 'Not Reported' - should be used as the default if no value is specified in
-# the column or if this column doesn't exist in source data
 
-## --------- Questions --------- ##
-# How should values be cased - lower case, proper/title case?
-# What should default value be if Not Reported in source data?
+common:
+    # The value was not reported, either to the original data source or to the
+    # DRC, could potentially mean that follow up could occur to gather the
+    # value if it's meaningful
+    - Not Reported
+
+    # In the current context the value isn't applicable
+    # (e.g. tumor location for a structural birth defect
+    - Not Applicable
+
+    # The value cannot be collected due to privacy, consent, etc other issues,
+    # follow up cannot occur
+    - Not Allowed To Collect
+
+    # Follow up was tried but the value isn't available
+    - Not Available
+
+    # Contributor explicitly reported that the value is unknown
+    - Reported Unknown
 
 study:
     release_status:
         choices:
             - Pending # Not sure about the choices here
-        default: Pending
 
 study_file:
     availability:
         choices:
             - Immediate Download
             - Cold Storage
-        default: Immediate Download
 
 participant:
     gender:
         choices:
             - Male
             - Female
-            - Unknown
-            - Not Reported
-        default: Not Reported
     race:
         choices:
           - White
@@ -36,16 +43,10 @@ participant:
           - Asian
           - Native Hawaiian or Other Pacific Islander
           - Other
-          - Unknown
-          - Not Reported
-        default: Not Reported
     ethnicity:
         choices:
             - Hispanic or Latino
             - Not Hispanic or Latino
-            - Unknown
-            - Not Reported
-        default: Not Reported
 
 diagnosis:
     diagnosis_category:
@@ -53,32 +54,22 @@ diagnosis:
             - Structural Birth Defect
             - Cancer
             - Other
-            - Unknown
-        default: Other
 
 phenotype:
     observed:
         choices:
             - Positive
             - Negative
-            - Unknown
-        default: Unknown
 
 outcome:
     vital_status:
         choices:
             - Alive
             - Dead
-            - Unknown
-            - Not Reported
-        default: Not Reported
     disease_related:
         choices:
             - Yes
             - No
-            - Unknown
-            - Not Reported
-        default: Not Reported
 
 biospecimen:
     analyte_type:
@@ -86,8 +77,6 @@ biospecimen:
             - DNA
             - RNA
             - Other
-            - Not Reported
-        default: Not Reported
 
 genomic_file:
     data_type:
@@ -96,12 +85,10 @@ genomic_file:
             - Submitted Aligned Reads Index
             - Simple Nucleotide Variation
             - Other
-        default: Other
     availability:
         choices:
             - Immediate Download
             - Cold Storage
-        default: Immediate Download
 
 sequencing_experiment:
     experiment_strategy:
@@ -116,8 +103,6 @@ sequencing_experiment:
             - Amplicon
             - Targeted Sequencing
             - Other
-            - Not Reported
-        default: Not Reported
     platform:
         choices:
             - Illumina
@@ -127,13 +112,8 @@ sequencing_experiment:
             - Complete Genomics
             - PacBio
             - Other
-            - Not Reported
-        default: Not Reported
     library_strand:
         choices:
             - Unstranded
             - First Stranded
             - Second Stranded
-            - Unknown
-            - Not Reported
-        default: Not Reported
--- a/dataservice/api/common/enumerations.yaml
+++ b/dataservice/api/common/enumerations.yaml
@@ -97,10 +97,6 @@ genomic_file:
             - Simple Nucleotide Variation
             - Other
         default: Other
-    file_format:
-        choices:
-            - ???
-        default: ???
     availability:
         choices:
             - Immediate Download
--- a/dataservice/api/common/enumerations.yaml
+++ b/dataservice/api/common/enumerations.yaml
@@ -132,8 +132,8 @@ sequencing_experiment:
     library_strand:
         choices:
             - Unstranded
-            - First_Stranded
-            - Second_Stranded
+            - First Stranded
+            - Second Stranded
             - Unknown
             - Not Reported
         default: Not Reported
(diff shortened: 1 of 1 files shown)


dataservice/api/common/enumerations.yaml
```diff
@@ -0,0 +1,165 @@
...
+            - Ion Torrent PGM
+            - Ion Torrent Proton
+            - PacBio RS
+            - Other
+            - Not Reported
+        default: Not Reported
+    library_strand:
+        choices:
+            - Unstranded
+            - First_Stranded
```
dankolbman: Why the underscores here?

dataservice/api/common/enumerations.yaml
```diff
@@ -0,0 +1,165 @@
...
+            - Illumina
+            - SOLiD
+            - LS454
+            - Ion Torrent
+            - Complete Genomics
+            - PacBio
+            - Other
+            - Not Reported
+        default: Not Reported
+    instrument_model:
```
dankolbman: Do we need an exhaustive list here? There's quite a few models out there

dataservice/api/common/enumerations.yaml
```diff
@@ -0,0 +1,165 @@
...
+            - ???
+        default: ???
+    availability:
+        choices:
+            - Immediate Download
+            - Cold Storage
+        default: Immediate Download
+
+sequencing_experiment:
+    experiment_strategy:
```
dankolbman: Do we have any non-wgs?

dataservice/api/common/enumerations.yaml
```diff
@@ -0,0 +1,165 @@
...
+    data_type:
+        choices:
+            - Submitted Aligned Reads
+            - Submitted Aligned Reads Index
+            - Simple Nucleotide Variation
+            - Other
+        default: Other
+    file_format:
+        choices:
+            - ???
```
dankolbman: Could probably use the file extensions, with any `.gz` trimmed off.

dataservice/api/common/enumerations.yaml
```diff
@@ -0,0 +1,165 @@
...
+        choices:
+            - Yes
+            - No
+            - Unknown
+            - Not Reported
+        default: Not Reported
+
+biospecimen:
+    analyte_type:
+        choices:
```
dankolbman: `Other` may be a good option here

dataservice/api/common/enumerations.yaml
```diff
@@ -0,0 +1,165 @@
...
+# 'Unknown' - contributor specified this in source data
+# 'Not Reported' - should be used as the default if no value is specified in
+# the column or if this column doesn't exist in source data
+
+## --------- Questions --------- ##
+# How should values be cased - lower case, proper/title case?
+# What should default value be if Not Reported in source data?
+
+study:
+    release_status:
```
dankolbman: Should this follow the same finite state as the release process?

dataservice/api/common/enumerations.yaml
znatty22: Probably don't need them. I copied and pasted from the kf-dictionary repo. See PR description

dataservice/api/common/enumerations.yaml
znatty22: Yea, thats what I was thinking but I wasn't sure

dataservice/api/common/enumerations.yaml
```diff
@@ -0,0 +1,165 @@
...
+            - Unknown
+            - Not Reported
+        default: Not Reported
+
+biospecimen:
+    analyte_type:
+        choices:
+            - DNA
+            - RNA
+            - Unknown
```
parimalak: `analyte_type` can also take these values https://github.com/kids-first/kf-api-dataservice/blob/master/dataservice/util/data_gen/analyte_type.txt (These values are taken from kf-dictionary) ; and some of CBTTC data also have `Blood`, `Tissue`, `Lys`.

dataservice/api/common/enumerations.yaml
fiendish: Agreed. Something mumble
```python
def file_type_from_name(file_name):
    """Get the primary file extension from a file name or path.
    "file_name: file path or name string e.g. "foo/bar/baz.qux.xyzzy"
    returns:
        The first file extension, e.g. qux.
    """
    format = None
    while file_name is not None:
        file_name, ext = os.path.splitext(file_name)
        if not ext:
            break
        format = ext
    return format
```
-ish

dataservice/api/common/enumerations.yaml
fiendish: I need clarification on what release status is for. If something isn't released, then it will be released on a particular date/schedule according to contract/tos/whatever. No?

dataservice/api/common/enumerations.yaml
```diff
@@ -0,0 +1,165 @@
...
+            - Cold Storage
+        default: Immediate Download
+
+participant:
+    gender:
+        choices:
+            - Male
+            - Female
+            - Unknown
+            - Not Reported
```
fiendish: Can you clarify a case where Unknown and Not Reported aren't the same thing? Maybe Other.

dataservice/api/common/enumerations.yaml
```diff
@@ -0,0 +1,165 @@
...
+        default: Not Reported
+
+diagnosis:
+    diagnosis_category:
+        choices:
+            - Structural Birth Defect
+            - Cancer
+            - Other
+            - Unknown
```
fiendish: Does Kids-first deal with non-cancer non-sbd?

fiendish (commented):
I think casing should be applied as deemed appropriate without a specific rule.

Maybe missing also...

family relationships?
mother, father, etc

Phenotype shouldn't be enumerated like this (too many, obvs), but it should come from an official list.

dataservice/api/common/enumerations.yaml
znatty22: Often times a data contributor/investigator will actually specify the value "Unknown" in their data or sometimes its "Other" or "N/A". I think `Unknown` could cover these cases. Where as `Not Reported` means that there was nothing in the column OR there was no column from which to extract the field. Its important to distinguish the two, especially for data QC.

dataservice/api/common/enumerations.yaml
znatty22: Yea that makes sense, so no need to include file_format as an enum

znatty22:
I left out family relationships because I thought we don't quite know what all the relationships will be so it seems easier to leave it as free text for now. @allisonheath could probably tell us more definitively

dataservice/api/common/enumerations.yaml
allisonheath: We really need to be consistent with missing/unknown across all fields. It would be best to centralize them otherwise I've seen inconsistency happen before. I like to be a bit more concrete if this covers all of the cases:

`Not Reported` - means that the value was not reported, either to the original data source or to the DRC, could potentially mean that follow up could occur to gather the value if it's meaningful
`Not Allowed to Collect` - means that the value cannot be collected due to privacy, consent, etc other issues, follow up cannot occur
`Not Applicable` -  in the current context the value isn't applicable (e.g. tumor location for a structural birth defect

Other ones to consider:
`Not Available` - means that follow up was tried but the value isn't available
`Other` - a nicer catchall than just `missing` or `unknown`, often good to couple with a note/reason

I can also see `Unknown` being useful when that's the value provided, perhaps an indication like `Reported Unknown`.

dataservice/api/common/enumerations.yaml
allisonheath: Agreed, can we just start with the ones have been / will be used for the next year? We can gather those and then expand this list as needed in the future?

dataservice/api/common/enumerations.yaml
allisonheath: We have RNA-Seq and WXS coming soon, similar to the models, can we start with those and add as needed?

dataservice/api/common/enumerations.yaml
allisonheath: These `analyte_type` are a mix of the actual analyte (molecule of study e.g. DNA, RNA) and the source (e.g. EBV Immortalized Normal, Blood), processing (e.g. Repli-G (Qiagen)) and preservation methods (e.g. FFPE). I'd suggest that we keep the `analyte_type` as `DNA`, `RNA` and `Other`. 

There are several ways to breakdown the fields if we'd like to capture them.

dataservice/api/common/enumerations.yaml
```diff
@@ -0,0 +1,165 @@
...
+            - Not Reported
+        default: Not Reported
+
+genomic_file:
+    data_type:
+        choices:
+            - Submitted Aligned Reads
+            - Submitted Aligned Reads Index
+            - Simple Nucleotide Variation
+            - Other
```
allisonheath: I'm not sure how much value really comes from these choices and they can be hard to define.

znatty22: Currently we assign `data_type` based on the file format (vcf, fq, crai, cram/bam, etc) and potentially whether the file is harmonized or not (i.e. some harmonized files would be assigned Aligned Reads). These were the values I was using when importing data. Again most of them come from GDC. 

Are you proposing that we don't have an enum for `data_type` or a different set of choices? If the latter, how should we determine the data type of the file

PR #283, opened 2018-05-01
🐛 Change modified_at field on update

### LINKED ISSUE #222: modified_at field is not updated
The `modified_at` is supposed to be updated anytime an object is modified in the database, however, this is currently not done. It should be as simple as extending the base class with an event listener.

dankolbman:
`modified_at` actually changes on update now.

fixes #222


znatty22 (approved):
👍

parimalak (approved):
👍

PR #299, opened 2018-05-03
✨ Add enum validators for entities

### LINKED ISSUE #48: List Desired Enums

parimalak:
Participant:
 - [x]  race
- [x] gender
- [x] ethnicity
Diagnosis:
 - [x] diagnosis_category
Phenotype:
 - [x] observed
Outcome:
 - [x] vital_status
- [x] disease_related
biospecimen:
- [x]   analyte_type
genomic_file:
- [x] data_type
- [x] availability
sequencing_experiment
- [x]  experiment_strategy
- [x] platform
- [x] library_strand
study:
- [x]   release_status
study_file:
- [x]  availability
closes #48 

GENDER_ENUM = {'Male', 'Female', 'Other'}
ETHNICITY_ENUM = {'Hispanic or Latino',
			'Not Hispanic or Latino'}
RACE_ENUM = {
'White', 'American Indian or Alaska Native',
 'Black or African American', 'Asian',
  'Native Hawaiian or Other Pacific Islander',
 'Other'}

ANALYTE_TYPE_ENUM = {"DNA", "RNA", "Other"}

AVAILABILITY_ENUM = {'Immediate Download',
			                  'Cold Storage'}

DIAGNOSIS_CATEGORY_ENUM = {'Structural Birth Defect',
				                       'Cancer',
				                          'Other'}

DATA_TYPE_ENUM = {'Aligned Reads',
				      'Aligned Reads Index',
				          'Simple Nucleotide Variation',
			         'Other'}

VITAL_STATUS_ENUM = {'Alive', 'Dead'}
DISEASE_RELATED_ENUM = {'Yes', 'No'}

OBSERVED_ENUM = {'Positive', 'Negative'}

EXPERIMENT_STRATEGY_ENUM = {'WGS', 'WXS', 'RNA-Seq', 'miRNA-Seq', 'Other'}
PLATFORM_ENUM = {'Illumina', 'SOLiD', 'LS454', 'Ion Torrent',
 'Complete Genomics', 'PacBio', 'Other'}
LIBRARY_STRAND_ENUM = {'Unstranded', 'First Stranded',
 'Second Stranded', 'Other'}

RELEASE_STATUS_ENUM = {'Pending', 'Waiting', 'Running', 'Staged', 'Publishing', 'Published', 'Failed', 'Canceled'}


znatty22:
@parimalak I know you have a WIP on this PR but I took a sneak peak :). You can do the following to reduce duplicate code and make your life a bit easier too :)

```python
class CommonEnum(Enum):
    """
    Enum class with values for all fields
    """
    Not_Reported = "Not Reported"
    Not_Applicable = "Not Applicable"
    Not_Allowed_To_Collect = "Not Allowed To Collect"
    Not_Available = "Not Available"
```
And then other enums can extend this one:

```python
class GenderEnum(CommonEnum)
```

parimalak:
@natasha i tried that. Enum class seems to be restricted for class extensions. see https://stackoverflow.com/questions/33679930/how-to-extend-python-enum?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

dataservice/api/common/custom_fields.py
```diff
@@ -95,3 +96,19 @@ def _deserialize(self, value, attr, data):
             raise ValidationError('Invalid date or datetime string')
 
         return value
+
+
+class EnumColumn(fields.Field):
```
dankolbman: Should probably be called an `EnumField` as a `Column` is typically a database term.

dataservice/api/diagnosis/models.py
```diff
@@ -66,3 +67,17 @@ class Diagnosis(db.Model, Base):
                                db.ForeignKey('participant.kf_id'),
                                doc='the participant who was diagnosed',
                                nullable=False)
+
+
+class DiagnosisCategoryEnum(Enum):
+
```
dankolbman: Remove \n

dataservice/api/genomic_file/models.py
```diff
@@ -51,3 +52,31 @@ class GenomicFile(db.Model, Base, IndexdFile):
                                          'sequencing_experiment.kf_id'))
     biospecimen_id = db.Column(KfId(), db.ForeignKey('biospecimen.kf_id'),
                                nullable=True)
+
+
+class DataTypeEnum(Enum):
+
```
dankolbman: Remove \n

dataservice/api/genomic_file/models.py
```diff
@@ -51,3 +52,31 @@ class GenomicFile(db.Model, Base, IndexdFile):
...
+    Simple_Nucleotide_Variation = "Simple Nucleotide Variation"
+    Other = "Other"
+    Not_Reported = "Not Reported"
+    Not_Applicable = "Not Applicable"
+    Not_Allowed_To_Collect = "Not Allowed To Collect"
+    Not_Available = "Not Available"
+
+
+class AvailabilityEnum(Enum):
+
```
dankolbman: Remove \n

dataservice/api/common/custom_fields.py
```diff
@@ -95,3 +96,19 @@ def _deserialize(self, value, attr, data):
...
+    """Validates against a given set of enumerated values."""
+
+    def __init__(self, enum, *args, **kwargs):
```
dankolbman: You always pass a `list` as enum here.
Why not just pass in an enum natively instead?
So this:
```
EnumColumn(enum=[s.value for s in RaceEnum])
```
Becomes:
```
EnumColumn(enum=RaceEnum)
```
Then you take care of the list comprehension inside the `_validate` function

dataservice/api/common/custom_fields.py
```diff
@@ -95,3 +96,19 @@ def _deserialize(self, value, attr, data):
...
+        super().__init__(*args, **kwargs)
+        self.enum = enum
+        self.validators.insert(0,
```
dankolbman: Does the order matter here? Maybe more clean to use `append()` instead, so no one is confused by the extra argument.

dataservice/api/biospecimen/models.py
```diff
@@ -98,3 +99,17 @@ class Biospecimen(db.Model, Base):
                                     backref=db.backref(
                                         'biospecimen', lazy=True),
                                     doc='genomic files this biospecimen')
+
+
+class AnalyteTypeEnum(Enum):
+
+    """
+    Enum class for analyte_type field with possible choices
+   """
```
dankolbman: Indent missmatch

dankolbman (commented):
I see the enums are being defined in the model, but they are only used in the schema. Maybe they should be moved to the schema file instead? Or maybe we should have both a field and a column validator that checks the value against the enum. One on the field, and one on the model column.

dataservice/api/common/custom_fields.py
parimalak: 👍

dataservice/api/common/custom_fields.py
```diff
@@ -95,3 +96,19 @@ def _deserialize(self, value, attr, data):
...
+                               OneOf([v.lower() for v in self.enum]))
+
+    def _validate(self, value):
+        if type(value) is self.enum:
```
znatty22: This method seems to do the same thing regardless of the if condition?

znatty22 (commented):
So I hate to be the bad news bear, but I think this approach is not actually going to work well :/ But the good news is I think I have an alternative solution!

The issue is with custom fields. I've realized we're not quite using them correctly. With the way we're using them, we have to always remember to set the `allow_none` and or `required` booleans manually for our custom fields which represent columns which require either of those two. For example, right now if you try to create a biospecimen without an `analyte_type` the DB will error out because there is no required=True flag set on EnumColumn for this column. The API will give you a 400 but just say `error saving changes`, instead of saying analyte type is required bc the DB errors out.

Luckily marshmallow_sqlalchemy allows you to do the following so that you can still use your custom field and also get the benefit of having those booleans set automatically for you (based on introspection of the column properties):

```python
analyte_type = field_for(Biospecimen, 'analyte_type',
                             field_class=EnumColumn
```
However, this doesn't work for us because enums are not classes :(. I'm submitting a PR against this one with a possible alternate solution for enum validation. Take a look and let me know what you think!

dataservice/api/study_file/schemas.py
```diff
@@ -3,9 +3,15 @@
 from dataservice.api.study_file.models import StudyFile
 from dataservice.api.common.schemas import BaseSchema, IndexdFileSchema
 from dataservice.extensions import ma
+from dataservice.api.genomic_file.schemas import AVAILABILITY_ENUM
```
dankolbman: Maybe this enum is better define in `common.schemas.IndexdFileSchema`?

dankolbman (commented):
Maybe worth adding a new set of tests to verify the enum validator works?

dataservice/api/study_file/schemas.py
parimalak: i think this field belongs to our model

znatty22 (commented):
Looks good with a couple comments/concerns ... You may want to specify in the description of this PR that this is a first shot at enum validation and it is case sensitive. That means that submitters will have to be careful to submit with proper casing. 
I'm not sure if we want to merge this now since we don't have the case insensitive validation worked out. It might make it pretty difficult for us to submit studies if all the enum fields have to have exact casing (depends on how messy data is)

dataservice/api/study_file/schemas.py
dankolbman: But it's shared between all file types, isn't it?

znatty22: Yep, agreed. I think it makes more sense to put this in IndexFileSchema since both genomic files and study files use this enum

parimalak: yes both in study_file and genomic_file. yeah, you are right i can define in common schema

dataservice/api/participant/schemas.py
```diff
@@ -1,17 +1,38 @@
...
+
 from dataservice.api.common.schemas import BaseSchema
 from dataservice.extensions import ma
 
 from dataservice.api.common.custom_fields import PatchedURLFor
+from dataservice.api.common.validation import enum_validation_generator
+
+# Enum Choices for participant fields
+
+GENDER_ENUM = {'Male', 'Female'}
```
fiendish: I think this needs 'Other'

dataservice/api/common/model.py
```diff
@@ -10,6 +10,9 @@
 from dataservice.extensions.flask_indexd import RecordNotFound
 from dataservice.api.common.id_service import uuid_generator, kf_id_generator
 
+COMMON_ENUM = {"Not Reported", "Not Applicable", "Not Allowed To Collect",
```
fiendish, line 13: I think this needs "Reported Unknown", maybe instead of "Not Available".

dataservice/api/phenotype/schemas.py
```diff
@@ -2,16 +2,23 @@
...
 from dataservice.api.phenotype.models import Phenotype
 from dataservice.api.common.schemas import BaseSchema
-from dataservice.api.common.validation import validate_age
+from dataservice.api.common.validation import (validate_age,
+                                               enum_validation_generator)
 from dataservice.extensions import ma
 
 
+OBSERVED_ENUM = {'Positive', 'Negative'}
+
```
fiendish, line 11: Why is this not 'True'/'False'?

dataservice/api/common/model.py
parimalak: we have both `Reported Unknown` and `Not Available`

dataservice/api/participant/schemas.py
parimalak: 👍

dataservice/api/phenotype/schemas.py
parimalak: To specify observed phenotype is positive or negative. For some cases it is expected that some feature needs to be present. if not, phenotype is  negatively observed . With `T/F` we miss that clarification.

dataservice/api/common/model.py
fiendish: I don't see reported unknown here

fiendish: 👍 I see it now

dataservice/api/genomic_file/schemas.py
```diff
@@ -1,17 +1,29 @@
...
 
 from dataservice.api.genomic_file.models import GenomicFile
-from dataservice.api.common.schemas import BaseSchema, IndexdFileSchema
+from dataservice.api.common.schemas import (BaseSchema, IndexdFileSchema,
+                                            AVAILABILITY_ENUM)
 from dataservice.api.common.custom_fields import PatchedURLFor
 from dataservice.extensions import ma
+from dataservice.api.common.validation import enum_validation_generator
+
+DATA_TYPE_ENUM = {'Submitted Aligned Reads',
```
dankolbman: I'm not sure if having the defaults on the data types makes sense:
- Not Applicable
- Not Available
- Not Reported
- Not Allowed To Collect

dankolbman: Also, are `Submitted` types just what we get from the source? Are there `harmonized` types too?

znatty22: I think in GDC the harmonized files just had the `Submitted` removed. But maybe we want to be more specific and call it `Harmonized Aligned Reads`, etc?

parimalak: in regards to defaults on data_type, for null values we at least have choice to fill. yeah, i did see for few columns where some of these defaults are unnecessary. Once, we get real data in some of these enums will be refactored

fiendish:
I would prefer the initial definition-fully-exists-before-runtime class-based approach rather than the sets of strings following commit dbdbec7af00cd9aef7ca1c01c117bb8787229b8a. Sets of strings means strings still have to be duplicated and copied around and an IDE won't warn if a string has a typo like it will when accessing a nonexistent class member. The class doesn't have to inherit from enum.

dataservice/api/sequencing_experiment/schemas.py
```diff
@@ -3,15 +3,32 @@
 from dataservice.api.sequencing_experiment.models import SequencingExperiment
 from dataservice.api.common.schemas import BaseSchema
 from dataservice.api.common.custom_fields import DateOrDatetime
-from dataservice.api.common.validation import validate_positive_number
+from dataservice.api.common.validation import (validate_positive_number,
+                                               enum_validation_generator)
 from dataservice.extensions import ma
 
+EXPERIMENT_STRATEGY_ENUM = {'WGS', 'WXS', 'RNA-Seq', 'Other'}
```
liberaliscomputing: I just had a discussion with @allisonheath and `miRNA-Seq` needs to be added since it should be distinguished from `RNA-Seq`.

parimalak: 👍

dataservice/api/common/model.py
fiendish, line 13: Needs "No Match" for (eg ontologies) when a specific domain exists but the correlated data does not match anything within the domain

parimalak: 👍 i think it is better to add `No Match` for the entities diagnosis, phenotype and biospecimens in their respective schemas

znatty22: I agree w @parimalak. `No Match` doesn't pertain to all enums so should not be included in common enums

dankolbman (approved):

dataservice/api/common/validation.py
```diff
@@ -36,3 +37,41 @@ def validate_kf_id(prefix, value):
     m = re.search(r, value)
     if not m:
         raise ValidationError('Invalid kf_id')
+
+
+class EnumValidator(OneOf):
```
znatty22: This class isn't doing anything differently than `OneOf` anymore, since you removed the part that made the validation case insensitive. I think we can probably get rid of this class and then wherever you have `EnumValidator` you can just replace that with `OneOf` (which I think is just in `enum_validation_generator`)

znatty22 (approved):
🥇

PR #305, opened 2018-05-04
✨ Move acl to root document

dankolbman:
Indexd has moved the acl field from within metadata to the root of the document object.


znatty22 (approved):
💯

parimalak (approved):
👍

dankolbman:
Tested in dev with indexd version:
```
{"commit": "f3cec9bac3d933eb1e11bac1f442360594a6f236", "version": "0.7.0-2-gf3cec9b"}
```
Which is the same version in qa and prd.

PR #309, opened 2018-05-04
🐛 Ignore mime type when parsing request body

dankolbman:
API was returning 500 whenever the `Content-Type` wasn't specified as `application/json`. This will force the body to be read as json and won't attempt to validate unexpected fields when there is no body.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -33,6 +33,14 @@ class TestAPI:
     and header checks
     """
 
+    @pytest.mark.parametrize('endpoint', ENDPOINTS)
+    def test_no_content_type(self, client, endpoint):
+        """ Test that no 500 is thrown when the content isnt specified """
+        resp = client.post(endpoint, data='{}')
+        assert resp.status_code < 500
+        resp = client.patch(endpoint, data='{}')
+        assert resp.status_code < 500
+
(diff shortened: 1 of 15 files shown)


tests/test_api.py
```diff
@@ -33,6 +33,14 @@ class TestAPI:
...
+    def test_no_content_type(self, client, endpoint):
+        """ Test that no 500 is thrown when the content isnt specified """
+        resp = client.post(endpoint, data='{}')
+        assert resp.status_code < 500
+        resp = client.patch(endpoint, data='{}')
+        assert resp.status_code < 500
+
     @pytest.mark.parametrize('endpoint,method,status_code',
                              [(ept, 'GET', 200) for ept in ENDPOINTS] +
                              [(ept+'/123', 'GET', 404) for ept in ENDPOINTS]
```
znatty22, line 46: Maybe also a test to ensure that mime-type isn't required would be good. In case someone creates a new resource and forgets to use request.get_json(force=True) for some reason. We'd catch that in tests

dankolbman: That is tested here (no headers in the post or patch)

znatty22: 👍

znatty22 (approved):
🥇

parimalak (approved):
👍

PR #312, opened 2018-05-07
🚑 Remove datamodel and migration version from status

dankolbman:
The working directory does not appear to be correct when docker runs the alembic command. We'll remove this for now to fix deployment, and figure out how to fix it later.


znatty22 (approved):

parimalak (approved):
👍

PR #316, opened 2018-05-08
🔧 Bucket secret default

dankolbman:
Won't try to load the bucket secret from vault if not in environment, or the secret in vault is empty.

### DIFF
--- a/config.py
+++ b/config.py
@@ -65,13 +65,13 @@ class ProductionConfig(Config):
         # Path for secrets in vault
         pg_secret = os.environ.get('DB_SECRET', 'secret/postgres')
         indexd_secret = os.environ.get('INDEXD_SECRET', 'secret/indexd')
-        bucket_secret = os.environ.get('BUCKET_SECRET', 'secret/bucket')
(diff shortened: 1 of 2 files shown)


config.py
```diff
@@ -90,7 +90,9 @@ def init_app(app):
         app.config['INDEXD_USER'] = indexd_secrets['data']['user']
         app.config['INDEXD_PASS'] = indexd_secrets['data']['password']
 
-        app.config['BUCKET_SERVICE_TOKEN'] = bucket_secrets['data']['token']
+        if bucket_secrets:
+            app.config['BUCKET_SERVICE_TOKEN'] = \
+                    bucket_secrets['data']['token']
```
znatty22: Will data and token keys always exist?

dankolbman: They should as long as bucket secrets comes back non-null

dankolbman: But I will make it more defensive

alubneuski (approved):

znatty22 (approved):
👍

Issue #313, opened 2018-05-07
Implement Basic Filter Parameters


znatty22:
In order to move forward with #104 we need to implement some basic support for filter parameters. 

Basic support means that users can pass in a list of query parameters in the query string of the request and those parameters will all be ANDed together to return a filtered result. Basic support will not include any other operators besides equal to for key, value parameters.

For example the request:
```
/participants?study_id=SD_00001111&is_proband=True
```
Will return all proband participants in study SD_00001111.

PR #317, opened 2018-05-08
✨ Basic Filter Params

znatty22:
Addresses #313 

A first attempt at implementation of basic filter parameters for all endpoints.

* Caveat - Current implementation does not querying for null values using filter parameters (even for fields which are allowed to be null). This may be due to a limitation of Python Requests which doesn't process query params with blank values. We may need to specify a special string to use for null and override Requests query string parser. Need to look into this further and potentially address in another PR

### DIFF
--- a/dataservice/api/common/schemas.py
+++ b/dataservice/api/common/schemas.py
@@ -162,21 +162,24 @@ def filter_schema_factory(model_filter_schema_cls):
     Allow partially populated schema
     Validate with strict=True - reuse model schema's validators
     """
+    # Exclude fields
     exclude = ('_links', )
     if hasattr(model_filter_schema_cls.Meta, 'exclude'):
         exclude += model_filter_schema_cls.Meta.exclude
 
-    return model_filter_schema_cls(strict=True, partial=True, exclude=exclude)
+    return model_filter_schema_cls(strict=True,
+                                   partial=True,
+                                   exclude=exclude)
 
 
 class FilterSchemaMixin(ma.ModelSchema):
     """
     Filter schema mixin inherited by all model filter schemas
     """
-
-    created_at = DateOrDatetime()
-    modified_at = DateOrDatetime()
     study_id = fields.Str()
+
+    class Meta:
+        dump_only = ()
 
     @validates('study_id')
     def valid(self, value):
--- a/dataservice/api/common/schemas.py
+++ b/dataservice/api/common/schemas.py
@@ -172,7 +172,7 @@ def filter_schema_factory(model_filter_schema_cls):
                                    exclude=exclude)
 
 
-class FilterSchemaMixin(ma.ModelSchema):
+class FilterSchemaMixin(ma.Schema):
     """
     Filter schema mixin inherited by all model filter schemas
     """
--- a/dataservice/api/common/schemas.py
+++ b/dataservice/api/common/schemas.py
@@ -1,7 +1,7 @@
-import re
 from dataservice.extensions import ma
 from marshmallow import (
     fields,
+    post_load,
     post_dump,
     pre_dump,
     validates_schema,
@@ -11,6 +11,8 @@ from marshmallow import (
 from flask import url_for, request
 from flask_marshmallow import Schema
 from dataservice.api.common.pagination import Pagination
+from dataservice.api.common.custom_fields import DateOrDatetime
+from dataservice.api.common.validation import validate_kf_id
 from dataservice.extensions import db
 
 
@@ -39,11 +41,7 @@ class BaseSchema(ma.ModelSchema):
 
     @validates('kf_id')
     def valid(self, value):
-        prefix = self.Meta.model.__prefix__
-        r = r'^'+prefix+r'_[A-HJ-KM-NP-TV-Z0-9]{8}'
-        m = re.search(r, value)
-        if not m:
-            raise ValidationError('Invalid kf_id')
+        validate_kf_id(self.Meta.model.__prefix__, value)
 
     @post_dump(pass_many=True)
     def wrap_envelope(self, data, many):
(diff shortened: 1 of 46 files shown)


dataservice/api/common/schemas.py
```diff
@@ -156,6 +154,42 @@ class PaginatedSchema(Schema):
...
+    exclude = ('_links', )
+    if hasattr(model_filter_schema_cls.Meta, 'exclude'):
+        exclude += model_filter_schema_cls.Meta.exclude
+
+    return model_filter_schema_cls(strict=True,
+                                   partial=True,
+                                   exclude=exclude)
+
+
+class FilterSchemaMixin(ma.ModelSchema):
```
dankolbman: The base Marshmallow `Schema` may be a better parent class here?

dataservice/api/common/schemas.py
```diff
@@ -156,6 +154,42 @@ class PaginatedSchema(Schema):
...
+
+    class Meta:
+        dump_only = ()
+
+    @validates('study_id')
+    def valid(self, value):
+        validate_kf_id('SD', value)
+
+    @post_load
+    def make_instance(self, data):
```
dankolbman, line 196: What does this do?

tests/test_filter_params.py
```diff
@@ -0,0 +1,422 @@
...
+            'file_format': 'fastq',
+            'size': 1000,
+            'urls': ['s3://bucket/key'],
+            'hashes': {'md5': str(uuid.uuid4())},
+            'controlled_access': False
+        },
+        '/cavatica-tasks': {'name': 'CavaticaTask_0'}
+
+    },
+    'filter_params': {
```
dankolbman: Can't this be constructed automatically?

dataservice/api/common/schemas.py
znatty22: When you use use_args from webargs, it does a schema load. So if you want it to return the dict instead of the object after the load, you just return the data which is a dict.

dataservice/api/common/schemas.py
znatty22: Hmm yea probably...I can't remember why I decided to use ModelSchema anymore. I'll change it to `Schema`

tests/test_filter_params.py
znatty22: Not sure what you mean?

tests/test_filter_params.py
```diff
@@ -0,0 +1,562 @@
+import pytest
```
znatty22, line 1: I didn't want to make too many testing changes in this PR but ...

I think I might be able to refactor test_api.py and have it use the same set of invalid fields in the entities fixture here. Would probably pull the entities fixture out of here and into somewhere else, maybe conftest? Also may be able to refactor test_pagination.py to use this.

dataservice/api/common/schemas.py
dankolbman: Isn't `post_load` just a hook? Or does it override something?

tests/test_filter_params.py
dankolbman: I mean you could generate a dict of endpoints with their fields based on the above dictionary.

tests/test_filter_params.py
```diff
@@ -0,0 +1,562 @@
...
+}
+
+
+class TestFilterParams:
+    """
+    Test filtering on pagnation endpoints
+    """
+
+    @pytest.fixture(scope='module')
+    def entities(client):
```
dankolbman: What's blocking this from being merged with the entities fixture in `conftest`?

tests/test_filter_params.py
```diff
@@ -0,0 +1,562 @@
...
+                ent.biospecimen = bs0
+
+            # CavaticaTask
+            for ent in entities[CavaticaApp]:
+                ent.cavatica_app = ca0
+
+            db.session.commit()
+
+    @pytest.mark.parametrize('model',
+                             [
```
dankolbman: I think you could replace this list with `ENTITY_ENDPOINT_MAP.values()`

tests/test_pagination.py
```diff
@@ -218,21 +217,59 @@ def test_study_filter(self, client, participants,
...
         resp = client.get(endpoint)
         resp = json.loads(resp.data.decode('utf-8'))
 
         assert len(resp['results']) == 0
         assert resp['limit'] == 10
         assert resp['total'] == 0
 
+    @pytest.mark.parametrize('study_id', ['blah', 3489, 'PT_00001111'])
+    @pytest.mark.parametrize('endpoint', [
+        ('/participants'),
```
dankolbman: Can this be replaced with `ENDPOINTS` from conftest?

dataservice/api/common/schemas.py
```diff
@@ -156,6 +154,39 @@ class PaginatedSchema(Schema):
     return PaginatedSchema
 
 
+def filter_schema_factory(model_filter_schema_cls):
```
dankolbman: Why not take the base schema and construct the filter schema here dynamically? That way you don't need to create an empty `*FilterSchema` for everything.

dataservice/api/common/schemas.py
znatty22: So it is a little obscure. `post_load` is a hook, but `make_instance` is also a method in the base class which does the deserialization from the data dict to the model. So i'm just having it override `make_instance` to return the dict, but doing this after load so that validation still happens. Seems confusing to me bc you would think that the deserialization has already happened post load, but this works. I found the suggestion here on webargs github:

Idk if there is a better way to do this?

https://github.com/sloria/webargs/issues/126

tests/test_filter_params.py
znatty22: Maybe, but I want to tightly control the set of filter_params for each endpoint, so I'm manually defining the dict

dataservice/api/common/schemas.py
znatty22: Well I wasn't sure exactly how to dynamically define classes otherwise yea thats exactly what I would want to do. Let me look into it

dankolbman: I think it would be something like `FilterClass = type('MyFilterClass', (base_cls, FilterSchemaMixin), {})`

tests/test_pagination.py
znatty22: 👍

tests/test_filter_params.py
znatty22: yep 👍

tests/test_filter_params.py
znatty22: The tests in test_api.py that depend on that need the foreign keys to parent entities tracked. The entities in test_filter_params don't have that yet. I think when I go to test the refactor child links PR, I may end up refactoring and merging with the entities fixture in conftest.

tests/conftest.py
```diff
@@ -26,136 +28,61 @@
...
+    SequencingExperiment: '/sequencing-experiments',
+    CavaticaTask: '/cavatica-tasks',
+    CavaticaTaskGenomicFile: '/cavatica-task-genomic-files'
+}
+
+ENDPOINT_ENTITY_MAP = {v: k for k, v in ENTITY_ENDPOINT_MAP.items()}
+
+ENDPOINTS = list(ENTITY_ENDPOINT_MAP.values())
+
+ENTITY_PARAMS = {
```
dankolbman: Can we move this to an external json file?

dankolbman (commented):
Can you clean up commit history? I think most of them have been reverted.

tests/conftest.py
znatty22: But then we can't call functions to generate values (right now just uuids and dates, but might be more later)

Should I move all the constants into their own file like `tests/params.py`??

dankolbman: It looks like the md5 hash is the only thing calling a function now? It could probably be hard-coded.

dankolbman (approved):
💯

parimalak (approved):
👍

dankolbman:
lgtm 👍

PR #334, opened 2018-05-14
✏️ Update genomic file data type enum

dankolbman:


znatty22 (approved):

dataservice/api/genomic_file/schemas.py
```diff
@@ -9,9 +9,12 @@
 
 DATA_TYPE_ENUM = {'Aligned Reads',
                   'Aligned Reads Index',
+                  'Unaligned Reads',
                   'Simple Nucleotide Variation',
                   'Variant Calls',
-                  'Individual Variant Calls',
```
parimalak, line 14: are we re-mapping `Individual Variant Calls` to `Variant Calls` ? bc in QA, i see 2186 genomic files with data_type `Individual Variant Calls`.

parimalak (approved):
👍

dataservice/api/genomic_file/schemas.py
dankolbman: yep

PR #338, opened 2018-05-14
✨Add ontology prefix validation

parimalak:
validation for ontology columns

### DIFF
--- a/dataservice/api/common/validation.py
+++ b/dataservice/api/common/validation.py
@@ -1,6 +1,8 @@
 import re
 from marshmallow import ValidationError
-from marshmallow.validate import OneOf
+from marshmallow.validate import OneOf, Validator
+
+from marshmallow.fields import Str
 
 MIN_AGE_DAYS = 0
 # Max value chosen due to HIPAA de-identification standard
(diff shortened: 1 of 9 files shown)


dataservice/api/common/validation.py
```diff
@@ -47,3 +49,53 @@ def enum_validation_generator(_enum):
...
+        """Perform validation on ``value``. Raise a :exc:`ValidationError`
+        if validation
+        does not succeed.
+        """
+        errors = []
+        kwargs = {}
+        for validator in self.validators:
+            try:
+                try:
+                    r = validator(value, field=self)
```
parimalak, line 69: we can remove this `FieldValidator` class once this PR https://github.com/marshmallow-code/marshmallow/issues/289 for marshmallow is merged. in this we are extending `_validate` function to allow field name to the custom validator function

dataservice/api/common/validation.py
```diff
@@ -47,3 +49,53 @@ def enum_validation_generator(_enum):
...
+    onto_dict = {
+        'mondo': 'MONDO:',
+        'uberon': 'UBERON:',
+        'icd':	'ICD10:',
+        'hpo':	'HP:',
+        'snomed': 'SNOMEDCT:',
+        'ncit': 'NCIT:'
+    }
+    if attribute is not None:
+        if attribute.startswith(onto_dict[prefix]) is False:
```
dankolbman, line 99: Probably best to check that prefix is in the onto_dict?

dataservice/api/diagnosis/schemas.py
```diff
@@ -20,6 +22,15 @@ class DiagnosisSchema(BaseSchema):
...
                                        DIAGNOSIS_CATEGORY_ENUM))
+    mondo_id_diagnosis = FieldValidator(attribute='mondo_id_diagnosis',
+                                        validate=validate_ontology_id_prefix)
+    icd_id_diagnosis = FieldValidator(attribute='icd_id_diagnosis',
+                                      validate=validate_ontology_id_prefix)
+    uberon_id_tumor_location = FieldValidator(
+        attribute='uberon_id_tumor_location',
+        validate=validate_ontology_id_prefix)
+    ncit_id_diagnosis = FieldValidator(attribute='ncit_id_diagnosis',
+                                       validate=validate_ontology_id_prefix)
```
dankolbman, line 37: maybe `validate_ontology_id_prefix` could be a generator that takes a prefix and returns a function that validates that field using the prefix? That way lookup dict won't be needed and you won't have to pass attributes to it.

znatty22: 👍 I agree thats what I would suggest as well ^

PR #351, opened 2018-05-23
🔧 Increase number of workers

dankolbman:
We are currently maxing out our containers at %50 CPU, so we should easily accommodate another worker. This is the CPU load from highly concurrent requests:
screen shot 2018-05-23 at 8 56 48 am


znatty22 (approved):
💯

parimalak (approved):
👍

PR #352, opened 2018-05-23
✨ Read groups

### LINKED ISSUE #343: read group metadata for source file in FASTQ format
The genomics harmonization team needs to know the read group information such as `paired_end*, read_group_name*, library_name, flow_cell_barcode, platform, sequencing_center, read_length etc` (* are required) for fastq inputs to be able to run the alignment workflow properly. Required a data model for `Read Group` like the one [GDC](https://docs.gdc.cancer.gov/Data_Dictionary/viewer/#?view=table-definition-view&id=read_group) has. 

Currently most GMKF sequencing center provided BAM file which has the Read Group information embed as part of the file header, but the BAMs that HudsonAlpha provided are with incorrect header, we are planning to run the workflow from FASTQ for those data. In addition to that, about 40% of the CBTTC source data are in FASTQ format. To have this model ready, so help the system to have the ability to take FASTQ files from other data generators. 

Suggested model:
- GDC: https://docs.gdc.cancer.gov/Data_Dictionary/viewer/#?view=table-definition-view&id=read_group
- CBTTC: https://github.com/d3b-center/cbttc-bank-seq/blob/master/sqlite/initial_csv/cbttc_fastqs.csv

dankolbman:
First pass at a model to store fastq metadata.

Closes #343


dataservice/api/read_group/models.py
```diff
@@ -0,0 +1,31 @@
...
+    :param flow_cell: The identifier for the group's flow cell
+    :param lane_number: The group's lane
+    :param quality_scale: The quality score encoding of the fastq file
+    """
+    __tablename__ = 'read_group'
+    __prefix__ = 'RG'
+
+    external_id = db.Column(db.Text(), nullable=True,
+                            doc='Name given to read group by the contributor')
+    paired_end = db.Column(db.Integer(),
```
parimalak, line 20: We have `is_paired_end` in `sequencing_experiment`. does `paired_end` stores the same information ?

yuankunzhu: typical alignment tools need to know if a read is the read-pair-group 1 or read-pair-group 2, and take that for the proper mapping.

dataservice/api/read_group/schemas.py
```diff
@@ -0,0 +1,44 @@
+from marshmallow_sqlalchemy import field_for
+
+from dataservice.api.common.schemas import BaseSchema
+from dataservice.api.common.custom_fields import DateOrDatetime
```
parimalak: i think this import is unused

dataservice/api/read_group/models.py
parimalak: 👍

dataservice/api/read_group/schemas.py
```diff
@@ -0,0 +1,44 @@
...
+from dataservice.extensions import ma
+
+PAIRED_END_ENUM = {1, 2}
+QUALITY_SCALE_ENUM = {'Illumina13', 'Illumina15', 'Illumina18',
+                      'Solexa', 'Sanger'}
+
+
+class ReadGroupSchema(BaseSchema):
+
+    paired_end = field_for(ReadGroup, 'paired_end',
```
znatty22, line 16: Based on yuankun's comment above, we should do a `validate_positive_number` here right?

dataservice/api/genomic_file/models.py
```diff
@@ -55,6 +56,12 @@ class GenomicFile(db.Model, Base, IndexdFile):
     biospecimen_id = db.Column(KfId(), db.ForeignKey('biospecimen.kf_id'),
                                nullable=True)
 
+    read_group = db.relationship(ReadGroup,
```
znatty22, line 59: First want to make sure I understand the relationship between genomic file and read group properly. From Yuankun's description in #343, sounds like there is a 1 genomic_file --> many read_groups relationship? If so, then this attribute should be `read_groups`.

tests/data.json
```diff
@@ -218,6 +225,19 @@
...
             ]
         },
+        "/read-groups": {
+            "valid": {
+                "paired_end": 1,
+                "lane_number": 3,
+                "flow_cell": "FL123",
+                "quality_scale": "Illumina15"
+            },
+            "invalid": [
```
znatty22, line 236: Maybe good to include negative number values here too for the attrs that need to be positive numbers

migrations/versions/bba7ff44701d_.py
```diff
@@ -0,0 +1,44 @@
+"""
+1.1.0 - Add Read Group model
```
parimalak: are we adding release numbers in the migrations?

dataservice/api/genomic_file/models.py
allisonheath: That's correct - there's often many read groups that get combined into one file.

dankolbman: I don't believe the intent here is to associate all read groups to their resulting bam but rather to link any read group metadata a fastq has back to it. So I'm not sure we will be storing relationships between bam genomic files and read groups.

dataservice/api/read_group/schemas.py
dankolbman: We use the `PAIRED_END_ENUM` which only allows values of `1` or `2` so it's much more strict than just a `validate_positive_number`.

znatty22: 👍

znatty22 (approved):
Looks good! Could you also update the draw.io diagram?

parimalak (approved):
👍

znatty22 (commented):
I just thought of something.. remember to go into genomic file schema and add the read groups to the gf links:         
'read_groups': ma.URLFor('api.read_group_list',
                                   genomic_file_id='<kf_id>')

dataservice/api/genomic_file/models.py
znatty22: I'm revisiting this PR bc I see a couple issues. I'm still confused on the relationship here. Is there a 1-1 relationship between genomic file and read group or is it 1 to many? If it is 1 to 1, then you probably want to add `uselist=False` parameter to the relationship. However, if it really is 1-1, then why not make these fields part of genomic file entity? If it is NOT 1-1, then this attribute really should be named `read_groups`. It looks like this might just be a typo since everywhere else in the code the attribute is referred to as `read_groups`.

dataservice/api/genomic_file/schemas.py
```diff
@@ -58,5 +58,7 @@ class Meta(BaseSchema.Meta, IndexdFileSchema.Meta):
             'api.sequencing_experiments',
             kf_id='<sequencing_experiment_id>'),
         'cavatica_task_genomic_files': ma.URLFor(
-            'api.cavatica_task_genomic_files_list', genomic_file_id='<kf_id>')
+            'api.cavatica_task_genomic_files_list', genomic_file_id='<kf_id>'),
+        'read_groups': ma.URLFor(
```
znatty22: Again, I'm not sure whether the link name should be `read_groups` or `read_group`. It depends on how the relationship should be defined.

dataservice/api/read_group/schemas.py
```diff
@@ -0,0 +1,43 @@
...
+
+    genomic_file_id = field_for(ReadGroup, 'genomic_file_id',
+                                required=True, load_only=True)
+
+    class Meta(BaseSchema.Meta):
+        resource_url = 'api.read_groups'
+        collection_url = 'api.read_groups_list'
+        model = ReadGroup
+        exclude = (BaseSchema.Meta.exclude +
+                   ('read_groups', ) +
```
znatty22: Remove the `'read_groups'` attribute since its not actually an attribute on the ReadGroup model. Change `'genomic_files' ` to `'genomic-file'` since a read group has 1 genomic_file as its parent.

znatty22 (changes requested):

dataservice/api/genomic_file/models.py
dankolbman: Aligned Reads are compiled from fastq files divided by Read Group, so the relation is 1-1. Not all files are fastq with information about the Read Group, which is why it is preferable to store in a different table.

tests/read_group/test_read_group_model.py
```diff
@@ -0,0 +1,176 @@
...
+        self.assertEqual(GenomicFile.query.count(), 1)
+        self.assertEqual(ReadGroup.query.count(), 1)
+        self.assertEqual(ReadGroup.query.first().genomic_file, gf)
+
+    def test_no_gf(self):
+        """
+        Test creation of read_group without a genomic_file
+        """
+        rg = ReadGroup()
+        self.assertRaises(IntegrityError, db.session.add(rg))
```
znatty22: This test will always pass
Change to:
```
db.session.add(rg)
        db.session.commit()
```

tests/read_group/test_read_group_model.py
```diff
@@ -0,0 +1,176 @@
...
+        db.session.commit()
+
+        self.assertEqual(ReadGroup.query.count(), 1)
+
+        db.session.delete(rg)
+        db.session.commit()
+
+        self.assertEqual(ReadGroup.query.count(), 0)
+
+    def test_delete_orphans(self):
```
znatty22, line 100: Maybe also add a test to ensure the DB cascading delete works. If you do db.session.delete, it will use the ORM's cascading delete functionality. If you do GenomicFile.query.filter_by(kf_id=<gf id>).delete() it will use the DB's delete and then you can check if the cascade worked

tests/conftest.py
```diff
@@ -163,17 +162,18 @@ def entities(client):
...
             db.session.add(r)
 
-        # Cavatica task genomic files
+        # Cavatica task genomic files and read group
         for i, (ct, gf) in enumerate(zip(
                 _entities[CavaticaTask], _entities[GenomicFile])):
             is_input = True
             if i % 2 == 0:
                 is_input = False
+                rg = ReadGroup(genomic_file_id=gf.kf_id)
```
znatty22: Remove these 3 lines here. Read groups will get created above in the main entity creation loop (unless you were to put it in the list of skipped entities which currently has FamilyRelationship and CavaticaTaskGenomicFile). Fam rels and ctgfs are special entities (association tables) that have to be created after all the main entities are created since they require two entities before they can be created. Fam rel needs two participants, Ctgf needs cav task and genomic file.

tests/test_api.py
```diff
@@ -85,6 +85,7 @@ def test_status_format(self, client, endpoint, method):
         ('/biospecimens', ['participant', 'sequencing_center']),
         ('/sequencing-experiments', ['sequencing_center']),
         ('/genomic-files', ['biospecimen', 'sequencing_experiment']),
```
znatty22: Read group link is formatted like a parent link so should be added to the parent link tests for genomic file: 
```
('/genomic-files', ['biospecimen', 'sequencing_experiment', 'read_group']),
```

tests/test_pagination.py
```diff
@@ -14,6 +14,7 @@
 from dataservice.api.diagnosis.models import Diagnosis
```
znatty22, line 14: I think read group is missing from the `test_pagination` method

tests/read_group/test_read_group_model.py
dankolbman: It won't pass when the non-null constraint isn't on the genomic_file relation?

znatty22: It'll pass regardless of whether you have a valid foreign key genomic_file_id on the read group. If you were to add this before the test: 
```
self._create_entities()
gf = GenomicFile.query.first()
rg = ReadGroup(genomic_file_id=gf.kf_id)
```
this won't throw an IntegrityError since there is a valid foreign key and the test will still pass. I think we want a test that tests an IntegrityError is only thrown when it should be?

dankolbman: If it doesn't throw an IntegretyError, it won't pass?

znatty22 (approved):
Looks good! 💯

PR #355, opened 2018-05-29
✨Add dbgap_consent_code to biospecimen

### LINKED ISSUE #327: Add dbgap consent  code to biospecimen
- [ ] Add dbgap_consent_code to biospecimen model
- [ ] Update tests and migration
- [ ]  Get consent code from dbgap xml and concatenate with dbgap study_id ; store the result in db_consent_code and update gen3 ACL's of gen_file and study_file

parimalak:
This PR adds the field to the biospecimen entity.  Planning to do the xml scrapping and automatic updation of field through api in another PR; as this script needs to run outside of api one in study_imports(for initial data) and as a lambda function(or as an independent task) that runs everyday and scraps the xml for the dbgap_consent_code.
closes #327


znatty22 (approved):
Should we add validation to this field in the future (formatting, etc)?

allisonheath:
Yeah, we should add validation, but that can be a separate ticket. ~We do need to decide if we're storing it with the `c` prefix or not.~ I see now we're including both the dbgap code and the consent code in the same manner that will be used via https://github.com/uc-cdis/fence/issues/219

allisonheath (approved):

PR #357, opened 2018-05-31
✏️ Fix migration version formatting

dankolbman:


znatty22 (approved):

migrations/versions/8f1ad897fb0f_.py
```diff
@@ -1,5 +1,5 @@
 """
-1.0.1 Add dbgap_consent_code to biospecimen
+1.0.1 - Add dbgap_consent_code to biospecimen
```
parimalak, line 2: i am just wondering we have different releases for migrations and code base. our code base now release of 1.1.2. are we gonna correlate these two with the first suffix _1_.1.2 and _1_.0.1 ?

parimalak (approved):

migrations/versions/8f1ad897fb0f_.py
dankolbman: Probably not. I think the first suffix will need to change whenever we introduce non-compatible changes to the model. So if we drop a field, we should bump the primary version.

PR #358, opened 2018-05-31
🔧 Increase worker count

dankolbman:
Should be enough to max our cpu usage on an ecs task

### DIFF
--- a/bin/supervisord.conf
+++ b/bin/supervisord.conf
@@ -25,6 +25,6 @@ stdout_logfile=/dev/stdout
 stdout_logfile_maxbytes=0
 
 [program:gunicorn]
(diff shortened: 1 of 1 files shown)


bin/supervisord.conf
```diff
@@ -25,6 +25,6 @@ stdout_logfile=/dev/stdout
 stdout_logfile_maxbytes=0
 
 [program:gunicorn]
-command=gunicorn manage:app -b localhost:5000 --workers 2
+command=gunicorn manage:app -b localhost:5000 --workers 5
```
znatty22, line 28: How'd you determine the number?

dankolbman: Informed guessing

znatty22 (approved):
👍 Lets give it a shot!

PR #359, opened 2018-05-31
✨ Add relation from biospecimen to diagnosis

### LINKED ISSUE #314: Maintain association of Biospecimen to Diagnosis
For the CBTTC data there are a number of cases where there are multiple tumor occurrences for the same patient, each with their own set of diagnoses and associated set of biospecimens acquired from a surgical event. RIght now, this relationship is lost in the current data model. I would suggest that we directly link biospecimen to diagnosis, but open to other ideas on how best to keep this relationship. Here are a few specific examples in the current available data, about nine participants exist currently with longitudinal genomic data, with about a hundred to come in the fall:

C21402:
- 7316-302 is Medulloblastoma, age at diagnosis of 1384
- 7316-2148 is Meningioma, age at diagnosis of 3504

C25830:
- 7316-278 is Medulloblastoma, age at diagnosis of 5534
- 7316-95 is Medulloblastoma, age at diagnosis of 6555
- 7316-333 is recurrent Medulloblastoma, age at diagnosis of 6992

dankolbman:
Closes #314


dataservice/api/biospecimen/schemas.py
```diff
@@ -49,5 +50,7 @@ class Meta(BaseSchema.Meta):
         'sequencing_center': ma.URLFor('api.sequencing_centers',
                                        kf_id='<sequencing_center_id>'),
         'genomic_files': ma.URLFor('api.genomic_files_list',
+                                   biospecimen_id='<kf_id>'),
+        'diagnoses': PatchedURLFor('api.diagnoses_list',
```
znatty22: I think you're ok using regular URLFor here since kf_id will never be null. PatchedURLFor is only when a foreign key could be null. The Biospecimen - Diagnosis relationship is exactly like Family - Participant. See [FamilySchema](https://github.com/kids-first/kf-api-dataservice/blob/27a4c0a73d4de1521fc50f62e5ec2de069519fa2/dataservice/api/family/schemas.py#L17) for reference.

dataservice/api/biospecimen/models.py
```diff
@@ -102,3 +103,7 @@ class Biospecimen(db.Model, Base):
                                     backref=db.backref(
                                         'biospecimen', lazy=True),
                                     doc='genomic files this biospecimen')
+    diagnoses = db.relationship(Diagnosis,
```
znatty22, line 106: Do we really want this delete cascade?  If you delete a biospecimen, the corresponding diagnoses might still be attached to the participant and we may still want them stored.

znatty22 (commented):
So right now it seems that someone could add two diagnoses from different participants to the same biospecimen. Maybe we could put an ORM event on the biospecimen to ensure that when diagnoses are added they share the same participant kf_id?

dataservice/api/biospecimen/schemas.py
```diff
@@ -49,5 +50,7 @@ class Meta(BaseSchema.Meta):
         'sequencing_center': ma.URLFor('api.sequencing_centers',
                                        kf_id='<sequencing_center_id>'),
         'genomic_files': ma.URLFor('api.genomic_files_list',
-                                   biospecimen_id='<kf_id>')
+                                   biospecimen_id='<kf_id>'),
+        'diagnoses': ma.URLFor('api.diagnoses_list',
```
dankolbman, line 54: Won't this not exist for many entries? Shouldn't it be patched?

dataservice/api/diagnosis/models.py
```diff
@@ -66,3 +68,48 @@ class Diagnosis(db.Model, Base):
...
+        """
+        Ensure that both the diagnosis and biospecimen
+        (referred to by biospecimen_id) have the same participant
+
+        If this is not the case then raise DatabaseValidationError
+        """
+        from dataservice.api.biospecimen.models import Biospecimen
+        from dataservice.api.errors import DatabaseValidationError
+
+        if ((self.kf_id is None) or
```
dankolbman: Returning the value if kf_id is None means that we can avoid this constraint when creating a new Diagnosis?

dataservice/api/biospecimen/schemas.py
znatty22: We would have to use PatchedURLFor if there was ever a chance that `kf_id` would be null. But since `kf_id` refers to the biospecimen itself, it will never be null. Maybe you're thinking of the case for the parent link (i.e. Family and participant, where in the participant's link to family, the family_id/foreign key could be null)?

dataservice/api/diagnosis/models.py
znatty22: Eh yea, this doesn't look right. The original thought was to do a check for self.kf_id and self.participant_id being null and circumvent the constraint logic if True. During testing, I noticed what I believe to be a timing issue. It seems that when diagnosis is initialized these two fields are set to null and the DatabaseValidationError exception gets thrown. I think what i want here is to check it `((self.kf_id == None and self.participant_id == None) or value == None)` and if True, return None.

dankolbman: `kf_id` will be `None` until the session is flushed. `participant_id` should  not be `None`, unless there is an issue with the tests, though?

migrations/versions/5988f77859e7_.py
```diff
@@ -0,0 +1,31 @@
+"""Add relationship between diagnosis and biospecimen
```
dankolbman: Can you add version here? I think it's `1.0.2`

dataservice/api/diagnosis/models.py
```diff
@@ -66,3 +68,49 @@ class Diagnosis(db.Model, Base):
...
+        """
+        Ensure that both the diagnosis and biospecimen
+        (referred to by biospecimen_id) have the same participant
+
+        If this is not the case then raise DatabaseValidationError
+        """
+        from dataservice.api.biospecimen.models import Biospecimen
+        from dataservice.api.errors import DatabaseValidationError
+
+        if ((self.kf_id is None and self.participant_id is None) or
```
dankolbman: If `participant_id` is none, won't the non-nullable `participant_id` be invalid?
Does this get in the way of that error? Why do you need to know the `kf_id` is not `None` here?

migrations/versions/5988f77859e7_.py
znatty22: Should I wait until you merge read groups #352 in? If so, will it then be 1.1.1?

dataservice/api/diagnosis/models.py
znatty22: This check has to do with that weird behavior I was showing you with the tests randomly failing. It seems like that happens in marshmallow's deserialization code. I'm wondering if there's a timing issue with the attribute validation event? Bc I can't figure out why the tests pass sometimes but not other times and I can't find a pattern. I think deserialization initializes an empty mapped object and then populates the object's fields later. When the object is initialized, I think the attribute validation event gets called and you then run into the situation where self.kf_id and self.participant_id are both None. Having this logic in the if statement seems to fix the problem and give me consistent behavior in the tests...

I was going to try changing from using attribute validation events to both before_insert and before_update ORM events, but then decided not to, since the current logic seems to work? Plus I want to avoid using two events.

parimalak (approved):
👍

dataservice/api/diagnosis/models.py
```diff
@@ -66,3 +69,65 @@ class Diagnosis(db.Model, Base):
...
+             'refers to participant {} and '
+             'biospecimen {} refers to participant {}')
+            .format(kf_id,
+                    target.participant_id,
+                    bsp.kf_id,
+                    bsp.participant_id))
+        raise DatabaseValidationError(target_entity, operation, message)
+
+
+@event.listens_for(Diagnosis, 'before_insert')
```
znatty22, line 120: Use ORM object events instead of object attribute events (i.e. set or validate). Saw weird and unpredictable behavior with attribute events during deserialization of JSON to ORM objects. 

Probably better to use `before_insert`, `before_update` events anyway because we only care about validating biospecimen_id before the commit happens. The frequency of occurrence on the attribute events is much greater and we don't need to check the value of biospecimen_id every time its changed.

parimalak (approved):
💯

znatty22 (approved):

PR #361, opened 2018-06-07
🐛 Fix repeated loading of jinja tempaltes

dankolbman:
Loading templates from the filesystem repeatedly was causing issues with concurrent reads from many gunicorn workers. This was possibly preventing gunicorn from starting > 1 worker.


parimalak (approved):

znatty22 (approved):
👍

PR #365, opened 2018-06-12
✨ Patch acl

dankolbman:
Only creates a new version of a document when the `size` or `hashes` property changes. Otherwise modifies the existing document.
Also changes the acl of all versions of a document when the acl is changed.

Test script used to confirm that all versions are updated:

```
export API=localhost:5000

echo ""
echo "Make new GF"
export RESP=`curl -s -XPOST $API/genomic-files -d '{"file_name": "test", "acl": ["TEST"], "size": 1234, "hashes": {"md5": "f5d566687637ba7b0e529f465b102693"}, "urls": ["s3://test/key"]}'`
echo $RESP | jq '.results.kf_id' --raw-output
echo $RESP | jq '.results.latest_did' --raw-output
export GFID=`echo $RESP | jq '.results.kf_id' --raw-output`

echo ""
echo "Update GF size and hash"
export RESP=`curl -H "Content-Type: application/json" -XPATCH -s $API/genomic-files/$GFID -d '{"file_name": "test", "acl": ["TEST"], "size": 5678, "hashes": {"md5": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"}, "urls": ["s3://test/key"]}'`
echo $RESP | jq '.results.kf_id' --raw-output
echo $RESP | jq '.results.latest_did' --raw-output

echo ""
echo "Update GF acl"
export RESP=`curl -H "Content-Type: application/json" -s -XPATCH $API/genomic-files/$GFID -d '{"file_name": "test", "acl": ["TEST", "TEST2"], "size": 5678, "hashes": { "md5": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"}, "urls": ["s3://test/key"]}'`
echo $RESP | jq '.results.kf_id' --raw-output
echo $RESP | jq '.results.latest_did' --raw-output

echo ""
echo "Update GF name"
export RESP=`curl -H "Content-Type: application/json" -s -XPATCH $API/genomic-files/$GFID -d '{"file_name": "test revision", "acl": ["TEST", "TEST2"], "size": 5678, "hashes": { "md5": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"}, "urls": ["s3://test/key"]}'`
echo $RESP | jq '.results.kf_id' --raw-output
echo $RESP | jq '.results.latest_did' --raw-output
```

### DIFF
--- a/dataservice/api/common/model.py
+++ b/dataservice/api/common/model.py
@@ -2,7 +2,8 @@ from datetime import datetime
 from flask import abort
 from requests.exceptions import HTTPError
 import sqlalchemy.types as types
-from sqlalchemy import event
+from sqlalchemy import event, inspect
+from sqlalchemy.orm import reconstructor
 from sqlalchemy.ext.declarative import declared_attr
 from sqlalchemy.dialects.postgresql import UUID
 
@@ -41,6 +42,33 @@ class IDMixin:
     uuid = db.Column(UUID(), unique=True, default=uuid_generator)
 
 
+class IndexdField():
+
+    def __init__(self, value):
+        self._value = value
+
+    def __get__(self, instance, owner=None):
+        return self._value
+
+    def __set__(self, instance, value):
+        """
+        If the value is being changed on an object that is already being
+        stored in the database, the modifed_at column will be explicitly
+        update. This must be done so that the object in the database is
+        update and the ORM update event is triggered. Otherwise, the event
+        will not be triggered and the object will not be updated in Indexd.
+
+        NB: The value of the field may only be *set* through assignment.
+        Collection functions such as list.append() or dict.__setitem__()
+        will not trigger this hook when invoked.
+        """
+        if instance and not value == self._value:
+            state = inspect(instance)
+            if state.persistent:
+                instance.modified_at = datetime.now()
+        self._value = value
+
+
 class IndexdFile:
     """
     Field reflection for objects that are stored in indexd
@@ -77,15 +105,16 @@ class IndexdFile:
     latest_did = db.Column(UUID(), nullable=False)
 
     # Fields used by indexd, but not tracked in the database
-    file_name = ''
-    urls = []
+    file_name = IndexdField('')
+    urls = IndexdField([])
     rev = None
-    hashes = {}
-    acl = []
+    hashes = IndexdField({})
+    acl = IndexdField([])
     # The metadata property is already used by sqlalchemy
-    _metadata = {}
-    size = None
+    _metadata = IndexdField({})
+    size = IndexdField(None)
 
+    @reconstructor
     def merge_indexd(self):
         """
         Gets additional fields from indexd
--- a/dataservice/api/genomic_file/resources.py
+++ b/dataservice/api/genomic_file/resources.py
@@ -108,12 +108,6 @@ class GenomicFileAPI(CRUDView):
             abort(404, 'could not find {} `{}`'
                   .format('genomic_file', kf_id))
 
-        # Merge will return None if the document wasnt found in indexd
-        merge = genomic_file.merge_indexd()
-        if merge is None:
-            abort(404, 'could not find {} `{}`'
-                  .format('genomic_file', kf_id))
-
         sch = GenomicFileSchema(many=False)
         return sch.jsonify(genomic_file)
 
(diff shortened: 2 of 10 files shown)


dataservice/extensions/flask_indexd.py
```diff
@@ -139,6 +139,7 @@ def update(self, record):
 
         :param record: A file-like object
         """
+        print('update')
```
parimalak: 💣

tests/mocks.py
```diff
@@ -93,6 +93,23 @@ def get(self, url, *args, **kwargs):
 
         return MockResp(resp=resp, status_code=self.status_code)
 
+    def get(self, url, *args, **kwargs):
```
parimalak: i see another get function with same definition.

dataservice/api/common/model.py
```diff
@@ -77,15 +105,16 @@ class IndexdFile:
...
-    acl = []
+    hashes = IndexdField({})
+    acl = IndexdField([])
     # The metadata property is already used by sqlalchemy
-    _metadata = {}
-    size = None
+    _metadata = IndexdField({})
+    size = IndexdField(None)
 
+    @reconstructor
```
znatty22, line 117: Now that you have the reconstructor decorator here, I think you could go into the indexd resource classes (gf and study file) and remove the explicit calls to merge_indexd in get and patch bc otherwise you're calling it it twice/making two calls out to index.

dataservice/api/common/model.py
```diff
@@ -41,6 +42,33 @@ def kf_id(cls):
     uuid = db.Column(UUID(), unique=True, default=uuid_generator)
 
 
+class IndexdField():
```
znatty22, line 45: Do you need this because previously when you updated indexd fields, the update event was not actually being fired?

parimalak (approved):
looks good 👍 see comments for minor changes

dataservice/extensions/flask_indexd.py
```diff
@@ -161,17 +162,55 @@ def update(self, record):
...
+
         self.check_response(resp)
         resp.raise_for_status()
 
-        did = resp.json()['did']
-        record.latest_did = did
-
         return record
 
+    def _update_all_acls(self, record):
```
znatty22, line 189: I can't think of a good case for when something might go wrong with one of the PUTs, but if any one of the PUTs came back with a non 200 status code should we check for that and abort the update? So that we can avoid mismatching acls across versions

fiendish:
why is hashes.md5 being given differently formatted values in different places? str(uuid.uuid4()) has some very un-md5-like hyphens in it.

dataservice/api/common/model.py
dankolbman: Yes. The ORM only fires the event when an `UPDATE` statement is made. If none of the mapped fields are updated, it won't fire.

dataservice/api/common/model.py
dankolbman: good catch

dataservice/extensions/flask_indexd.py
dankolbman: You'd still have mismatching acls if you updated doc A and doc B, but failed on doc C.
If you fail on doc C, you'd have to re-put doc A and doc B, and what if you fail when updating doc B etc etc ad absurdum.
This will need to be a check in our reports or indexd cleanup scripts.

dataservice/api/common/model.py
znatty22: 👍

dataservice/extensions/flask_indexd.py
znatty22: yea true..

dataservice/api/genomic_file/resources.py
```diff
@@ -108,12 +108,6 @@ def get(self, kf_id):
             abort(404, 'could not find {} `{}`'
                   .format('genomic_file', kf_id))
 
-        # Merge will return None if the document wasnt found in indexd
```
znatty22, line 111: Need the same in study file resource

dataservice/api/common/model.py
znatty22: Ooh also probably want to reexamine indexd_pagination. You're explicitly calling merge_indexd there too.  index_pagination might even need to be modified..

dankolbman:
I think that all hashes are string now. They should have never been `UUID` types.

dataservice/api/common/model.py
dankolbman: Seems like it doesn't call the re-constructor during the pagination. Tried removing it, but no calls go out to indexd during my tests.

fiendish:
uuid.uuid4().hex will give a string without hyphens that looks more like a real md5 value

znatty22 (approved):
Looks good! 💯 🥇

PR #373, opened 2018-06-15
🐛 Fix 500 error on /studies?study_id

### LINKED ISSUE #368: Investigate why invalid filter paramaters return 500
The endpoint `GET /studies?study_id=SD_XXXXXXXX` returns a `500` response.
The preferred request for this data would be `GET /studies/SD_XXXXXXXX`, however, adding the filter should not cause any unexpected behavior, returning the same response as the preferred endpoint.

znatty22:
Resolves #368

### DIFF
--- a/dataservice/api/study/resources.py
+++ b/dataservice/api/study/resources.py
@@ -33,6 +33,8 @@ class StudyListAPI(CRUDView):
             resource:
               Study
         """
+        filter_params.pop('study_id', None)
+
         q = (Study.query
              .filter_by(**filter_params))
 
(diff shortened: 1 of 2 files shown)


dataservice/api/study/resources.py
```diff
@@ -33,6 +33,8 @@ def get(self, filter_params, after, limit):
             resource:
               Study
         """
+        filter_params.pop('study_id', None)
```
dankolbman, line 36: maybe we want to wrap this into `CRUDView` somehow?

dankolbman (approved):

parimalak (approved):

PR #382, opened 2018-06-20
♻️ Refactor Family Relationships

### LINKED ISSUE #336: Family relationship link is misleading - Implement get all family relationships for a participant
The family relationships link in the participant's _links is formatted like this: 
`family_relationhips: /family-relationships?participant_id=PT_9KYB7NF2`
If PT_9KYB7NF2 is in the `participant_id` column in the family relationship table, then this will return results. If PT_9KYB7NF2 is in the other column, `relative_id`, then this will return 0 results, causing the user to think 0 relationships exist for this participant. 

We need to solve this in 2 steps: 

1. Modify the family relationship model to be less confusing. Rename columns to participant_1 and participant_2. People are confused on what participant and relative are and think that participant refers to proband. Eliminate this by making the relationship columns very generic. 

2. Implement a special endpoint that returns ALL family relationships for a given participant's KF ID (`/family-relationships?participant_id=<participant kf id>` ) or include the participant's relationships in the /families/<family KF ID> endpoint.  Could join the family with family relationships (through participant) and return the family with its relationships embedded in the response.

znatty22:
Resolves #336 

- [x] Add special query with a custom filter parameter, participant_id which is a participant's KF ID. A request with this query parameter will return all family relationships for the participant and their family members IF that participant has a family. If the participant does not have a family, then just the immediate/direct family relationships for the participant will be returned.
    
    - /family-relationships?participant_id=\<kf id\>

- [x] Rename family relationship participant columns:
    - participant_id -> participant1_id
    - relative_id -> participant2_id
- [x] Update migrations
- [x] Update ERD
- [x] Test query all family relationships, model and resource

cc @grant-guo and Rosi can you take a look and let me know if you have questions. Slack is down right now otherwise I'd message you.


dataservice/api/family_relationship/models.py
```diff
@@ -35,62 +35,101 @@ class FamilyRelationship(db.Model, Base):
...
-        primaryjoin=relative_id == Participant.kf_id,
+        primaryjoin=participant2_id == Participant.kf_id,
         backref=db.backref('incoming_family_relationships',
                            cascade='all, delete-orphan'))
 
     @classmethod
-    def query_all_relationships(cls, participant_kf_id):
-        """
-        Find all relationship records given a the kf_id of a particpant
+    def query_all_relationships(cls, participant_kf_id=None,
```
dankolbman, line 74: It would be nice if we could figure out how to turn this into an extension of Query only for FamilyRelationships.
Then we could do:
```
FamilyRelationships.query.filter_by(foo=bar).all_relationships(participant_id='blah').count()
```

dataservice/api/family_relationship/models.py
```diff
@@ -35,62 +35,101 @@ class FamilyRelationship(db.Model, Base):
...
+        if model_filter_params is None:
+            model_filter_params = {}
+        q = FamilyRelationship.query.filter_by(**model_filter_params)
+
+        # Get family relationships and join with participants
+        q = q.join(Participant, or_(FamilyRelationship.participant1,
+                                    FamilyRelationship.participant2))
+
+        # Do this bc query.get() errors out if passed None
+        if participant_kf_id:
```
dankolbman, line 99: If there is no `participant_kf_id`, does it return all family relationships?

dataservice/api/family_relationship/models.py
```diff
@@ -35,62 +35,101 @@ class FamilyRelationship(db.Model, Base):
...
         db.ForeignKey('participant.kf_id'),
         nullable=False,
         doc='kf_id of the other participant in the relationship')
 
-    participant_to_relative_relation = db.Column(db.Text(), nullable=False)
+    participant1_to_participant2_relation = db.Column(db.Text(),
+                                                      nullable=False)
 
-    relative_to_participant_relation = db.Column(db.Text())
+    participant2_to_participant1_relation = db.Column(db.Text())
```
dankolbman, line 59: Is this nullable because it's assumed to be populated by the other relationship? Maybe it should still be explicitly non-nullable anyway.

dataservice/api/family_relationship/models.py
```diff
@@ -35,62 +35,101 @@ class FamilyRelationship(db.Model, Base):
...
-                                         self.participant_to_relative_relation,
-                                         self.relative)
+        return '<{} is {} of {}>'.format(
+            self.participant1.kf_id,
+            self.participant1_to_participant2_relation,
+            self.participant2.kf_id)
 
 
-@event.listens_for(FamilyRelationship.participant_to_relative_relation, 'set')
+@event.listens_for(FamilyRelationship.participant1_to_participant2_relation,
```
dankolbman: Perhaps generalize this for both `participant1_to_participant2_relation` and `participant2_to_participant1_relation`?

migrations/versions/ce359a457d3d_.py
```diff
@@ -0,0 +1,43 @@
+"""
```
dankolbman, line 1: Did you try this on a db with data? Want to make sure that the existing constraints don't break with the name changes.

dataservice/api/family_relationship/models.py
znatty22: Yea thats prob a good idea

dataservice/api/family_relationship/models.py
znatty22: Yea prob a good idea

migrations/versions/ce359a457d3d_.py
znatty22: yep, things seem to work, nothing broke and didn't lose any data

dataservice/api/family_relationship/models.py
znatty22: Yes

dataservice/api/family_relationship/models.py
znatty22: Think I'm going to leave this out for now since the migration could break things. First we can get this PR deployed as is and then check the db in all envs to ensure that there are no nulls in `participant2_to_participant1_relation`, if there are any, populate w a placeholder. Then we can deploy a change to make this column non-nullable

dataservice/api/family_relationship/models.py
```diff
@@ -123,13 +123,37 @@ def __repr__(self):
...
+
+    # Participant 2 to 1 relation
+    rel = target.participant2_to_participant1_relation
+    rev = REVERSE_RELS.get(rel.lower() if rel else None)
+    if rev:
+        target.participant1_to_participant2_relation = rev
+
+
+@event.listens_for(FamilyRelationship, 'before_insert')
+def family_relationship_on_insert(mapper, connection, target):
```
dankolbman: can you use two decorators here instead of two functions?

dankolbman (approved):
LGTM

parimalak (approved):
Update the erd before you merge.

PR #385, opened 2018-06-21
🐛 Fix jenkins

dankolbman:


dankolbman:
Turns out it was in issue with not using the right config during tests.

PR #392, opened 2018-07-09
🐛 Use correct capitalization for more than one race

parimalak:


fiendish (approved):
Looks great. 

Can you also add it to https://github.com/kids-first/kf-api-dataservice/blob/51a821c9971b3e229e0b9555b8669fd9bde87acb/dataservice/util/data_gen/data_generator.py#L57

😃

parimalak:
Data generator script got obsolete. We are no longer updating all changes  because of complexity of relations and use cases.

fiendish:
Gotcha. Fine by me.

znatty22 (approved):
👍

dankolbman:
What is the plan for retroactively migrating these values?

parimalak:
Right now ETL taking care of the updates as we are persisting kf_id/s study wise. if multiple studies needs to be updated; we might have to think through how often we are going to acknowledge these changes. Based on that we can run study specific etl or a separate script/s to update values.

PR #394, opened 2018-07-16
✨ Add biospecimen_genomic_file association table and endpoints

parimalak:
- [x] Add biospecimen_genomic_file assocaition table and endpoints
- [x] Add and update tests for biospecimen_genomic_file
- [x] Update erd
- [x] Update migrations


dataservice/api/genomic_file/schemas.py
```diff
@@ -26,7 +26,9 @@ class Meta(BaseSchema.Meta, IndexdFileSchema.Meta):
 
         exclude = (BaseSchema.Meta.exclude +
                    ('biospecimen', 'sequencing_experiment',) +
-                   ('cavatica_task_genomic_files', 'read_group',))
+                   ('cavatica_task_genomic_files',
+                    'biospecimen__genomic_files',
```
dankolbman: Double underscore

migrations/versions/088f00b10226_.py
```diff
@@ -0,0 +1,53 @@
+"""
+1.2.0- Add biospecimen_genomic_file association table
```
dankolbman: Should have space before the `-`

tests/genomic_file/test_genomic_file_resources.py
```diff
@@ -18,7 +18,7 @@
 
 GENOMICFILE_URL = 'api.genomic_files'
 GENOMICFILE_LIST_URL = 'api.genomic_files_list'
-EXPECTED_TOTAL = ENTITY_TOTAL + 102*2
+EXPECTED_TOTAL = ENTITY_TOTAL + 1*2
```
dankolbman: I think we were using >100 gfs so we could run some pagination tests. Could we make it so that more gfs are put in to the test data?

dataservice/api/biospecimen/models.py
```diff
@@ -98,11 +100,14 @@ class Biospecimen(db.Model, Base):
...
-                                    doc='genomic files this biospecimen')
     diagnoses = db.relationship(Diagnosis,
                                 backref=db.backref('biospecimen',
                                                    lazy=True))
+    genomic_files = association_proxy(
+        'biospecimen_genomic_files', 'genomic_file',
+        creator=lambda genomic_file:
+        BiospecimenGenomicFile(genomic_file=genomic_file))
+
+    biospecimen_genomic_files = db.relationship(BiospecimenGenomicFile,
```
dankolbman, line 111: does this really need to be defined? I don't think we'll need to access it directly.

dankolbman:
I think we need to do this in two prs:

1. PR1: Add BiospecimenGenomicFile table, use some many_many suffix on the temporary duplicate attribtues
2. Merge and populate new table using the current biospecimen-genomicfile relations
3. PR2: Drop the biospecimen_id fk on genomicfile and rename the temporary attributes

migrations/versions/088f00b10226_.py
```diff
@@ -0,0 +1,53 @@
...
+                                            'biospecimen.kf_id'], ),
+                    sa.ForeignKeyConstraint(['genomic_file_id'], [
+                                            'genomic_file.kf_id'], ),
+                    sa.PrimaryKeyConstraint('kf_id'),
+                    sa.UniqueConstraint('genomic_file_id', 'biospecimen_id'),
+                    sa.UniqueConstraint('uuid')
+                    )
+    op.drop_constraint('genomic_file_biospecimen_id_fkey',
+                       'genomic_file', type_='foreignkey')
+    op.drop_column('genomic_file', 'biospecimen_id')
```
dankolbman: We can't drop this yet

parimalak:
@dankolbman  agree with you. created  first PR to  add biospecimen-genomic-files endpoints and model #398. once this gets merged we can migrate data and drop the foreign key.

parimalak:
As suggested by @dankolbman , After careful testing decided to do the migrations in alembic instead of doing in two schema migrations.

dataservice/api/biospecimen_genomic_file/resources.py
```diff
@@ -0,0 +1,163 @@
...
+        template:
+          path:
+            new_resource.yml
+          properties:
+            resource:
+              BiospecimenGenomicFile
+        """
+        body = request.get_json(force=True)
+        try:
+            app = (BiospecimenGenomicFileSchema(strict=True)
```
dankolbman: `bs_gf` may be a better name than `app`?

dankolbman (approved):
Tested this with a dump of current qa.
Migration worked and I was able to trace back genomic files to biospecimens through the many-many table.

dataservice/api/biospecimen_genomic_file/resources.py
parimalak: 👍

dataservice/api/biospecimen/models.py
znatty22: I think we do need this defined bc we need the cascade all delete orphan part. We want it so that when you delete a Biospecimen, it will delete the rows in the association table that reference that biospecimen. Usually you don't have to specify the cascade's manually for many-to-many, but since we're using sqlalchemy association object rather than association table to define the association table we have to specify the cascade manually on both the biospecimen and genomic file tables. Alternatively, we might be able to define both the cascades in the association obj class BiospecimenGenomicFile .. but its prob fine the way it is

migrations/versions/088f00b10226_.py
```diff
@@ -0,0 +1,139 @@
...
+
+def data_upgrades():
+    """
+    COPY genomic_file.biospecimen_id TO biospecimen_genomic_file.biospecimen_id
+    COPY genomic_file.kf_id TO biospecimen_genomic_file.genomic_file_id
+    DROP genomic_file.biospecimen_id
+    """
+    connection = op.get_bind()
+
+    biospecimen_genomic_file = sa.Table(
```
znatty22, line 56: These table instantiations look like we're creating the tables again or something..I see a lot of duplicate code here and above in upgrade(). Is there a way to get a reference to `biospecimen_genomic_file` table and `genomic_file` table without defining all of the columns?

parimalak: Here we are creating views with old and new columns which we want to change.  we defined `created_at` and `modified_at` fields for `biospecimen_genomic_file` bc how our paginated results works.

dankolbman: We need to define these manually because using the code defined in the models can be unreliable during an upgrade Where the tables in the database could be any sort of schema.

znatty22 (changes requested):
I don't see any updates in biospecimen/genomic file model tests for the cascading deletes. Probably a good idea to test that when you delete either a biospecimen or genomic_file, the associated rows in biospecimen_genomic_file are also deleted.

tests/genomic_file/test_genomic_file_models.py
```diff
@@ -206,67 +167,51 @@ def test_update_acl_only(self):
...
 
         # Check database
-        biospecimen = Biospecimen.query.filter_by(
-            kf_id=biospecimen_id).one()
+
+        biospecimen = Biospecimen.query.first()
         self.assertEqual(len(biospecimen.genomic_files), 0)
 
-    def test_delete_via_biospecimen(self):
+    def test_cascade_delete_via_biospecimen(self):
```
znatty22, line 191: Almost there, I think you need one more test like this one. Test that deleting a genomic file deletes the associated rows in BiospecimenGenomicFile.

znatty22 (approved):
👍

PR #398, opened 2018-07-17
WIP✨ Add biospecimen-genomic-files endpoint

parimalak:
This is the first PR that includes `biospecimen_genomic_file` association table and endpoints.
migrations for dropping `biospecimen_id`  from genomic_file and updating its dependent entities will be done in another PR.


dankolbman:
I'm wondering if maybe it would be better to handle this migration directly inside of the alembic migration instead of doing two different schema releases.
Can we try this?
See [this](https://stackoverflow.com/questions/43153346/update-column-content-during-alembic-migration) for an example of how the columns can be modified within the migration

PR #401, opened 2018-07-26
✨ Add biospecimen_diagnosis association table

parimalak:
- [x] Add biospecimen_diagnosis table and endpoints
- [x]  Add and update tests
- [x] Update migrations and erd


junjun-zhang:
The code change looks ok to me.

Just a thought although may be not that relevant at this moment. With more relationships added to the model recently (eg, many-to-many between diagnosis and biospecimen; and many-to-many between biospecimen and genomic files), the model seems becoming more like a `graph db`. Nothing wrong with being like a `graph db` if it's what it takes to model the data. Something to be mindful about is that ETL process will become more complex each time new relationship is introduced and keeping ETL up with source model changes could become more difficult over time.

Another perspective is that we may need to think about alternatives to the `all-inclusive` *_centric Elasticsearch indexes, eg, `file_centric` index has **all** other related entities (diagnosis, specimen, phenotype, family info etc) and their relationships nested under each file document. Such gigantic `all-inclusive` indexes could be hard to comprehend and make use. One alternative could be building more smaller, single-purposed indexes. Each index along will not be able to support use cases where integrated filtering is needed, a solution to that may include multi-step, multi-query, wizard style UI/UX. More to think about for sure.

migrations/versions/ce01ada6ac80_.py
```diff
@@ -0,0 +1,54 @@
...
+                                            'biospecimen.kf_id'], ),
+                    sa.ForeignKeyConstraint(
+                        ['diagnosis_id'], ['diagnosis.kf_id'], ),
+                    sa.PrimaryKeyConstraint('kf_id'),
+                    sa.UniqueConstraint('diagnosis_id', 'biospecimen_id'),
+                    sa.UniqueConstraint('uuid')
+                    )
+    op.drop_constraint('diagnosis_biospecimen_id_fkey',
+                       'diagnosis', type_='foreignkey')
+    op.drop_column('diagnosis', 'biospecimen_id')
```
dankolbman, line 43: We have to do a data migration similar to how we did bs-gf migration.

dataservice/api/biospecimen_diagnosis/resources.py
```diff
@@ -0,0 +1,164 @@
...
+    BiospecimenDiagnosis
+)
+from dataservice.api.biospecimen_diagnosis.schemas import (
+    BiospecimenDiagnosisSchema
+)
+from dataservice.api.common.views import CRUDView
+from dataservice.api.common.schemas import filter_schema_factory
+
+
+class BiospecimenDiagnosisListAPI(CRUDView):
```
dankolbman: We really need to move away from having the user manage these associations via individual requests.
Would it be too much to try that here?

dankolbman (changes requested):

migrations/versions/ce01ada6ac80_.py
parimalak: we don't have associated data between biospecimen and diagnosis. do we still need to data migrations?

dataservice/api/biospecimen/models.py
```diff
@@ -111,3 +131,114 @@ class Biospecimen(db.Model, Base):
...
+    """
+    Ensure that both the diagnosis and biospecimen
+    have the same participant
+    If this is not the case then raise DatabaseValidationError
+    """
+    from dataservice.api.errors import DatabaseValidationError
+    # Return if biospecimen is None
+    if not target:
+        return
+    print(target.diagnoses)
```
dankolbman: 🔥

dataservice/api/biospecimen/models.py
```diff
@@ -111,3 +131,114 @@ class Biospecimen(db.Model, Base):
...
+    # Get diagnosis by id
+    ds = None
+    if target.diagnoses:
```
dankolbman: I think these two lines can be removed and line 186 can become `if not target or not target.diagnoses:`

dataservice/api/biospecimen/models.py
```diff
@@ -111,3 +131,114 @@ class Biospecimen(db.Model, Base):
...
+        ds = Diagnosis.query.get(target.diagnoses[0].kf_id)
```
dankolbman: is this not equivalent to `target.diagnoses[0]`?

dataservice/api/biospecimen/models.py
```diff
@@ -111,3 +131,114 @@ class Biospecimen(db.Model, Base):
...
+
+    # If biospecimen and diagnosis doesn't exist, return and
+    # let ORM handle non-existent foreign key
+    if ds is None:
```
dankolbman: This could be removed if the above changes are made.

dataservice/api/diagnosis/models.py
```diff
@@ -69,48 +68,38 @@ class Diagnosis(db.Model, Base):
...
     if not target:
         return
-
     # Get biospecimen by id
     bsp = None
-    if target.biospecimen_id:
-        bsp = Biospecimen.query.get(target.biospecimen_id)
-
-    # If biospecimen doesn't exist, return and
+    if target.biospecimens:
```
dankolbman: This logic can also be reduced into line 82.

migrations/versions/ce01ada6ac80_.py
dankolbman: Hmm, as in we haven't added any yet? I still feel that we should add it, just in case. It should be almost identical to the last migration.

tests/test_api.py
```diff
@@ -81,7 +81,7 @@ def test_status_format(self, client, endpoint, method):
         ('/family-relationships', ['participant1', 'participant2']),
         ('/phenotypes', ['participant']),
         ('/outcomes', ['participant']),
-        ('/diagnoses', ['participant', 'biospecimen']),
+        ('/diagnoses', ['participant']),
```
dankolbman, line 84: shouldn't the link to biospecimen also exist?

tests/test_api.py
```diff
@@ -126,10 +126,11 @@ def test_parent_links(self, client, entities, endpoint, parents):
         ('/sequencing-centers', ['sequencing_experiments', 'biospecimens']),
         ('/participants', ['diagnoses', 'phenotypes', 'outcomes',
                            'biospecimens']),
-        ('/biospecimens', ['diagnoses', 'biospecimen_genomic_files']),
+        ('/biospecimens', ['biospecimen_genomic_files']),
```
dankolbman, line 129: Shouldn't the link to diagnoses exist, but not the biospecimen_genomic_files?

dankolbman:
I'm seeing that the `diagnoses` field is being included in the `result` body, that should probably be emitted from the schema.

dankolbman:
I've pulled and tested this manually and everything seems to work nicely 🎉 It would be nice if there were a unit test for using the api to add in the relation between diagnosis and biospecimen, though.

dataservice/api/biospecimen/models.py
parimalak: 👍

tests/test_api.py
parimalak: ya it should be `biospecimens`

tests/test_api.py
parimalak: ah, `diagnoses` link should exist and also `biospecimen_genomic_files`. i think you meant `biospecimen_diagnoses`?

dataservice/api/biospecimen/models.py
```diff
@@ -111,3 +131,108 @@ class Biospecimen(db.Model, Base):
...
+             'refers to participant {} and '
+             'biospecimen {} refers to participant {}')
+            .format(ds.kf_id,
+                    ds.participant_id,
+                    bsp.kf_id,
+                    bsp.participant_id))
+        raise DatabaseValidationError(target_entity, operation, message)
+
+
+def validate_biospecimen(target):
```
znatty22: This looks like its just validating the first diagnosis in Biospecimen.diagnoses? 

I think any time a diagnosis is appended to Biospecimen.diagnoses, you want to do the validation. See https://docs.sqlalchemy.org/en/latest/orm/events.html

dataservice/api/biospecimen/models.py
```diff
@@ -111,3 +131,108 @@ class Biospecimen(db.Model, Base):
...
+@event.listens_for(BiospecimenDiagnosis, 'before_update')
+def biospecimen_diagnosis_on_update(mapper, connection, target):
+    """
+    Run preprocessing/validation of diagnosis before update
+    """
+    validate_diagnosis_biospecimen(target)
+
+
+@event.listens_for(Biospecimen, 'before_insert')
+def biospecimen_on_insert(mapper, connection, target):
```
znatty22: I think this needs to change to the append event listener, see comment above

dataservice/api/biospecimen/models.py
```diff
@@ -111,3 +131,108 @@ class Biospecimen(db.Model, Base):
...
+    """
+    Run preprocessing/validation of diagnosis before insert
+    """
+    validate_biospecimen(target)
+
+
+@event.listens_for(Biospecimen, 'before_update')
+def biospecimen_on_update(mapper, connection, target):
```
znatty22: Once you change to append event listener, you can remove this

dataservice/api/diagnosis/models.py
```diff
@@ -69,48 +68,34 @@ class Diagnosis(db.Model, Base):
...
                                nullable=False)
-    biospecimen_id = db.Column(KfId(),
-                               db.ForeignKey('biospecimen.kf_id'),
-                               doc='the biospecimen which recieved the '
-                               'diagnosis',
-                               nullable=True)
 
 
-def validate_diagnosis_biospecimen(target):
+def validate_diagnosis(target):
```
znatty22, line 73: Similar to comment on validate_biospecimen in biospecimen/model.py

dataservice/api/biospecimen/models.py
```diff
@@ -111,3 +131,108 @@ class Biospecimen(db.Model, Base):
     biospecimen_genomic_files = db.relationship(BiospecimenGenomicFile,
                                                 backref='biospecimen',
                                                 cascade='all, delete-orphan')
+    diagnoses = db.relationship('Diagnosis', secondary='biospecimen_diagnosis',
```
znatty22, line 134: Hmm I'm not sure I understand how this works or what it's doing. It kind of looks like an association proxy to Diagnosis? 
Do we need something like:
```
biospecimen_diagnoses = db.relationship('BiospecimenDiagnosis',
                                                  backref='Biospecimen',
                                                  cascade='all, delete-orphan')
```

dataservice/api/diagnosis/models.py
```diff
@@ -69,48 +68,34 @@ class Diagnosis(db.Model, Base):
                                db.ForeignKey('participant.kf_id'),
                                doc='the participant who was diagnosed',
                                nullable=False)
-    biospecimen_id = db.Column(KfId(),
```
znatty22: Why don't we have a relationship to the biospecimen_diagnosis here? 
```
biospecimen_diagnoses = db.relationship('BiospecimenDiagnosis',
                                                  backref='Biospecimen',
                                                  cascade='all, delete-orphan')
```

znatty22 (commented):
Maybe I missed something in discussions, but I don't see a way to create BiospecimenDiagnosis records through the API? How will we link these two things during ingest?

dataservice/api/biospecimen/models.py
dankolbman: Can this validation function be removed entirely?
I believe the check for the same participant is a constraint solely on the relationship between the two entities, not the biospecimen.

parimalak: I think we could remove biospecimen_validation also diagnosis_validation but I need to test few things before we do that.

dataservice/api/diagnosis/models.py
```diff
@@ -69,65 +68,55 @@ class Diagnosis(db.Model, Base):
...
+    from dataservice.api.biospecimen.models import Biospecimen
+    # Return if biospecimen is None or
+    # if diagnosis doesn't exist, return and
     # let ORM handle non-existent foreign key
-    if bsp is None:
+    if not target or not target.biospecimens:
         return
-
+    # Get diagnosis by id and bisopecimen by id
+    bsp = Biospecimen.query.get(target.biospecimens[0].kf_id)
```
dankolbman: do you need to check all biosecimens?

parimalak: I don't think so. marshmallow seems to be taking care of all objects in the nested request.

dankolbman: How does that happen?
Won't `target.biospecimens` be a list of all biospecimens for that diagnosis?

parimalak: ya but the validation is called for all the biospecimens. I added tests with multiple biospecimens (both correct) and(1 incorrect and 1 correct). validation worked exactly how we wanted.

dankolbman (approved):

dataservice/api/biospecimen/models.py
```diff
@@ -111,3 +131,57 @@ class Biospecimen(db.Model, Base):
...
+                                backref=db.backref(
+                                    'biospecimens'))
+
+
+def validate_biospecimen(target):
```
znatty22, line 139: This still seems to only validate the first diagnosis in the list of target.diagnoses. I think you either need to call the validation on every item in the list or call it any time a diagnosis is appended to a biospecimen's list of diagnoses.

dataservice/api/diagnosis/models.py
znatty22: Again same comment as before, I think you either need to call the validation on every item in the list or call it any time a biospecimen is appended to a diagnosis's list of biospecimens.

dataservice/api/diagnosis/models.py
znatty22: I don't think this actually works. Code below doesn't throw any exceptions but it should: 

```python
    def test(self):
        s = Study(external_id="study")
        sc = SequencingCenter(name='Baylor')

        # Participant 0
        p0 = Participant(external_id='p0', study=s)
        b0 = Biospecimen(analyte_type='DNA', participant=p0,
                         sequencing_center=sc)
        d0 = Diagnosis(external_id='d0', participant=p0)

        p1 = Participant(external_id='p1', study=s)
        b1 = Biospecimen(analyte_type='RNA', participant=p1,
                         sequencing_center=sc)
        db.session.add(s)
        db.session.commit()

        # Participant 0 - Link their biosp b0 and diagnosis d0
        bd0 = BiospecimenDiagnosis(biospecimen_id=b0.kf_id,
                                   diagnosis_id=d0.kf_id)
        db.session.add(bd0)
        db.session.commit()

        # Participant 1 - Link their biop b1 to Participant 0 diagnosis d0
        bd1 = BiospecimenDiagnosis(biospecimen_id=b1.kf_id,
                                   diagnosis_id=d0.kf_id)
        db.session.add(bd1)
        db.session.commit()

```

It's because only the first item in the list (whether biospecimens or diagnoses) get's validated.

parimalak: The validations are called individually even if I pass multiple objects. this test is failing because we removed validation on BiospecimenDiagnosis and the requests are not validated at all.

tests/biospecimen/test_biospecimen_models.py
```diff
@@ -281,6 +284,53 @@ def test_one_to_many_relationship_delete(self):
...
+        p1 = Participant(external_id='p1', study=s)
+        b1 = Biospecimen(analyte_type='RNA', participant=p1,
+                         sequencing_center_id=sc.kf_id)
+        db.session.add(s)
+        db.session.commit()
+        # Participant 1 - Link their biop b1 to Participant 0 diagnosis d0
+        bd1 = BiospecimenDiagnosis(biospecimen_id=b1.kf_id,
+                                   diagnosis_id=dg.kf_id)
+        db.session.add(bd1)
+        self.assertRaises(DatabaseValidationError, db.session.commit)
```
znatty22: Change to: 
```python
with self.assertRaises(DatabaseValidationError):
            db.session.commit()
```

tests/diagnosis/test_diagnosis_models.py
```diff
@@ -131,7 +117,7 @@ def test_add_invalid_biospecimen(self):
 
         # Try linking through diagnosis
         d = Diagnosis.query.first()
-        d.biospecimen_id = b.kf_id
+        d.biospecimens = [b]
         self.assertRaises(DatabaseValidationError, db.session.commit)
```
znatty22: Change to: 
```python
with self.assertRaises(DatabaseValidationError):
            db.session.commit()
```

tests/diagnosis/test_diagnosis_models.py
```diff
@@ -140,11 +126,60 @@ def test_add_invalid_biospecimen(self):
...
```
znatty22: Change to: 
```python
with self.assertRaises(DatabaseValidationError):
            db.session.commit()
```

tests/diagnosis/test_diagnosis_models.py
```diff
@@ -140,11 +126,60 @@ def test_add_invalid_biospecimen(self):
...
+        # Create diagnosis
+        data = {
+            'external_id': 'd1',
+            # non-existent required param: participant_id
+        }
+        d = Diagnosis(**data)
+
+        # Add to db
+        db.session.add(d)
+        self.assertRaises(IntegrityError, db.session.commit)
```
znatty22: Change to: 
```python
            db.session.commit()
```

tests/diagnosis/test_diagnosis_models.py
```diff
@@ -140,11 +126,60 @@ def test_add_invalid_biospecimen(self):
...
+        # Create diagnosis
+        data = {
+            'external_id': 'd1',
+            'participant_id': ''  # empty blank foreign key
+        }
+        d = Diagnosis(**data)
+
+        # Add to db
+        db.session.add(d)
         self.assertRaises(IntegrityError, db.session.commit)
```
znatty22: Change to: 
```python
            db.session.commit()
```

znatty22 (commented):
See comments in line..change all assertRaises to use with assertRaises

znatty22 (approved):
💯

PR #407, opened 2018-08-07
✨ Add visible to base model

parimalak:


dataservice/api/common/model.py
```diff
@@ -29,6 +29,7 @@ def __init__(self, *args, **kwargs):
 class IDMixin:
     """
     Defines base ID columns common on all Kids First tables
+    :param visible: Flags visibility of data from the dataservice
```
dankolbman: Does it make the most sense to have this in the ID mixin?

dankolbman:
I think we also need to add this field to all the schemas, too, so that etl will know whether to publish or not.

dataservice/api/common/model.py
```diff
@@ -179,4 +180,7 @@ class Base(IDMixin, TimestampMixin):
     """
     Defines base SQlAlchemy model class
     """
-    pass
+    visible = db.Column(db.Boolean(),
+                        nullable=False,
+                        server_default='true',
```
znatty22, line 185: Just curious, are you using `server_default` (vs. default) because you specifically want the boolean to be represented as `true` in PG?

znatty22 (approved):
👍

dataservice/api/common/model.py
parimalak: Moving the doc string to base model

parimalak:
@dankolbman doesn't the `visible` field in the model will default to the schema?

dankolbman (approved):

PR #412, opened 2018-08-15
✨ Refactor ReaGroup GenomicFile Relation

znatty22:
This PR changes the 1 to 1 relationship between ReadGroup and GenomicFile to be many to many. The `paired_end` field, previously on ReadGroup will now be on the GenomicFile.

Also, orphaned read groups (those which do not have any associated genomic files) are automatically deleted when they become orphans


config.py
```diff
@@ -44,7 +44,7 @@ class TestingConfig(Config):
     SERVER_NAME = "localhost"
     TESTING = True
     WTF_CSRF_ENABLED = False
-    # SQLALCHEMY_DATABASE_URI = 'postgres://postgres@localhost:5432/test'
+    SQLALCHEMY_DATABASE_URI = 'postgres://postgres@localhost:5432/test'
```
dankolbman: This is killing Jenkins

dataservice/api/read_group/models.py
```diff
@@ -17,17 +20,47 @@ class ReadGroup(db.Model, Base):
...
+                              nullable=False)
+
+    genomic_file_id = db.Column(KfId(),
+                                db.ForeignKey('genomic_file.kf_id'),
                                 nullable=False)
+    read_group = db.relationship('ReadGroup')
+    genomic_file = db.relationship('GenomicFile')
+
+    def __repr__(self):
+        return "{}-{}".format(self.read_group.external_id,
```
dankolbman: Do we use this format anywhere else?
We should probably stay close to the typical `<Object >` format that python uses.
Maybe `<ReadGroupGenomicFile {self.read_group.__repr__()} - {self.genomic_file.__repr__()}>`?

znatty22: Ah sorry, that was there for debugging, let me change it

dataservice/api/genomic_file/resources.py
```diff
@@ -58,6 +61,11 @@ def get(self, filter_params, after, limit):
                  .join(Biospecimen.participant)
                  .filter(Participant.study_id == study_id))
 
+        from dataservice.api.read_group.models import ReadGroupGenomicFile
+        if read_group_id:
+            q = (q.join(ReadGroupGenomicFile)
+                 .filter(ReadGroupGenomicFile.read_group_id == read_group_id))
```
parimalak, line 67: I think this query still have the same problem what we have with biospecimen-diagnosis

dataservice/api/read_group/models.py
```diff
@@ -17,17 +20,43 @@ class ReadGroup(db.Model, Base):
...
+                                          'genomic_file_id'),)
+    read_group_id = db.Column(KfId(),
+                              db.ForeignKey('read_group.kf_id'),
+                              nullable=False)
+
+    genomic_file_id = db.Column(KfId(),
+                                db.ForeignKey('genomic_file.kf_id'),
                                 nullable=False)
+    read_group = db.relationship('ReadGroup')
+    genomic_file = db.relationship('GenomicFile')
```
parimalak, line 55: are these relationships used in the delete listening event?

dataservice/api/genomic_file/resources.py
znatty22: It doesn't, I tested it to make sure we're not getting back extra genomic files

dataservice/api/read_group/models.py
znatty22: No they're not used, but they're used in tests. Also just FYI, the delete listening event only deletes read groups that have no genomic files associated with them. There is no automatic deletion of genomic files

dankolbman (approved):
👍

parimalak (approved):
update erd before merging.

PR #413, opened 2018-08-17
👷 Load bucket url from vault

dankolbman:
Loads the bucket service url from vault, if the secret path is in the environment. Will fall back to use the url in the environment, if it is availabel.


alubneuski (approved):

parimalak (approved):

znatty22 (commented):
🥇

znatty22 (approved):
Oops forgot to approve haha

PR #416, opened 2018-08-20
🐛 Fix biospecimens and diagnoses links by diagnosis_id and biospecimen_id

parimalak:
issue : /diagnoses?biospecimen_id=BS_id ,  /biospecimen?diagnosis_id=DG_idresulted in internal server error
fix :  Added queries to join the biospecimens/diagnoses with BiospecimenDiagnosis

### DIFF
--- a/dataservice/api/biospecimen/resources.py
+++ b/dataservice/api/biospecimen/resources.py
@@ -38,22 +38,22 @@ class BiospecimenListAPI(CRUDView):
             resource:
               Biospecimen
         """
-        # Get study id and remove from model filter params
+        # Get study id, diagnosis_id and remove from model filter params
         study_id = filter_params.pop('study_id', None)
         diagnosis_id = filter_params.pop('diagnosis_id', None)
         # Apply filter params
         q = (Biospecimen.query
              .filter_by(**filter_params))
 
-        if diagnosis_id:
-            q = q.join(Biospecimen.diagnoses).\
-                filter(BiospecimenDiagnosis.diagnosis_id == diagnosis_id)
-
-        # Apply study filter param
         from dataservice.api.participant.models import Participant
+        # Apply study_id filter and diagnosis_id filter
+        # Apply study filter param
         if study_id:
             q = (q.join(Participant.biospecimens)
                  .filter(Participant.study_id == study_id))
+        if diagnosis_id:
+            q = q.join(Biospecimen.diagnoses).\
+                filter(BiospecimenDiagnosis.diagnosis_id == diagnosis_id)
 
         return (BiospecimenSchema(many=True)
                 .jsonify(Pagination(q, after, limit)))
--- a/dataservice/api/biospecimen/resources.py
+++ b/dataservice/api/biospecimen/resources.py
@@ -5,7 +5,8 @@ from webargs.flaskparser import use_args
 
 from dataservice.extensions import db
 from dataservice.api.common.pagination import paginated, Pagination
-from dataservice.api.biospecimen.models import Biospecimen
+from dataservice.api.biospecimen.models import (
+    Biospecimen, BiospecimenDiagnosis)
 from dataservice.api.biospecimen.schemas import (
     BiospecimenSchema,
     BiospecimenFilterSchema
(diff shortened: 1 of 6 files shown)


tests/test_pagination.py
```diff
@@ -397,3 +396,27 @@ def _check_link(self, link_str, params):
         assert float(q_params.get('after')[0])
         for k, v in params.items():
             assert q_params.get(k)[0] == v
+
+    @pytest.mark.parametrize('endpoint,expected_total', [
+        ('/biospecimens', 1),
+        ('/diagnoses', 1),
+    ])
+    def test_all_filter(self, client, participants,
```
dankolbman: This test would be better fit for the test_resources file for the two types.

dataservice/api/diagnosis/resources.py
```diff
@@ -38,16 +38,26 @@ def get(self, filter_params, after, limit):
...
-        if study_id:
+        if biospecimen_id:
+            q = q.join(Diagnosis.biospecimens).\
+                filter(BiospecimenDiagnosis.biospecimen_id == biospecimen_id)
+
+        # Apply study_id filter and biospecimen_id filter
+        if biospecimen_id and study_id:
+            q = q.join(Diagnosis.biospecimens).\
+                filter(BiospecimenDiagnosis.biospecimen_id == biospecimen_id).\
+                filter(Participant.study_id == study_id)
```
dankolbman: Still not sure I understand how this works.

dataservice/api/diagnosis/resources.py
```diff
@@ -38,16 +38,26 @@ def get(self, filter_params, after, limit):
...
-        # Apply study_id filter
-        from dataservice.api.participant.models import Participant
-        if study_id:
+        if biospecimen_id:
+            q = q.join(Diagnosis.biospecimens).\
+                filter(BiospecimenDiagnosis.biospecimen_id == biospecimen_id)
+
+        # Apply study_id filter and biospecimen_id filter
+        if biospecimen_id and study_id:
+            q = q.join(Diagnosis.biospecimens).\
```
dankolbman: I get this warning. You may not have to join here:
```
SAWarning: Pathed join target Diagnosis.biospecimens has already been joined to; skipping
```

dataservice/api/diagnosis/resources.py
```diff
@@ -38,16 +38,23 @@ def get(self, filter_params, after, limit):
...
+        if biospecimen_id and study_id:
+            q = q.join(Participant.diagnoses).\
+                join(Diagnosis.biospecimens).\
+                filter(BiospecimenDiagnosis.biospecimen_id == biospecimen_id).\
+                filter(Participant.study_id == study_id)
+        elif study_id and not biospecimen_id:
             q = (q.join(Participant.diagnoses)
                  .filter(Participant.study_id == study_id))
-
+        elif biospecimen_id and not study_id:
```
dankolbman: I think the first if could be removed and then just do
```
if study_id:
    ...
if biospecimen_id:
    ...
```

dataservice/api/biospecimen/resources.py
```diff
@@ -37,18 +38,27 @@ def get(self, filter_params, after, limit):
...
+        diagnosis_id = filter_params.pop('diagnosis_id', None)
         # Apply filter params
         q = (Biospecimen.query
              .filter_by(**filter_params))
 
-        # Apply study filter param
         from dataservice.api.participant.models import Participant
-        if study_id:
+        # Apply study_id filter and diagnosis_id filter
+        if diagnosis_id and study_id:
```
dankolbman: You should be able to remove this and change the two `elif`s below to `if`s

parimalak: all the three conditions have different joins.  if we remove the `elifs`  two of the conditions will be true at the same time and the query throws multiple joins on the same table.

dankolbman: I think they are doing the same joins though?

parimalak: ahh, I see what you are saying if both conditions are satisfied both the conditions queries will be executed.

dankolbman (approved):
looks good!

znatty22 (approved):
👍

PR #420, opened 2018-08-22
🐛 fix parsing of uniqueness constraint error

fiendish:
this works on both single and multiple column constraints

### DIFF
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -1,6 +1,15 @@
+import copy
 import json
+import re
+
 import pytest
 
+from tests.conftest import (ENDPOINT_ENTITY_MAP, ENTITY_PARAMS,
+                            _add_foreign_keys)
+
+CONSTRAINT_ERR_RE = re.compile(
+    'could not create (\w+)\. one with \(.+\) = \(.+\) already exists\.'
+)
 
 class TestErrors:
     """ Test general error handling """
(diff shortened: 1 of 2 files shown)


fiendish:
fixes #421

tests/test_errors.py
```diff
@@ -31,3 +36,51 @@ def test_fk_not_exists(self, client, endpoint, kf_id):
...
+
+    def test_uniqueness_constraints(self, client, entities):
+        """ Test integrity error from uniqueness violations """
+        # apparently some of our operations are destructive to ENTITY_PARAMS
+        # state used by other tests
+        global ENTITY_PARAMS
+        ENTITY_PARAMS = copy.deepcopy(ENTITY_PARAMS)
+
+        # init with dummy data
+        for endpoint, inputs in ENTITY_PARAMS['fields'].items():
```
dankolbman: Is this duplicating what's in the `entities` fixture?
Would it be better to look up one of the relationships and try to add it again?

fiendish: It does try to add again below to test the constraint messages. Maybe I've misunderstood the question or the implication, but before this point I think the test's dataservice is empty. This just makes sure that isn't empty anymore in the easiest way possible, otherwise you have to deal with whatever all the foreign key requirements are to get the first entities in at all and I didn't think that was in scope for the testing here.  (it deepcopies ENTITY_PARAMS because otherwise _add_foreign_keys breaks subsequent tests for some reason. I'm not sure why that would be the case. I expected that each test would run in a freshly constructed environment, but 🤷‍♂️ )

tests/test_errors.py
```diff
@@ -31,3 +36,51 @@ def test_fk_not_exists(self, client, endpoint, kf_id):
         else:
             message = 'Longer than maximum length 11'
             assert message in response['_status']['message']
+
+    def test_uniqueness_constraints(self, client, entities):
```
znatty22: Could you add a test for family relationships? I did a quick test and saw that the table in the error message wasn't quite parsed correctly: 
```
{
  "_status": {
    "code": 400, 
    "message": "could not create .  with (participant1_id, participant2_id, participant1_to_participant2_relation, participant2_to_participant1_relation) = (PT_3X5795SV, PT_0B265MYJ, mother, Child) already exists."
  }
}
``` 
If there was a way to make this a bit more generic, where you don't have to have an explicit test for each desired table, that would be great. Then you could just have this test take in a pytest parameterized list containing tables (sequencing_center, biospecimen_genomic_file, biospecimen_diagnosis, read_group_genomic_file, family_relationship, etc), similar to the other tests we have (see the test above this one - test_fk_not_exists)

tests/test_errors.py
znatty22: Hmm, I concur with Dan's comment. You should be able to use the entities fixture and add the same entity again. I'll try to look at this tomorrow

znatty22:
Thanks for fixing this @fiendish!

tests/test_errors.py
dankolbman: Perhaps it would be better to not use the entities fixture and instead load up the minimal data manually? This would be a lot more clear in the test as well as avoid having to set up all the data in the api.

fiendish: Let's talk about this tomorrow. I need to learn more about this entities fixture of which you speak.

tests/test_errors.py
fiendish: > If there was a way to make this a bit more generic, where you don't have to have an explicit test for each desired table, that would be great.

I agree, though it wasn't immediately clear at the time how to make it happen in a way that didn't violate any principles of test validity. I'll see if I can come up with something.

dankolbman (approved):
lgtm 🎉

znatty22 (approved):
👍  Awesome, thanks for condensing the test!

PR #422, opened 2018-08-24
✨Rename field volume_ml to volume_ul

parimalak:


znatty22 (approved):
💯

dankolbman (changes requested):
Needs a migration.

migrations/versions/cdbd609e2bd5_.py
```diff
@@ -0,0 +1,37 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('biospecimen', sa.Column(
+        'volume_ul', sa.Float(), nullable=True))
+    op.drop_column('biospecimen', 'volume_ml')
+    op.add_column('biospecimen_diagnosis', sa.Column(
```
znatty22: This looks like it maybe shouldn't be here? I thought you made that change to the model in another PR we already merged

dankolbman: why is this here?

migrations/versions/cdbd609e2bd5_.py
```diff
@@ -0,0 +1,37 @@
...
+
+Revision ID: cdbd609e2bd5
+Revises: ce01ada6ac80
+Create Date: 2018-08-27 10:28:59.687358
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+import dataservice
```
znatty22, line 12: Doesn't seem to be used anywhere

migrations/versions/cdbd609e2bd5_.py
```diff
@@ -0,0 +1,37 @@
...
+down_revision = 'ce01ada6ac80'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('biospecimen', sa.Column(
+        'volume_ul', sa.Float(), nullable=True))
+    op.drop_column('biospecimen', 'volume_ml')
```
dankolbman: We want to rename, not drop, or else we will lose all the data.

dankolbman (changes requested):
I think you want to use the rename column function instead of dropping and adding a new column.

migrations/versions/cdbd609e2bd5_.py
parimalak: good catch, apparently alembic doesn't automatically recognize the column name changes will change this manually.

migrations/versions/cdbd609e2bd5_.py
parimalak: I think this was missed in previous migration

migrations/versions/cdbd609e2bd5_.py
parimalak: use this import while upgrading database

znatty22 (approved):
👍

dankolbman (approved):
lgtm

baileyckelly:
@rosibaj This change is in QA now (Biospecimen -> volume_ul)

PR #436, opened 2018-08-31
✨update outcome enum to deceased

parimalak:


dankolbman:
Does there need to be a data migration for this?

dankolbman:
Looks like there's 977 values that will need to be updated.

parimalak:
ya, I thought we can patch it without doing the migration because it is just data change without any model change.

dankolbman (approved):

znatty22 (approved):
👍 Looks good

PR #437, opened 2018-09-06
🐛 Fix join for genomic files study filter

dankolbman:
Groups by genomicfile after filtering to eliminate duplication from many-many joins with biospecimen.


znatty22 (changes requested):
Looks good could you just change the title to match the code change

znatty22 (approved):
Change title of PR to match code change

parimalak (approved):
👍

PR #443, opened 2018-09-18
♻️ Don't use rev in indexd requests

dankolbman:
New indexd version enforces a schema that does not allow the rev field in the body of the request.

### DIFF
--- a/dataservice/extensions/flask_indexd.py
+++ b/dataservice/extensions/flask_indexd.py
@@ -205,6 +205,8 @@ class Indexd(object):
                 if doc['version'] is None:
                     del doc['version']
                 url = '{}{}?rev={}'.format(self.url, did, doc['rev'])
+                # rev is not allowed in put schema
+                del doc['rev']
                 resp = self.session.put(url, json=doc)
                 # Update the record's rev if it's the record being modified
                 if record.latest_did == did:
--- a/tests/genomic_file/test_genomic_file_resources.py
+++ b/tests/genomic_file/test_genomic_file_resources.py
@@ -247,6 +247,27 @@ def test_get_one(client, entities):
     assert resp['file_format'] == gf.file_format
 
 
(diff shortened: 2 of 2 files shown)


dataservice/extensions/flask_indexd.py
```diff
@@ -205,6 +205,8 @@ def _update_all_acls(self, record):
                 if doc['version'] is None:
                     del doc['version']
                 url = '{}{}?rev={}'.format(self.url, did, doc['rev'])
+                # rev is not allowed in put schema
+                del doc['rev']
```
znatty22, line 209: Instead of doing this, could you remove `rev` from `fields` list above?

tests/genomic_file/test_genomic_file_resources.py
```diff
@@ -247,6 +247,27 @@ def test_get_one(client, entities):
...
+    orig_calls = indexd.put.call_count
+
+    genomic_file = resp['results']
+    body = genomic_file
+    response = client.patch(url_for(GENOMICFILE_LIST_URL)+'/'+body['kf_id'],
+                           headers={'Content-Type': 'application/json'},
+                           data=json.dumps(body))
+    resp = json.loads(response.data.decode("utf-8"))
+
+    # Updates were made to the three versions
```
znatty22, line 266: Where did the 3 versions come from?

dataservice/extensions/flask_indexd.py
dankolbman: Still need the rev to construct the url

tests/genomic_file/test_genomic_file_resources.py
dankolbman: The mock indexd client returns three versions.

znatty22 (approved):
👍

parimalak (approved):

PR #447, opened 2018-09-20
 ✨Move consent_type from participant to biospecimen

parimalak:


migrations/versions/7f0b46d388a1_.py
```diff
@@ -0,0 +1,81 @@
...
+    biospecimen = sa.Table(
+        'biospecimen',
+        sa.MetaData(),
+        sa.Column('consent_type', sa.Text(), nullable=True))
+    results = connection.execute(sa.select([
+        participant.c.consent_type
+    ])).fetchall()
+
+    for cons in results:
+        connection.execute(biospecimen.insert().values(
```
dankolbman: How does this work? I would think that this would make a new biospecimen for every participant.

parimalak:
@dankolbman I changed the `insert` statement to `update` and tested locally. migrations are correctly updating the values.

migrations/versions/7f0b46d388a1_.py
```diff
@@ -0,0 +1,80 @@
...
+        sa.Column('consent_type', sa.Text(), nullable=True))
+    biospecimen = sa.Table(
+        'biospecimen',
+        sa.MetaData(),
+        sa.Column('consent_type', sa.Text(), nullable=True))
+    results = connection.execute(sa.select([
+        participant.c.consent_type
+    ])).fetchall()
+    for cons in results:
+        connection.execute(biospecimen.update().values(
```
dankolbman: What does this do?
Don't you also need to make sure you select the right biospecimen to update? As I understand it, 'results' is just a long list of consent_types. Shouldn't you also need to select the participant's biospecimens and then update each of those?
I think it may be something like `biospecimen.filter(participant_id=participant_id).update().values(consent_type=cons)`

dankolbman (approved):
Migrations look better!

znatty22 (approved):

PR #448, opened 2018-09-21
⚡️ Indexd bulk fetching

### LINKED ISSUE #442: Bulk request documents from indexd
Indexd now supports bulk get operations. This will be useful in paginated endpoints to reduce the number of requests made to indexd.

dankolbman:
## Need to release latest gen3 into production before release.

Using local indexd
Single request per entity in result
```
Getting 15719 results from /genomic-files in size of 100
Expected requests: 157
150/157
Total Requests: 158
Total Elapsed time: 232.80341982841492
Average Resp time: 1.4707968987341766
```

Using page prefetching
```
Getting 15719 results from /genomic-files in size of 100
Expected requests: 157
150/157
Total Requests: 158
Total Elapsed time: 32.46433401107788
Average Resp time: 0.2029419873417721
```

Fixes #442

### DIFF
--- a/dataservice/api/common/pagination.py
+++ b/dataservice/api/common/pagination.py
@@ -2,6 +2,7 @@ from flask import request, current_app
 from functools import wraps
 from dateutil import parser
 from datetime import datetime
+from dataservice.extensions import db, indexd
 
 
 def paginated(f):
@@ -101,6 +102,18 @@ def indexd_pagination(q, after, limit):
 
     :returns: A Pagination object
     """
+    def prefetch_indexd(after):
+        """ Compute dids for the page and have indexd fetch them in bulk """
+        model = q._entities[0].mapper.entity
+        gfs = (q.order_by(model.created_at.asc())
+                .filter(model.created_at > after)
+                .with_entities(model.latest_did)
+                .limit(limit).all())
+        dids = [gf[0] for gf in gfs]
+        indexd.prefetch(dids)
+
+    indexd.clear_cache()
+    prefetch_indexd(after)
     pager = Pagination(q, after, limit)
     keep = []
     refresh = True
@@ -108,17 +121,21 @@ def indexd_pagination(q, after, limit):
     # Continue updating the page until we get a page with no deleted files
     while (pager.total > 0 and refresh):
         refresh = False
-        # Move the cursor ahead to the last valid file
-        next_after = keep[-1].created_at if len(keep) > 0 else after
-        # Number of results needed to fulfill the original limit
-        remain = limit - len(keep)
-        pager = Pagination(q, next_after, remain)
-
         for st in pager.items:
             if hasattr(st, 'was_deleted') and st.was_deleted:
                 refresh = True
             else:
                 keep.append(st)
+
+        # Only fetch more if we saw there were some items that were deleted
+        if refresh:
+            # Move the cursor ahead to the last valid file
+            next_after = keep[-1].created_at if len(keep) > 0 else after
+            # Number of results needed to fulfill the original limit
+            remain = limit - len(keep)
+
+            prefetch_indexd(next_after)
+            pager = Pagination(q, next_after, remain)
 
     # Replace original page's items with new list of valid files
     pager.items = keep
--- a/dataservice/extensions/flask_indexd.py
+++ b/dataservice/extensions/flask_indexd.py
@@ -17,12 +17,18 @@ class Indexd(object):
 
     def __init__(self, app=None):
         self.app = app
+        # Used to store documents prefetched for a page
+        self.page_cache = {}
         if app is not None:
             self.init_app(app)
 
     def init_app(self, app):
         app.config.setdefault('INDEXD_URL', None)
         self.url = app.config['INDEXD_URL']
+        if self.url:
+            self.bulk_url = '/'.join(self.url.split('/')[:-1] + ['bulk/'])
+        else:
+            self.bulk_url = None
         if hasattr(app, 'teardown_appcontext'):
             app.teardown_appcontext(self.teardown)
         else:
@@ -42,6 +48,20 @@ class Indexd(object):
         if hasattr(ctx, 'indexd_session'):
             ctx.indexd_session.close()
 
+    def prefetch(self, dids):
+        """
+        Fetch a list of documents by did into the page cache.
+        """
+        # If running in dev mode, don't call indexd
+        if self.url is None or self.bulk_url is None:
+            return
+        resp = self.session.post(self.bulk_url + 'documents', json=dids).json()
+        for doc in resp:
+            self.page_cache[doc['did']] = doc
+
+    def clear_cache(self):
+        self.page_cache = {}
+
     def get(self, record):
         """
         Retrieves a record from indexd
@@ -54,6 +74,9 @@ class Indexd(object):
         # If running in dev mode, don't call indexd
         if self.url is None:
             return record
+
+        if record.latest_did in self.page_cache:
+            return self.page_cache[record.latest_did]
 
         url = self.url + record.latest_did
         resp = self.session.get(url)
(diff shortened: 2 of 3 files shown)


dataservice/api/common/pagination.py
```diff
@@ -101,6 +102,17 @@ def indexd_pagination(q, after, limit):
...
+        """ Compute dids for the page and have indexd fetch them in bulk """
+        model = q._entities[0].mapper.entity
+        gfs = (q.order_by(model.created_at.asc())
+                .filter(model.created_at > after)
+                .with_entities(model.latest_did)
+                .limit(limit).all())
+        dids = [gf[0] for gf in gfs]
+        indexd.prefetch(dids)
+
+    prefetch_indexd()
```
fiendish: This looks like it always prefetches twice without needing to. I think you only need the one down inside of `while (pager.total > 0 and refresh)`?

dataservice/api/common/pagination.py
```diff
@@ -112,6 +124,8 @@ def indexd_pagination(q, after, limit):
         next_after = keep[-1].created_at if len(keep) > 0 else after
         # Number of results needed to fulfill the original limit
         remain = limit - len(keep)
+
+        prefetch_indexd()
```
fiendish: It looks like this might not be prefetching the right dids here because `prefetch_indexd` absorbs the value of `after` given to `indexd_pagination` and doesn't track `next_after`

dataservice/extensions/flask_indexd.py
```diff
@@ -33,14 +39,28 @@ def new_session(self):
...
 
+    def prefetch(self, dids):
+        """
+        Fetch a list of documents by did into the page cache.
+        """
+        # If running in dev mode, don't call indexd
+        if self.url is None or self.bulk_url is None:
+            return
+        print(self.url, self.bulk_url)
+        self.page_cache = {}
```
fiendish: Clearing the cache here means that the looped calls to indexd.prefetch inside of indexd_pagination will step on each other. Is that a problem? It seems like it would be better to clear the cache at the beginning of indexd_pagination instead.

fiendish (changes requested):

dataservice/extensions/flask_indexd.py
```diff
@@ -33,14 +39,28 @@ def new_session(self):
...
             ctx.indexd_session.close()
 
+    def prefetch(self, dids):
+        """
+        Fetch a list of documents by did into the page cache.
+        """
+        # If running in dev mode, don't call indexd
+        if self.url is None or self.bulk_url is None:
+            return
+        print(self.url, self.bulk_url)
```
fiendish: print?

dataservice/api/common/pagination.py
dankolbman: Yeah I think this is unnecessarily fetching twice, though the first still has to be there so that it fetches before pulling in from the database when the `Pagination` object is created.

dataservice/extensions/flask_indexd.py
dankolbman: Yeah, that does seem to be an issue. I feel a little uneasy about clearing the cache outside of the indexd client though.

dataservice/api/common/pagination.py
```diff
@@ -101,25 +102,41 @@ def indexd_pagination(q, after, limit):
...
                 keep.append(st)
 
+        # Only fetch more if we saw there were some items that were deleted
+        if refresh:
+            # Move the cursor ahead to the last valid file
+            next_after = keep[-1].created_at if len(keep) > 0 else after
+            # Number of results needed to fulfill the original limit
+            remain = limit - len(keep)
+
+            prefetch_indexd(next_after)
```
fiendish: if the indexd bulk fetch happens _after_ the Pagination object is made instead of before, then you could also speed up empty returns by not checking indexd when total is 0.

dataservice/api/common/pagination.py
```diff
@@ -101,25 +102,41 @@ def indexd_pagination(q, after, limit):
...
+        model = q._entities[0].mapper.entity
+        gfs = (q.order_by(model.created_at.asc())
+                .filter(model.created_at > after)
+                .with_entities(model.latest_did)
+                .limit(limit).all())
+        dids = [gf[0] for gf in gfs]
+        indexd.prefetch(dids)
+
+    indexd.clear_cache()
+    prefetch_indexd(after)
```
fiendish, line 116: if indexd bulk fetch were to happen _after_ the Pagination object is initialized instead of before, then you could also speed up empty returns by not checking indexd this first time when total is 0. Is that doable?

fiendish:
I bet the timing numbers are even better now.

dataservice/api/common/pagination.py
dankolbman: Yeah there's probably a couple optimizations to be made around reducing the number of db queries here. It would involve adding some sort of branching into the Pagination object, though, and I'd prefer to keep it simple as it's generalized to all entities at the moment.

Right now, if the Pagination object were constructed first, it would result in the old behavior of constructing each object with its own request to indexd. Note that if there is a total=0, indexd won't actually be called as the query for gfs will return empty.

fiendish: got it. All clear from me, then.

fiendish (approved):

dataservice/api/common/pagination.py
```diff
@@ -101,25 +102,41 @@ def indexd_pagination(q, after, limit):
 
     :returns: A Pagination object
     """
+    def prefetch_indexd(after):
```
znatty22, line 105: Hmm indexd_pagination is confusing to me now.. What I would think you want is: 

1. Execute the query to get all of the objs for the page like you normally would with Pagination - without executing the request to indexd per instantiation of an indexd model.  
2. Collect the dids from the objs in the page of results
3. Send the bulk load request out to indexd to get the page of indexd docs
4. Iterate over the indexd docs and merge with the query results from step 1 
5. Return the results

I realize you'd have to refactor (maybe remove merge_indexd from constructor) and decouple things quite a bit (separate the indexd get request from the actual merging of an indexd doc with an indexd model instance). So maybe that's why you didn't do it this way.

But if you were able to implement the above, then you could probably also get rid of the entire while loop that checks for deleted files right? You would just be able to return the page right away.

dataservice/api/common/pagination.py
```diff
@@ -101,25 +102,41 @@ def indexd_pagination(q, after, limit):
...
     while (pager.total > 0 and refresh):
         refresh = False
-        # Move the cursor ahead to the last valid file
-        next_after = keep[-1].created_at if len(keep) > 0 else after
-        # Number of results needed to fulfill the original limit
-        remain = limit - len(keep)
-        pager = Pagination(q, next_after, remain)
-
         for st in pager.items:
             if hasattr(st, 'was_deleted') and st.was_deleted:
```
znatty22, line 125: Isn't this never going to happen now? Before the loop, you first populate the indexd page cache. Then you construct the Pagination obj, which results in a bunch of calls to indexd.get() which results in a bunch of indexd page cache lookups which means the requests to indexd never go out. So then you won't know if a file was deleted?

dataservice/api/common/pagination.py
dankolbman: That's true, but it would require decoupling the indexd property loading on construction of the object. Perhaps you could load the indexd properties only when you attempt to access them, but I'm not sure if that is possible.

dataservice/api/common/pagination.py
dankolbman: I believe this will still happen because the `merge_indexd()` is still called for every object in the page, only, it reads from the cache if available. If it's not in the cache, it will attempt to get it from indexd, in which case, it will return 404 and mark it as `was_deleted`.
Will look more closely at this to confirm.

fiendish:
Is this going to be finalized?

dankolbman:
@fiendish this is done from my POV. Waiting for more feedback from @znatty22 on if it needs to be refactored to be more understandable.

znatty22:
I think it could be refactored but it doesn't *need* to be. I think the last comment I made still needs to be addressed tho. @dankolbman was going to confirm:
> I believe this will still happen because the merge_indexd() is still called for every object in the page, only, it reads from the cache if available. If it's not in the cache, it will attempt to get it from indexd, in which case, it will return 404 and mark it as was_deleted.
Will look more closely at this to confirm.

baileyckelly:
Moving to backlog.

PR #450, opened 2018-10-03
📝 Add design document for imaging model

dankolbman:
Draft for additions to the datamodel needed for new file types.


docs/imaging_model/README.md
```diff
@@ -0,0 +1,47 @@
...
+only one additonal study in the future.
+Will this model be utilized at some point across more studies, or is it
+only being implemented to accomadate an outlier?
+
+The model changes introduce many more file types which will result in many
+more files being tracked in the dataservice and indexd.
+Indexd is currently a bottleneck, and will likely remain one.
+Will it be able to accommodate such an increase in quantity?
+
+
```
dankolbman: Also discussed general report and event tables that could relate to any other table but will put off for later discussion.

fiendish: How is indexd a bottleneck, and why is it? 
Generally speaking, mapping one identifier to another *should* be trivial and extremely fast.

docs/imaging_model/README.md
```diff
@@ -0,0 +1,47 @@
...
+inherit the Indexd model similar to GenomicFiles.
+
+![Imaging Model](ImagingModelTables.png)
+
+### ACLs
+
+These file types are typically not submitted to dbGaP and so can not be
+directly controlled through accession numbers and consents.
+Instead, these files will have ACLs consisting of the study kf_id and the
+most restricted set of consent codes for that participant's biospecimen.
```
fiendish: > most restricted set of consent codes for that participant's biospecimen.

"...either the consent code from the attached biospecimen (if attached directly to a biospecimen) or the most restrictive of all of the consent codes for that participant's biospecimens (if attached to a participant)."

docs/imaging_model/README.md
```diff
@@ -0,0 +1,47 @@
...
+These files will need to be moved to RadiologyFiles.
+Because this was intentionally done as a temporary solution, we will not make
+an attempt to preserve the kf_ids given to the files.
+
+
+## Difficulties
+
+These models and data files are currently only supported by CBTTC with
+only one additonal study in the future.
+Will this model be utilized at some point across more studies, or is it
```
fiendish, line 58: I don't think that this matters except for prioritizing development. The design should make sense regardless of how many studies end up contributing this kind of data.

docs/imaging_model/README.md
dankolbman: We are not just mapping an identifier, though. We need to create documents within indexd and any interaction with files requires sub-requests to indexd.

docs/imaging_model/README.md
allisonheath: There's no real hierarchy to the consent codes - either people have access based on their research project or they don't. While these files might not be managed by dbGaP, we'll need to be able to take in some way whether the files are open access, or which specific consent group they should be covered under.

docs/imaging_model/README.md
allisonheath: I think the main thing for here is that CBTTC is cancer, and so this is cancer-focused with the pathology slides and MRIs. So it should make sense for other (likely cancer) studies that have those modalities of data. Why I think it would be useful to look at TCIA (http://www.cancerimagingarchive.net/) . However, I expect other imaging, such as photographs for phenotyping to require further development in the future.

docs/imaging_model/README.md
fiendish: Sure, but aside from the time it takes to deliver our packets via lethargic carrier pigeon, indexd itself should be lightning fast because it's doing hardly anything. Is there a way to speed up the request pathway?

dankolbman: It's not terribly slow in performing it's logic, but it requires a web request per read and at least two http requests for everything else. This is a lot of overhead that slows it down when doing things in bulk.

allisonheath:
From discussion today: 
- Move the reports to be directly associated with the biospecimens
- Discussed biospecimen implications with histology images, there's probably a need for revisiting this, but at this time we'll keep as it where the histology image is connected to (just the one) virtual biospecimen

allisonheath (dismissed):
See comment in the PR convo

znatty22 (dismissed):
Can we add these notes in the implementation section:

1. RadiologyFile and RadiologyReport have a many-to-many relationship and they are both linked to Participant. When linking a file with a report, we must ensure that they both share the same participant. This is similar to what we do when linking biospecimens and diagnoses.

2. HistologyImage and PathologyReport have a many-to-many relationship and they are both linked to Biospecimen. When linking an image with a report, we must ensure that they both share the same biospecimen. This is similar to what we do when linking biospecimens and diagnoses.

docs/imaging_model/README.md
```diff
@@ -0,0 +1,65 @@
...
+only being implemented to accomadate an outlier?
+
+The model changes introduce many more file types which will result in many
```
allisonheath, line 61: Also should note that means there will be multiple prefixes for the identifiers each type of file. I think this is mostly a good thing because it'll be a quick indication to the user the type of file. The con would be if you misclassified a file that gets released could have two different kf_ids. In this case though, the did should stay the same, correct?

allisonheath (changes requested):

docs/imaging_model/README.md
dankolbman: How likely are you really to misclassify a file, though? Although theoretically possible to maintain the did, it probably wouldn't be worth the effort and we would move it by deleting it from the former table and adding it to the correct one.

znatty22 (approved):
👍

docs/imaging_model/README.md
allisonheath: I think as we get more types of files it becomes increasingly likely. I've also seen the files delivered mixed up as well, so unless you're checking all of the files you can miss them.

docs/imaging_model/README.md
```diff
@@ -0,0 +1,64 @@
...
+![Imaging Model](ImagingModelTables.png)
+
+### Radiology File and Report
+
+RadiologyFile and RadiologyReport have a many-to-many relationship to
+each-other and one-to-many from Participant.
+When linking a file with a report, we must ensure that they both share the
+same participant. This is similar to what we do when linking biospecimens and
+diagnoses.
+
```
allisonheath, line 28: Also realizing we forgot to include operations reports (they are listed in the original ticket). These are actually the surgical operations reports and this is starting to highlight how we're really missing the concept of a clinical event to relate the appropriate images, reports, biospecimens and diagnoses. It may be reasonable to associate them directly with the participant, but in reality they're tangentially associated with certain radiology images/reports (pre/post op), but wouldn't necessarily be related to follow up radiology images/reports. The problem here is the same that caused us to associate diagnoses and biospecimens. If I'm looking at a genomic file, how do I know which radiology/operation data is associated with that genomic data, as a patient can have multiple different tumors over time.

allisonheath (changes requested):

docs/imaging_model/README.md
```diff
@@ -0,0 +1,64 @@
...
+Many MRI files have already been loaded as Genomic Files.
+These files will need to be moved to RadiologyFiles.
+Because this was intentionally done as a temporary solution, we will not make
+an attempt to preserve the kf_ids given to the files.
+
+
+## Difficulties
+
+These models and data files are currently only supported by CBTTC with
```
rosibaj, line 56: What about extension for other studies image files? Other studies could have file types that are non-radiological but are images.  The model should work for those as well.

docs/imaging_model/README.md
```diff
@@ -0,0 +1,64 @@
...
+MRI images, histology images, and reports.
+Some of these file types relate directly to the biospecimen rather than the
+participant, eg: proteomics and histology.
+In order to capture and distribute these files, we need to model them
+appropriately.
+
+## Implementation
+
+Files will need to be distributed through gen3, so the models will need to
+inherit the Indexd model similar to GenomicFiles.
```
rosibaj, line 15: It would make sense to have a central File entity that listed the common elements of files/index_d terms that apply to everything, otherwise you have redundant fields. e.g of standard fields 
- file name
- latest_did 
- id
- file format
- md5
- size

Will there be more file types that don't fit in either genomic file or imaging file that other extensions would need to be made in a similar manner?  It seems weird that "files" all have different id types.

znatty22: @rosibaj We do have a common File entity. In the embedded image on README.md, the orange colored IndexdFile model has all of the common fields you're referring to. I think you have to hit the View button to actually see the rendered markdown which includes the image

docs/imaging_model/README.md
dankolbman: In that case, we would add new tables with appropriate metadata.

docs/imaging_model/README.md
```diff
@@ -0,0 +1,64 @@
...
+metadata fields unique to that type of file.
+
+![Imaging Model](ImagingModelTables.png)
+
+### Radiology File and Report
+
+RadiologyFile and RadiologyReport have a many-to-many relationship to
+each-other and one-to-many from Participant.
+When linking a file with a report, we must ensure that they both share the
+same participant. This is similar to what we do when linking biospecimens and
```
junjun-zhang, line 26: Is it true that every RadiologyReport must have one or more RadiologyFiles related to it? If so, it should be better to remove the relationship between RadiologyReport and Participant because this relationship can be derived from relationship between RadiologyFile and Participant, hence is redundant.

Also, the many-to-many relationship between RadiologyFile and RadiologyReport will need to be modelled using a middle table that has one-to-many relationships to both RadiologyFile and RadiologyReport. Such middle table may include three columns as composite key, something like below:

| participant_id | radiology_file_id | radiology_report_id |

This will at db level guarantee `When linking a file with a report, we must ensure that they both share the
same participant.`

Same comment for `HistologyImage` and `PathologyReport` tables below.

allisonheath: @junjun-zhang it's not true, that's actually why we added it (we started with what you suggested), we can have RadiologyReports and the RadiologyFiles can be missing. A patient can also have multiple RadiologyImages that are considered at different times as part of different RadiologyReports over time. 

However, I think we need to put this PR on hold as it's clear our data model isn't sufficient to represent what we need to with clinical events/encounters. We're going to be delving into OMOP CDM to see if this is a better general solution.

docs/imaging_model/README.md
allisonheath: @znatty22 yes, in the code, but in the actual database and API this will be manifested as different entities, right?

dankolbman:
Closing while we explore alternative models.

PR #453, opened 2018-10-09
♻️ Standardize read_group_genomic_file entity

znatty22:
Refactor read_group, genomic_file, and read_group_genomic_file model/resources so that they all follow standard Kids First Dataservice format. 

Each entity has a model which has fields that map 1-1 to cols in that entity's db table. Each entity has CRUD endpoints that allow CRUD operations on just that entity in the db. This includes many-to-many association tables like read_group_genomic_file.

**Now there is a /read-group-genomic-files endpoint exposed and you can do all CRUD operations on it.**

**Can do filtering via:** 
- `/read_groups?genomic_file_id=foo` to get read_groups associated with a particular genomic_file 
- `/read_group-genomic_files?genomic_file_id=foo` to get read_group_genomic_files associated with a particular genomic file
- `/genomic-files?read_group_id=foo` to get genomic_files associated with a particular read_group 
- `/read_group-genomic_files?read_group_id=foo` to get read_group_genomic_files associated with a particular read_group 

@liberaliscomputing @fiendish Since @parimalak is out for the next 3 weeks, I'll need a second approval from one of you two.


dataservice/api/common/views.py
```diff
@@ -174,8 +175,22 @@ def endpoints(cls):
...
+            - Sends request as an event to sns
+
+            - Execute each request with sqlalchemy autoflush turned off.
+              This prevents the model event listeners from triggering
+              inadvertently. It happens when the db session goes out of scope
+              and when Marshmallow loads an object into the session and does a
+              merge during a patch request. Its better to explicitly flush
+              the session.
         """
+        # Autoflush off
```
fiendish: This comment doesn't add information

tests/read_group/test_read_group_resources.py
```diff
@@ -196,92 +135,3 @@ def _create_save_to_db(self):
         db.session.commit()
 
         return rg
-
-    def test_filter_by_gf(self):
```
fiendish, line 200: why delete these tests?

tests/genomic_file/test_genomic_file_resources.py
```diff
@@ -366,60 +303,7 @@ def exc():
     assert GenomicFile.query.count() == init + 1
 
 
-def test_filter_by_rg(client, indexd):
```
fiendish: why delete these tests?

dataservice/api/common/views.py
fiendish, line 190: obvious

fiendish (changes requested):

tests/read_group/test_read_group_resources.py
znatty22: These tests are no longer needed since we've removed the functionality that it tests. You can no longer go to the /genomic-files endpoint and filter by read_group_id nor can you go to the /read-groups endpoint and filter by by genomic_file_id. You'll now have to go to the /read-group-genomic-files endpoint to filter associations

znatty22: The general filtering tests which include  /read-group-genomic-files are in tests/test_filter_params.py

tests/genomic_file/test_genomic_file_resources.py
znatty22: These tests are no longer needed since we've removed the functionality that it tests. Same rationale as above.

fiendish (approved):
🙆‍♂️

tests/read_group/test_read_group_resources.py
fiendish: > You can no longer go to the /genomic-files endpoint and filter by read_group_id nor can you go to the /read-groups endpoint and filter by by genomic_file_id. You'll now have to go to the /read-group-genomic-files endpoint to filter associations

lol ok

dataservice/api/read_group/models.py
```diff
@@ -27,9 +27,14 @@ class ReadGroup(db.Model, Base):
     quality_scale = db.Column(db.Text(),
                               doc='The scale used to encode quality scores')
 
-    genomic_files = db.relationship('GenomicFile',
-                                    secondary='read_group_genomic_file',
-                                    backref=db.backref('read_groups'))
+    genomic_files = association_proxy(
```
dankolbman, line 30: Is this just for convenience?

dataservice/api/read_group/models.py
```diff
@@ -27,9 +27,14 @@ class ReadGroup(db.Model, Base):
...
+        'read_group_genomic_files', 'genomic_file',
+        creator=lambda genomic_file:
+        ReadGroupGenomicFile(genomic_file=genomic_file))
```
dankolbman: Kinda difficult to tell that this is actually the body of the lambda. Indent or move up a line?

dataservice/api/read_group/models.py
```diff
@@ -43,20 +48,20 @@ class ReadGroupGenomicFile(db.Model, Base):
...
     read_group_id = db.Column(KfId(),
                               db.ForeignKey('read_group.kf_id'),
                               nullable=False)
 
     genomic_file_id = db.Column(KfId(),
                                 db.ForeignKey('genomic_file.kf_id'),
                                 nullable=False)
-    read_group = db.relationship('ReadGroup')
-    genomic_file = db.relationship('GenomicFile')
+    external_id = db.Column(db.Text(),
```
dankolbman, line 59: Do we have any of these?

dataservice/api/read_group/models.py
znatty22: Yea its for testing mostly so that I can directly access the genomic_files of a read_group

dataservice/api/read_group/models.py
znatty22: This is super useful during ingest debugging and dev because you can set the external_id of a read_group_genomic_file to be the read_group external_id concatenated with the genomic_file external_id. I do this for pretty much any entity representing an association table. Its so nice to be able to query the entity and see the external_ids in addition to the KF IDs in the results.  If we didn't have this field then when I query  read_group_genomic_file I would get back a list of those entities along with the KF ID for the read_group and the KF ID for the genomic file for each result and then I'd have to go into my KF ID - external ID cache file to look up what the external IDs are for the read group and the genomic file.

dankolbman (approved):
🎉

dankolbman (changes requested):
Need to wait for next release

dataservice/api/read_group/models.py
fiendish: We should stop abusing "external ID" for internal debugging values. Maybe add another generic field to all entities.

znatty22: Why? Its where I actually do store the contributor's ID for an entity, and I do use that to help debug/verify that all of my entities got into the dataservice

znatty22: Maybe we should just add external_id to the base model so everything has it?

fiendish:
boo waiting!

dataservice/api/read_group/models.py
fiendish: external_id should be in the base model, but also synthetic_id should be there too. Using the same field for both muddies the meaning. If they don't provide their own id, us filling in that field with something else is deceptive

dataservice/api/read_group/schemas.py
```diff
@@ -40,15 +31,6 @@ class Meta(BaseSchema.Meta):
     _links = ma.Hyperlinks({
         'self': ma.URLFor(Meta.resource_url, kf_id='<kf_id>'),
         'collection': ma.URLFor(Meta.collection_url),
-        'genomic_files': ma.URLFor('api.genomic_files_list',
-                                   read_group_id='<kf_id>')
+        'read_group_genomic_files': ma.URLFor(
```
fiendish, line 39: This link on its own is useless to the API user when what they really want is the list of genomic files.

fiendish (changes requested):
> You can no longer go to the /genomic-files endpoint and filter by read_group_id nor can you go to the /read-groups endpoint and filter by by genomic_file_id. You'll now have to go to the /read-group-genomic-files endpoint to filter associations

I think this is a mistake and adds an extra (not useful) layer to the API.

dataservice/api/genomic_file/schemas.py
```diff
@@ -77,15 +67,6 @@ class Meta(BaseSchema.Meta, IndexdFileSchema.Meta):
             'api.cavatica_task_genomic_files_list', genomic_file_id='<kf_id>'),
         'biospecimen_genomic_files': ma.URLFor(
             'api.biospecimen_genomic_files_list', genomic_file_id='<kf_id>'),
-        'read_groups': ma.URLFor(
-            'api.read_groups_list', genomic_file_id='<kf_id>')
+        'read_group_genomic_files': ma.URLFor(
```
fiendish, line 77: This link on its own is useless to the API user when what they really want is the list of read groups.

dataservice/api/read_group/schemas.py
znatty22: Hmm yea probably should have kept this functionality. I think I thought I couldn't do this for some reason if I reverted back to the "standard" endpoint format.. let me see if i can add this functionality back in while still keeping the /read-group-genomic-files endpoint

dataservice/api/genomic_file/schemas.py
znatty22: Ditto ^

fiendish (commented):
test_filter_by_study?

znatty22:
@fiendish test filter by study for all of our endpoints is in tests/test_pagination.py. 

Any model attributes used as filters get tested in tests/test_filter_params.py. Any non-model attributes (i.e. study_id or say diagnosis_id for /biospecimens) are special filters that can't be tested with the same logic as in test_filter_params.py since that module assumes the filter fields being tested are attributes on the model instance. 

That's why for special filters we test them separately, usually inside the resource tests for a particular model. OR in the case of study_id, we test this filter in test_pagination.py.

I'm sure we can do a much better job organizing and naming our tests, and maybe should consider another refactor. But for now ¯\\\_(ツ)\_\/¯

fiendish (approved):

dankolbman (approved):

rosibaj (approved):
Makes sense to standardize - Since the portal ETL does not inject read-group and ignores read-group on genomic-files I don't think this will break anything. (@joneubank can also confirm)

dataservice/api/read_group/models.py
```diff
@@ -43,20 +48,20 @@ class ReadGroupGenomicFile(db.Model, Base):
...
+                            doc='external id used by contributor')
 
 
-@event.listens_for(GenomicFile, 'after_delete')
+@event.listens_for(ReadGroupGenomicFile, 'after_delete')
```
dankolbman, line 63: Is this still necessary? I would think that proper cascading would deal with this appropriately.

znatty22: @dankolbman Yea we still need this. Remember cascade delete orphans are usually for child concepts like ReadGroupGenomicFile, not parent concepts like ReadGroup. So we need something that will listen for when all children of the parent are deleted, and then delete the parent since it will be an orphan at that point.

dankolbman (approved):

Issue #431, opened 2018-08-29
Suggestions on expanding identifier space


allisonheath:
Several different threads have been talking about how with some of the large count entities, such as phenotypes, diagnosis and files we're starting to feel cramped in our identifier space. Chances of collision are getting higher and even with suggestions like #423 there's potential to start having to do a lot of retries to get an ID. 

Two initial strategies to discuss (other ideas welcome) would be:

1) To uniformly expand the identifier space across all entities. E.g similar style but longer identifiers and/or something similar like https://github.com/ulid/spec.

2) To selectively expand identifier space for entities with large numbers, e.g. Files, Phenotypes and Diagnoses could use UUID4, especially because for phenotypes and diagnoses users don't typically have any use for specific identifiers for those.

dankolbman:
Option 2 sounds pretty good so we can keep tidy ids on the high-vis entities like participants, studies, and biospecimens.

We should also consider refactoring our diagnoses/phenotype models as that could help reduce their number considerably. Right now we have events denormalized with the phenotype. We could improve quality and mapping consistency by normalizing the phenotype/diagnoses with their ontology ids out from the events/source text/timestamp data. This would also be helpful in establishing an temporally focused model.

fiendish:
My preference would be to not generate IDs randomly at all since they're all coming from a central arbiter and could just be an incrementing counter. I don't see any benefit for the dataservice to have nonsequential identifiers.

Random ID generation exists to solve problems that we don't have (distributed simultaneous ID generation with unknown initial states) while introducing problems for scenarios that we do have (accidental reuse).

Of the two options presented above, I would vote for option 1 using ULID or something like it across the board, because I don't favor fragmentation and because, if you ignore how much of its bit-space is wasted on all of the times when entities are not being created, it basically acts like a complicated incrementer. For the sake of biospecimens that have been embedded inside genomic files (😭) we would need to still interoperate with the old style.

PR #454, opened 2018-10-09
✨ UID design considerations document

fiendish:
need feedback and discussion


fiendish:
Pursuant to #431

docs/kf_uids/uid_considerations.md
```diff
@@ -0,0 +1,220 @@
...
+---
+
+## Notes
+
+* Key generation (how many bits, how they're allocated) is distinct from encoding (which characters to use).
+* Many encoding techniques have functional disconnects between goal and implementation, e.g.
+  * **ulid** says "No special characters (URL safe)", but it's not necessary to use *no* special characters to be URL safe! **$-_.+!*'(),** are all safe without urlencoding and significantly visually distinct.
+  * **Crockford's Base32 encoding** says "[uppercase] L Can be confused with 1". This flies in the face of all available evidence. What font does that? Why not also remove G/6, B/8, Z/2, or S/5 then?
+* Many key encoding techniques make up arbitrary rules that apply only to very specific scenarios.
+  * **Crockford's Base32 encoding** says "Humans should be able to accurately transmit the symbols to other humans using a telephone." This is not a valid concern for our system. That is good for when you need to call a customer service agent and tell them your account ID number, which is very far removed from any of our use cases. We have no scenarios where this restriction should apply.
```
dankolbman, line 14: One of the underlying motivations is that identifiers be used in natural conversation, so I would say that this is a concern.

docs/kf_uids/uid_considerations.md
```diff
@@ -0,0 +1,220 @@
...
+* **Randomness never guarantees uniqueness.** Ever. Regardless of the number space or generator strength. We can do better than that.
```
dankolbman, line 15: Except for when it does well enough for our needs.

docs/kf_uids/uid_considerations.md
```diff
@@ -0,0 +1,220 @@
...
+* None of the 5 basic UUID/GUID variants described on wikipedia.org/wiki/Universally_unique_identifier ensure uniqueness. They are all misnamed.
```
dankolbman, line 16: They ensure uniqueness for practical applications, which is what we have.

docs/kf_uids/uid_considerations.md
```diff
@@ -0,0 +1,220 @@
...
+
+## We definitely want
+
+* Eliminate key reuse and collisions
+* Compact representation
+* Visually differentiable (no commonly hyperambiguous characters)
+* Typable
+
+## We don't need
```
dankolbman, line 25: These things seem non-critical but possibly beneficial.

docs/kf_uids/uid_considerations.md
```diff
@@ -0,0 +1,220 @@
...
+(many systems value these things for sometimes legitimate and sometimes illegitimate reasons)
+
+* Chronological sorting (though recording time is useful for uniqueness)
+* Reading out loud over the phone
+* Easy typing on mobile keyboards
+* Checksums
+
+## We don't know if we need
+
+* Decentralized generation (could rule out keyservers)
```
dankolbman, line 36: We should assume we need this, in case of dataservice divergence and convergence at a later time.

docs/kf_uids/uid_considerations.md
```diff
@@ -0,0 +1,220 @@
...
+---
+
+## Key Generation Techniques
+
+(randomness, external incrementer, internal incrementer, physical ID, logical ID)
+
+### Methods for differentiating my own keys from each other
+
+* Random generation in a very large space
+  * <font color="darkorange">Insufficient. Improbable things happen all the time.</font>
```
dankolbman, line 47: Sufficient for our needs with enough entropy.

docs/kf_uids/uid_considerations.md
```diff
@@ -0,0 +1,220 @@
...
+* Counting
+  * <font color="lightseagreen">Guaranteed for serial generation within max_increment.</font>
+* Timestamp
+  * <font color="lightseagreen">Guaranteed for extra-interval generation.</font>
```
dankolbman, line 51: Not guaranteed with finite precision clocks. Perhaps better if mixed with random generation.

docs/kf_uids/uid_considerations.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Methods for differentiating my keys from someone else's keys
```
dankolbman, line 53: Does this seem like it would be useful?

docs/kf_uids/uid_considerations.md
```diff
@@ -0,0 +1,220 @@
...
+
+* **Twitter (snowflake)** uses <font color="lightseagreen">a timestamp, a node ID, and a counter</font>
+* **Instagram** uses a <font color="lightseagreen">timestamp, a node ID, and a counter</font>
+* **Flickr** uses <font color="lightseagreen">a central ticket server</font> with a simple <font color="lightseagreen">counter</font>
+  * (technically multiple servers where the number space is subdivided among them to avoid single-point-of-failure)
+* **TUID** uses <font color="lightseagreen">a timestamp, a node ID, a counter, <font color="darkorange">and some random bits</font></font>
+* **ULID** uses <font color="lightseagreen">a timestamp and a counter <font color="darkorange">that starts randomized</font></font>
+* **MongoDB ObjectId** uses <font color="lightseagreen">a timestamp, <font color="darkorange">some random bits</font>, and a counter <font color="darkorange">that starts randomized</font></font>
+
+### Node IDs
```
dankolbman, line 72: Is a node a process in this case?

docs/kf_uids/uid_considerations.md
```diff
@@ -0,0 +1,220 @@
...
+### Server ID + Counter
+
+* 19 bits for Server ID (500 thousand different servers)
+* 45 bits for Counter
+  * 35 trillion UIDs per server
+  * If exhausted (lol), request a new Server ID
+
+### Session ID + Counter
+
+* 44 bits for Session ID (17 trillion sessions)
```
dankolbman: What is a session here?

docs/kf_uids/uid_considerations.md
```diff
@@ -0,0 +1,220 @@
...
+* 20 bits for Counter
+  * 1 million UIDs per session
+  * If exhausted (lol), get a new Session ID
+
+### Millisecond Timestamp + Node ID + Randomized Sub-chronon Counter
+
+* 41 bits for milliseconds since custom epoch (from now until 69.5 years from now)
```
dankolbman: Seconds since epoch would probably work just as well.

docs/kf_uids/uid_considerations.md
fiendish: > One of the underlying motivations is that identifiers be used in natural conversation

I've yet seen no practical scenarios where this would add any value. Even with our current 8 character IDs, the most we ever communicate verbally is "SD P9 yada yada" or some approximate vocalization like "zifgediggis", and that only ever happens when debugging which means that typing or copy/pasting it would have been better anyway.

docs/kf_uids/uid_considerations.md
fiendish: That's what I meant by "extra-interval generation". When the clock steps forward you get a different number than any previous one. Timestamps are definitely not sufficient on their own.

docs/kf_uids/uid_considerations.md
fiendish: Yes, it's absolutely necessary for two different processes to not generate overlapping IDs.

docs/kf_uids/uid_considerations.md
fiendish: Or a device (if a given device only runs one generating process at a time) or a session.

docs/kf_uids/uid_considerations.md
fiendish: Like a process except if your session runs out of available keyspace before you finish then you can just ask for another one.

"Hi, I would like to give you a bunch of kfids, can I have a session ID?"
"Your session ID is 5"
"...<stream of IDs using session ID 5>..."
"Hi, I would like to give you a bunch of kfids, can I have a session ID?"
"Your session ID is 6"
"...<stream of IDs using session ID 6>..."

docs/kf_uids/uid_considerations.md
fiendish: I think it depends on who and where keys are being generated and how many bits you want to allocate. I included second-based timestamps too for reference.

docs/kf_uids/uid_considerations.md
fiendish: safe > sorry

docs/kf_uids/uid_considerations.md
fiendish: > They ensure uniqueness for practical applications

By definition they do not.

docs/kf_uids/uid_considerations.md
fiendish: Do you have any practical problems in mind that any of these would solve?

docs/kf_uids/uid_considerations.md
fiendish: Even distributed dataservices could still get their UIDs from official keyservers, or they could use assigned dataservice Node IDs. That doesn't require decentralized key generation. Decentralized means keys generated entirely by the client (my laptop for instance).

docs/kf_uids/uid_considerations.md
dankolbman: To achieve a 50% chance of one or more collisions, you need to generate on the order of `10**18`. This an enormous number. [see](https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions)

docs/kf_uids/uid_considerations.md
dankolbman: It depends on your definition of practical.

docs/kf_uids/uid_considerations.md
dankolbman: Chronological sorting seems the most useful, as it's already how we paginate results in dataservice.

Not sure how checksums would be applied, but not likely to be used as we make updates constantly.

docs/kf_uids/uid_considerations.md
fiendish: Can you put bounds on sufficient? I think we can/should do better than improbable, because improbable things happen too.

docs/kf_uids/uid_considerations.md
fiendish: Basically think of session IDs as disposable rather than persistent. That's the only difference.

docs/kf_uids/uid_considerations.md
fiendish: That's true, but I believe that it's a misapplication of probability to treat "infrequent" as "won't happen". Your first collision could come up on the second roll. Probability says nothing about when a collision will occur.

docs/kf_uids/uid_considerations.md
fiendish: I do agree that chronological sorting has the most practical value of those items, but would it be different than indexing on the dedicated timestamp field that we already have?

docs/kf_uids/uid_considerations.md
dankolbman: Well, I'd say a acceptable bound of collision probability is around 1 in *10**9* (about the inverse of a moderate number of entities we assign ids). In order to reach a point where we have a one in *10**9*  probability of collision (ie we surpass our tolerance for collision), we would need *10**14* entities. Making the conservative assumption that we add *10**6* entities per day, that would give us *10**8* days until we pass our threshold. That's a quarter million years of time. Probably longer than the Kids First will endure.

docs/kf_uids/uid_considerations.md
dankolbman: Having a decentralized key service is probably something we want to avoid.

docs/kf_uids/uid_considerations.md
fiendish: Or you could get a collision on the first try and then never get another one for 10**verybig, but you still got one quickly.

docs/kf_uids/uid_considerations.md
fiendish: Avoid why? (did you mean centralized instead of decentralized?)

docs/kf_uids/uid_proposal.md
```diff
@@ -0,0 +1,38 @@
+# Proposal
+
+Based on [uid_considerations.md](uid_considerations.md) and [uid_migration.md](uid_migration.md)
+
+## 64 bit keys
+
+### Nodeless (much better than what we have now)
```
allisonheath: From today, I think Nodeless is the better option and operationally easier to manage.

docs/kf_uids/uid_proposal.md
```diff
@@ -0,0 +1,38 @@
...
+* 31 bit seconds since custom epoch (now until 68 years from now)
+* 10 bit dataservice instance ID (1024 server instances)
+  * Requires a dedicated ID assignment service (When an autoscale dataservice task spins up, the first thing it does is request a Node ID from a single task node assignment service. The assignment service just increments after each request up to 1024 and then resets.)
+* 23 bit randomized counter (8 million keys in every second)
+
+## Encoding
+
+Studies need to keep old key format for compatibility with Amazon S3 bucket name restrictions, but we can improve the encoding for everything else.
+
+Either base57 or better-than-crockford-base32 (see [uid_considerations.md # Modifying Crockford's Base32 symbol set](uid_considerations.md))
```
allisonheath: Decided that, while flawed, Crockford's Base23 symbol set is good enough for our needs and means one less thing changing with the id migration.

docs/kf_uids/uid_proposal.md
```diff
@@ -0,0 +1,38 @@
...
+Based on [uid_considerations.md](uid_considerations.md) and [uid_migration.md](uid_migration.md)
+
+## 64 bit keys
+
+### Nodeless (much better than what we have now)
+
+* 31 bit seconds since custom epoch (now until 68 years from now)
+* 33 bit randomized counter (8.5 billion keys for every second)
+* Cross fingers
+
```
allisonheath, line 10: One thing that came up is which bits go where, this is mostly an aesthetic choice. Three options:

1. Seconds bits and then randomized bits
2. Randomized and then second bits
3. Interleaving bits

3 would look mostly random. 1 would make the identifiers easily sortable/scannable with a sense of time. 2 would look random, but a careful eye would see the last bit is incremental. Each has pros/cons in terms of user usage, examples might help @fiendish ?

If we don't do a mapping table for migration, there might be other considerations about using zero time to migrate the previous identifiers.

docs/kf_uids/uid_migration.md
```diff
@@ -0,0 +1,23 @@
...
+
+* GET requests where the requested ID looks like the old style should bounce through a translation lookup to get back the corresponding new ID and then return the entity with the new ID in place.
+* Will need a new table in database with two columns: (Old ID, New ID).
+  * When adding a new entry to that table, it should see if any existing entries have their New ID equal to the new Old ID. If so, they should be updated to have new New ID. That way all old IDs will always point to the latest ID for the corresponding entity, even if we migrate IDs multiple times.
+
+### No need to re-header existing genomic files
+
+Our harmonized genomic files have had biospecimen kfids baked into their bits (not just s3 tags). Updating those would cost extra 💰 beyond just the labor involved. (How much, I don't know. Probably not _thaaaaat_ much.)
+
+We don't need to though. As long as the GET translation is in place, the old KFIDs will continue to be valid.
```
allisonheath, line 19: After some discussion about having padding bits (time = 0) and/or keeping the old kf_ids and just creating new ones, seems like the best path forward is to replace all of the kf_ids with the new format and have a mapping as described above. The main think to ensure is what the workflow looks like for the portal and for users who have genomic files in hand with BS IDs, how they would retrieve that information from the portal.

allisonheath (dismissed):
Added my comments from today's discussion. We should update the document with the decisions we made today and then circulate with both OICR and the BIX teams (esp with the migrations) before approving.

docs/kf_uids/uid_migration.md
allisonheath: Also just for my own notes - this doesn't apply to the study ids which will remain the old scheme.

PR #455, opened 2018-10-10
♻️ Standardize biospecimen_diagnosis entity

znatty22:
Need to rebase once #453 merges

Refactor biospecimen, diagnosis, and biospecimen_diagnosis model/resources so that they all follow standard Kids First Dataservice format. 

Each entity has a model which has fields that map 1-1 to cols in that entity's db table. Each entity has CRUD endpoints that allow CRUD operations on just that entity in the db. This includes many-to-many association tables like biospecimen_diagnosis.

Can filter by:
- `/biospecimens?diagnosis_id=foo` to get biospecimens associated with a particular diagnosis
-  `/biospecimen-diagnoses?diagnosis_id=foo` to get biospecimen_diagnoses associated with a particular diagnosis
- `/diagnoses?biospecimen_id=foo` to get diagnoses associated with a particular biospecimen
- `/biospecimen-diagnoses?biospecimen_id=foo` to get biospecimen_diagnoses associated with a particular biospecimen


fiendish (changes requested):
> Instead of doing /biospecimens?diagnosis_id=foo to get biospecimens associated with a particular diagnosis now one will do /biospecimen-diagnoses?diagnosis_id=foo
Instead of doing /diagnoses?biospecimen_id=foo to get diagnoses associated with a particular biospecimen now one will do /biospecimen-diagnoses?diagnosis_id=foo

I think this is a mistake. Why can't /biospecimens?diagnosis_id=foo etc continue to work? If I'm looking for biospecimens, not some intermediate relationship entity, I don't want to have to bounce through an extra step to get them.

dankolbman:
Hm yes, would it be possible to continue to allow the filter by diagnosis id on biospecimens and vice versa? We could still expose the intermediate model to allow batch processing, but keeping that filter and the links would allow nicer usability.

znatty22:
@fiendish @dankolbman Yea, I just commented on the other PR. I think a while back I tried to do something which made me think that I couldn't have this functionality with the "standard" endpoint implementation. Thats probably not true. Gonna try and bring that filtering functionality back

fiendish:
oh no, some tests were not successful! ☠️ ☠️ ☠️

fiendish:
oh no, some tests were not successful! ☠️ ☠️ ☠️

dataservice/api/biospecimen/models.py
```diff
@@ -131,56 +110,43 @@ class Biospecimen(db.Model, Base):
...
     biospecimen_genomic_files = db.relationship(BiospecimenGenomicFile,
                                                 backref='biospecimen',
                                                 cascade='all, delete-orphan')
-    diagnoses = db.relationship('Diagnosis', secondary='biospecimen_diagnosis',
-                                backref=db.backref(
-                                    'biospecimens'))
 
 
-def validate_biospecimen(target):
+class BiospecimenDiagnosis(db.Model, Base):
```
dankolbman, line 122: It's often good to avoid moving around blocks like this unless it's really needed so as to make commits a little cleaner.

znatty22: I had to move BiospecimenDiagnosis below Biospecimen, bc my reference to Biospecimen inside BiospecimenDiagnosis wasn't resolving.

dataservice/api/biospecimen/schemas.py
```diff
@@ -54,11 +52,12 @@ class Meta(BaseSchema.Meta):
         'participant': ma.URLFor('api.participants', kf_id='<participant_id>'),
         'sequencing_center': ma.URLFor('api.sequencing_centers',
                                        kf_id='<sequencing_center_id>'),
-        'diagnoses': PatchedURLFor('api.diagnoses_list',
-                                   biospecimen_id='<kf_id>'
-                                   ),
         'biospecimen_genomic_files': ma.URLFor(
```
fiendish, line 55: I know it's not technically part of this, but can we also get genomic_files?

fiendish (approved):

dataservice/api/biospecimen/schemas.py
znatty22: Let me make another issue and PR for that

tests/biospecimen_diagnosis/test_biospecimen_diagnosis_resources.py
```diff
@@ -0,0 +1,294 @@
...
+        response = response['_status']
+        assert 400 == response['code']
+        assert 'cannot be linked with' in response['message']
+        bd = BiospecimenDiagnosis.query.first()
+
+        # Check db - nothing should have updated
+        for k, v in kwargs.items():
+            assert getattr(bd, k) == v
+
+    def test_delete(self):
```
dankolbman, line 183: Is it worth testing the cascade deletes? Maybe it's already done

znatty22: Yea the test is actually in test_diagnosis_models.py [here](https://github.com/kids-first/kf-api-dataservice/pull/455/files#diff-44639382ba251660ddff1492dbf82c7eL131)

dankolbman (approved):

PR #456, opened 2018-10-11
🔥 Remove cascade delete genomic file on sequencing-experiment delete

znatty22:
When a sequencing experiment is deleted, its associated genomic files will not be deleted.


fiendish (commented):
I don't like this plan. Dataservice changes should be free to cascade. The indexd interaction that breaks portal links is the real problem.

dankolbman:
@fiendish logically, deleting a file because you delete it's sequencing experiment doesn't make much sense.

znatty22:
@fiendish I agree w @dankolbman, it really doesn't make sense to delete all of the genomic files with a sequencing experiment when a sequencing experiment is deleted. The genomic files may or may not still have valid biospecimens they're associated with. Doesn't make sense to nix them

dankolbman (approved):
lgtm

fiendish (approved):

PR #466, opened 2018-10-25
📝 Design doc - Sequencing Experiments and Downstream Genomic Files

znatty22:


docs/genomic_file/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Multi-specimen Genomic Files
+We decided that we will not have a separate model to represent multi-specimen genomic files. 
+We will keep the many-to-many relationship between biospecimens and genomic files.
+
+## Sequencing Experiments and Harmonized Files
+
+People need to be able to search for files by various sequencing experiment attributes. There are 2 options for enabling this:
+
+Option 1 - Move the relevant fields (i.e. experimental strategy, etc.) on the sequencing experiment to the genomic files.
```
znatty22: @allisonheath will check to see if this is possible

allisonheath: Ok, this is a bit trickier than I thought.

For the purposes of this change, all files I think we should carry over:
`experiment_strategy`
`platform`
`instrument_model`
`is_paired_end`

At the moment it's unlikely for these to have different values, but as with many things bioinformatics nothing is impossible. I'd initially say we should have a standard enum to indicate that there's multiple values and document how people could get the specific values based on the BS id. Probably more ideally would be the ability to store the values as arrays, but not sure how feasible that is? 

I also think we should likely move these off of SE entirely and they will exist only on the appropriate GFs and/or RGs:
`max_insert_size`
`mean_insert_size`
`mean_depth`
`total_reads`
`mean_read_length`

cc @yuankunzhu

allisonheath (changes requested):

docs/genomic_file/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+    - `experiment_strategy` (text or array)
+    - `platform` (text or array)
+    - `instrument_model` (text or array)
+    - `is_paired_end` (boolean or array)
+
+\* At the moment it's unlikely for these to have different values,
+but as with many things bioinformatics nothing is impossible. Initially, we
+should have a standard enum to indicate that there are multiple values. However,
+it would be more ideal to store the values as arrays. We need to explore the
```
dankolbman: What is the motivation of using arrays? If we want to use arrays, it would make more sense to have these fields exist in a different model with one-many relationship.

dankolbman: If these potentially have multiple values, wouldn't it imply the below have multiple values?

allisonheath: 1) Really the "raw data" from a sequencing experiment is the read-level data, that is FASTQ, BAM, CRAMs - so they make the most sense to be directly tied to a sequencing experiment 

2) Everything else, gVCF, VCF, etc is a processing level that is using that raw data to determine genomic variants/features, but they don't contain the reads themselves 

3) However, when people are looking for variant data they typically want to search and filter by whether those gVCF/VCF came from whole genome vs whole exome, and now with newer platforms/instruments like pacbio/nanopore appearing it's likely also want to distinguish between Illumina and those

4) So to meet the searchability need with the current model we could either attach sequencing experiment directly to those files, resulting in many-to-many relationships, require the portal ETL to walk through the logic, or push the fields needed for searching onto the files themselves

5) After some discussion, the last one was chosen. Traditionally things like gVCF/VCFs were typically only created with one type of experimental strategy, but increasingly there have been new methods that include multiple strategies to improve power. For example, high depth whole exome sequencing combined with whole genome for wider coverage (this is actually how the Kids First sequencing is done for tumors) as well as combining RNA-seq with DNA methods as they're all different inspections of the same underlying information. 

6) To address that potential if it was easy to utilize arrays that would help prevent unnecessary changes in the future. However, sounds like there is both some technical challenges as well as the consideration that we might want the search fields to say something like "WXS/WGS" versus bucketing them into the two categories. 

7) So if it's easier to get this PR through - it should be fine to go ahead and use it as a text field, just wanted to capture the above

docs/genomic_file/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+
+### Model Change
+
+\* Data types for fields below will remain unchanged. See SequencingExperiment
+model definition for types.
+
+
+1. Add the common SE fields to GenomicFile. Do not remove them from
+SequencingExperiment yet (we want to keep them for now)
```
dankolbman: I see. These are rollup fields with values downstream file sequencing metadata. At the very least, we should name them to indicate that. It's quite confusing otherwise.

dankolbman (commented):
This design seems like we're setting ourselves up for a lot of tedious munging tasks and possibility if data quality issues here with only a rough and unenforceable plan for how to address them.

dankolbman:
how about instead of copying these four fields from `SequentialExperiment` to `GenomicFile` and worrying about trying to keep the two consistent, we instead expose the fields from the genomic file's sequencing experiment directly on the genomic file. For right now these responses will return the value in an array, although it will only be a direct mapping from the field to the sequencing experiment field.
Later, we can turn this field into an aggregation field that queries all outputfile's sequencing experiment's and collects the available values for that field. In this way, we won't have to worry about keeping things up to date, or migrating them.

allisonheath:
@dankolbman yeah, if that's feasible that sounds like a much better direction

dankolbman:
Can we pluralize these too, so they make more sense?

    - `experiment_strategies`
    - `platforms`
    - `instrument_models`

Also not sure about the paired end field
    - `is_paired_end`

Maybe `has_paired_ends` or don't worry about it for now.

znatty22:
> we instead expose the fields from the genomic file's sequencing experiment directly on the genomic file.

Yea, I do like this idea better to avoid inconsistent data on `sequencing_experiment` and `genomic_file`. But note that the ingest will still be somewhat "complicated" because we still need to ingest the task_genomic_files so that we know which unharmonized files relate to which harmonized files. With this relation, we can find a harmonized file's associated unharmonized file(s) and from that we can find out which sequencing experiment(s) are associated. If we do this query, then it definitely makes sense to store these fields as arrays in response for genomic_file.

allisonheath:
@dankolbman could you update this with what was actually done for the SE fields and/or close this out and open a new one?

allisonheath:
So basically at the end of all of this we went with connecting the files directly back to the SE per #492. If so, can we close this out? @znatty22

PR #470, opened 2018-10-31
🐛 Update populate db

ConnorBarnhill:
- api/participant/models
  - fix typo causing AttributeError: 'AliasGroup' object has no attribute 'particpants'
- util/data_gen/data_generator
  - rename 'Workflow' to 'CavaticaTask'
  - fix randint call that was causing an index to be out of range
  - rename participant/relative to participant1/participant2
  - move consent_type from Participant to Biospecimen
  - add source_text_ prefix to tumor_location, tissue_type, anatomical_site, tumor_descriptor, diagnosis, phenotype fields
  - change volume_ml to volume_ul
  - remove SequencingExperiment query from _create_genomic_files (querying before sequencing experiments are created)
  - fix bug in _create_sequencing_centers where creating SequencingCenter twice


allisonheath:
Ended up again renaming several of the entities again, closing for now as changes are out of date, but we should fix in the future.

PR #473, opened 2018-11-12
 ✨Add biospecimen_id and genomic_file_id for genomic file and biospecimen endpoint

### LINKED ISSUE #457: Add custom filters to /biospecimens and /genomic-files
In addition to having the normal filters (any column on biospecimen_genomic_file table) on /biospecimen-genomic-files, we want to have:

- /biospecimens?genomic_file_id=foo to get back biospecimens associated with a genomic file
- /genomic-files?biospecimen_id=foo to get back genomic files associated wiht a biospecimen

parimalak:
Fixes #457

### DIFF
--- a/dataservice/api/biospecimen/resources.py
+++ b/dataservice/api/biospecimen/resources.py
@@ -6,7 +6,8 @@ from dataservice.extensions import db
 from dataservice.api.common.pagination import paginated, Pagination
 from dataservice.api.biospecimen.models import (
     Biospecimen,
-    BiospecimenDiagnosis
+    BiospecimenDiagnosis,
+    BiospecimenGenomicFile
 )
 from dataservice.api.biospecimen.schemas import (
     BiospecimenSchema,
@@ -42,6 +43,7 @@ class BiospecimenListAPI(CRUDView):
         # Get study id, diagnosis_id and remove from model filter params
         study_id = filter_params.pop('study_id', None)
         diagnosis_id = filter_params.pop('diagnosis_id', None)
+        genomic_file_id = filter_params.pop('genomic_file_id', None)
 
         # Apply filter params
         q = (Biospecimen.query
@@ -56,6 +58,10 @@ class BiospecimenListAPI(CRUDView):
         if diagnosis_id:
             q = (q.join(BiospecimenDiagnosis)
                  .filter(BiospecimenDiagnosis.diagnosis_id == diagnosis_id))
+        if genomic_file_id:
+            q = (q.join(BiospecimenGenomicFile)
+                 .filter(BiospecimenGenomicFile.genomic_file_id ==
+                 genomic_file_id))
 
         return (BiospecimenSchema(many=True)
                 .jsonify(Pagination(q, after, limit)))
--- a/dataservice/api/biospecimen/schemas.py
+++ b/dataservice/api/biospecimen/schemas.py
@@ -51,7 +51,7 @@ class BiospecimenSchema(BaseSchema):
         collection_url = 'api.biospecimens_list'
         exclude = (BaseSchema.Meta.exclude +
                    ('participant', 'sequencing_center') +
-                   ('genomic_files', 'biospecimen_genomic_files'))
+                   ('biospecimen_genomic_files', ))
 
     _links = ma.Hyperlinks({
         'self': ma.URLFor(Meta.resource_url, kf_id='<kf_id>'),
@@ -64,14 +64,21 @@ class BiospecimenSchema(BaseSchema):
         'biospecimen_diagnoses': ma.URLFor(
             'api.biospecimen_diagnoses_list', biospecimen_id='<kf_id>'),
         'diagnoses': ma.URLFor('api.diagnoses_list',
-                               biospecimen_id='<kf_id>')
+                               biospecimen_id='<kf_id>'),
+        'genomic_files': ma.URLFor('api.genomic_files_list',
+                                   biospecimen_id='<kf_id>')
     })
 
 
 class BiospecimenFilterSchema(BiospecimenSchema):
 
     diagnosis_id = fields.Str()
+    genomic_file_id = fields.Str()
 
     @validates('diagnosis_id')
     def valid_diagnosis_id(self, value):
         validate_kf_id('DG', value)
+
+    @validates('genomic_file_id')
+    def valid_genomic_file_id(self, value):
+        validate_kf_id('GF', value)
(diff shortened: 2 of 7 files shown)


dataservice/api/biospecimen/resources.py
```diff
@@ -56,6 +58,10 @@ def get(self, filter_params, after, limit):
         if diagnosis_id:
             q = (q.join(BiospecimenDiagnosis)
                  .filter(BiospecimenDiagnosis.diagnosis_id == diagnosis_id))
+        if genomic_file_id:
+            q = (q.join(BiospecimenGenomicFile)
+                 .filter(BiospecimenGenomicFile.genomic_file_id ==
+                 genomic_file_id))
```
dankolbman, line 64: Any chance of this causing duplication of the genomic files when there are multiple biospecimens that relate to it?

tests/biospecimen/test_biospecimen_resources.py
```diff
@@ -288,17 +291,24 @@ def test_filters(self):
...
+        bsgf = BiospecimenGenomicFile.query.filter_by(
+            genomic_file_id=gf.kf_id).count()
         assert bds == 2
+        assert bsgf == 2
 
         # Send get request
         filter_params = {'diagnosis_id': d.kf_id,
-                         'study_id': s.kf_id}
+                         'study_id': s.kf_id,
+                         'genomic_file_id': gf.kf_id}
```
dankolbman: It may be worth testing these independently as it difficult to tell if each is working if the same results are returned for each in isolation.

tests/biospecimen/test_biospecimen_resources.py
```diff
@@ -318,11 +328,16 @@ def test_filters(self):
         s = Study.query.filter_by(external_id='s0').first()
         bds = BiospecimenDiagnosis.query.filter_by(
             diagnosis_id=d.kf_id).count()
+        gf = GenomicFile.query.filter_by(external_id='study0-b1-gf0').first()
```
dankolbman, line 380: Same as line 302?

tests/genomic_file/test_genomic_file_resources.py
```diff
@@ -336,6 +340,22 @@ def test_filter_by_rg(client, indexd):
     for gf in gfs:
         assert gf['external_id'] in _ids
 
+    # Send get request
```
dankolbman, line 401: Might want to put this in a different test

dataservice/api/biospecimen/resources.py
znatty22: I don't think so... When you do the join between Biospecimen and BiospecimenGenomicFile the result can for sure have duplicate biospecimens or genomic files bc of the many-to-many. But because we only do this join with a filter by genomic_file_id, we will never have duplicate biospecimens in the result since we have the unique constraint on (biospecimen_id, genomic_file_id). The same is true for genomic_file/resources.py.

dataservice/api/biospecimen/schemas.py
```diff
@@ -71,7 +71,12 @@ class Meta(BaseSchema.Meta):
 class BiospecimenFilterSchema(BiospecimenSchema):
```
znatty22, line 73: I think you're missing the link in `_links` for genomic files:
```python
'genomic_files': ma.URLFor('api.genomic_files_list',
                                   biospecimen_id='<kf_id>')
```

dataservice/api/genomic_file/schemas.py
```diff
@@ -84,7 +84,12 @@ class Meta(BaseSchema.Meta, IndexdFileSchema.Meta):
 class GenomicFileFilterSchema(GenomicFileSchema):
```
znatty22, line 90: I think you're missing the link in `_links` for biospecimens:
```python
'biospecimens': ma.URLFor('api.biospecimens_list',
                                  genomic_file_id='<kf_id>')
```

tests/biospecimen/test_biospecimen_resources.py
znatty22: Yea prob a good idea for each filter to be tested independently and all together

znatty22 (changes requested):

dataservice/api/biospecimen/schemas.py
```diff
@@ -64,14 +64,21 @@ class Meta(BaseSchema.Meta):
...
+                               biospecimen_id='<kf_id>'),
+        'genomic_files': ma.URLFor('api.genomic_files_list',
+                                   biospecimen_id='<kf_id>')
     })
 
 
 class BiospecimenFilterSchema(BiospecimenSchema):
 
     diagnosis_id = fields.Str()
+    genomic_file_id = fields.Str()
```
dankolbman, line 76: What does this do? I don't think the biospecimen has a single `genomic_file_id`, does it?

dataservice/api/genomic_file/schemas.py
```diff
@@ -77,14 +77,21 @@ class Meta(BaseSchema.Meta, IndexdFileSchema.Meta):
...
+                                 genomic_file_id='<kf_id>'),
+        'biospecimens': ma.URLFor('api.biospecimens_list',
+                                  genomic_file_id='<kf_id>')
     }, description='Resource links and pagination')
 
 
 class GenomicFileFilterSchema(GenomicFileSchema):
 
     read_group_id = fields.Str()
+    biospecimen_id = fields.Str()
```
dankolbman, line 93: Where does this come from?

tests/biospecimen/test_biospecimen_resources.py
```diff
@@ -288,17 +291,73 @@ def test_filters(self):
...
+        response = json.loads(response.data.decode('utf-8'))
+        assert 2 == response['total']
+        assert 2 == len(response['results'])
+        results = response['results']
+        for r in results:
+            assert r['external_sample_id'] in {'study0-p0-b0',
+                                               'study0-p0-b1'}
+
+        # test diagnois_id filter
+        filter_params = {'diagnosis_id': d.kf_id}
```
dankolbman, line 342: 👍

dataservice/api/biospecimen/schemas.py
parimalak: This field captures what we specify while filtering in the query. ya biospecimes might not have single genomic_file and vice versa. I think we cannot provide `_links` for this reason.

dankolbman: ah I think I got confused by how code was collapsed

dankolbman (approved):

dataservice/api/biospecimen/schemas.py
znatty22: Hmm @parimalak @dankolbman ok so do you think that _links for a resource should purely reflect the relations of the underlying database entities? Idk if I like that.. Should links express how you can go from one resource to other resource(s) that are related to it. A biospecimen IS related to a genomic file, through a many to many relationship, which is implemented with the intermediate association table. So I think having a link for biospecimens, that returns biospecimens related to a genomic file is valid (and vice versa).

We should decide and standardize bc I think right now we do have _links that are of the type I described in the latter part of the above

znatty22 (approved):

PR #476, opened 2018-11-15
🎨 Make capitalization consistent for resources

dankolbman:
The swagger docs had inconsistencies where readme directories were not formatted to be equivalent to actual resources, resulting in two categories for some resources, eg: `GENOMICFILE` and `GENOMIC_FILE`.

Some of the templates were using `resource.lower()` where others just used `resource` making some endpoint inconsistent, eg `Get a GenomicFile` vs `Delete a genomicfile`

### DIFF
--- a/dataservice/api/docs/resources.py
+++ b/dataservice/api/docs/resources.py
@@ -40,9 +40,10 @@ class Swagger(View):
 
         tags = []
         for d in glob.glob('dataservice/api/*/README.md'):
-            name = d.split('/')[-2].capitalize()
+            name = d.split('/')[-2].replace('_', ' ')
+            name = name.title().replace(' ', '')
             tag = {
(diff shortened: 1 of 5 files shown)


parimalak (approved):
🥇

fiendish (approved):

dataservice/api/docs/resources.py
```diff
@@ -40,9 +40,10 @@ def dispatch_request(self):
 
         tags = []
         for d in glob.glob('dataservice/api/*/README.md'):
-            name = d.split('/')[-2].capitalize()
+            name = d.split('/')[-2].replace('_', ' ')
+            name = name.title().replace(' ', '')
```
fiendish, line 44: as long as this output is the desired one

PR #478, opened 2018-11-16
✨ Summarize sequencing experiment fields on genomic file

dankolbman:
Reflects `platforms`, `experiment_strategies`, and `platforms` of a sequencing experiment on the genomic file. In the future, these values will also include summary of all upstream file's sequencing experiments so as to allow them to be discovered when one searches for files that are the product of a specific value of them.

Mirrors `is_paired_end` from sequencing experiment onto the genomic file.

### DIFF
--- a/dataservice/api/genomic_file/models.py
+++ b/dataservice/api/genomic_file/models.py
@@ -1,3 +1,5 @@
+from sqlalchemy.ext.hybrid import hybrid_property
+
 from dataservice.extensions import db
 from dataservice.api.common.model import Base, IndexdFile, KfId
 from dataservice.api.task.models import (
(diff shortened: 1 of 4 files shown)


dataservice/api/genomic_file/models.py
```diff
@@ -68,3 +77,59 @@ class GenomicFile(db.Model, Base, IndexdFile):
...
+            return [self.sequencing_experiment.experiment_strategy]
+        else:
+            return []
+
+    @experiment_strategies.expression
+    def experiment_strategies(cls):
+        from dataservice.api.sequencing_experiment.models import (
+            SequencingExperiment
+        )
+        return [SequencingExperiment.sequencing_strategy]
```
znatty22: Should it be `[SequencingExperiment.experiment_strategy]`?

znatty22 (approved):
👍

parimalak (approved):

PR #486, opened 2018-12-12
✨ Add access_urls to indexd models to point to gen3 locations

### LINKED ISSUE #483: Add access_urls hybrid property to genomic file
Currently, the genomic file `urls` field returns the list of file locations from gen3. These are raw file locations in the case that we own the file, however, there are some files that live within another gen3 deployment.

We should unify these to all be urls pointing to gen3 deployments and return them under `access_urls` . In the current situation, this means that all urls to files in our `s3` buckets should be replaced with urls.

dankolbman:
Adds `access_urls` field to the Indexd model.
This field is a hybrid field that will simple replace any `s3://` urls with a urls to the object in our gen3, eg: `https://gen3.kidsfirstdrc.org/data/<did>`.
We will assume that any non-`s3://` is a url that already points to a valid gen3 `/data/<did>` location.
Closes #483

### DIFF
--- a/dataservice/api/common/model.py
+++ b/dataservice/api/common/model.py
@@ -109,7 +109,7 @@ class IndexdFile:
     def access_urls(self):
         """
         Access urls should contain only links out to gen3 data endpoints
-        that are used to download the file's themselves.
+        that are used to download the files themselves.
 
         For urls that are already https:// urls, we will consider them as
         valid gen3 locations, for urls that are s3:// protocol, we will assume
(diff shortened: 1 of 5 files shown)


dataservice/api/common/model.py
```diff
@@ -105,6 +105,25 @@ def constructor(self):
         # Update fields from indexd
         self.merge_indexd()
 
+    @property
+    def access_urls(self):
+        """
+        Access urls should contain only links out to gen3 data endpoints
+        that are used to download the file's themselves.
```
fiendish:
```suggestion
        that are used to download the files themselves.
```

fiendish (approved):
looks ok, though I might replace instances of the term "gen3" with "persistent link proxy" or just "proxy" for generality

parimalak (approved):

PR #492, opened 2019-01-14
♻️ Make Seq Exp and Genomic File Many to Many

### LINKED ISSUE #491: SequencingExperiment and GenomicFile Model and Endpoint Changes
See #466 for design details.

We need to be able to search for downstream genomic files (i.e. gVCF/VCF) by 
sequencing experiment information pertaining to the original harmonized 
genomic files (i.e. bams). 

After more discussion, we've determined the simplest
and more performant solution would be to make the relationship between sequencing
experiment and genomic file many to many so that we can directly link downstream
genomic files to their associated sequencing experiments.

This ticket covers the implementation of the data model change and the associated endpoints

**To do:** 
- Back out the changes that added reflected sequencing experiment fields on genomic file model
- Model change - make SE GF relationship many-to-many 
- /sequencing-experiment-genomic-files
- /genomic-files?sequencing_experiment_id=\<se kf id\>
- /sequencing-experiments?genomic_file_id=\<gf kf id\>

znatty22:
Closes #491. Discussion details here #466  

- [x] Remove reflected seq exp fields on genomic file model
- [x] Modify seq exp, gf models for m2m relation
- [x] Add `/sequencing-experiment-genomic-files` endpoint
- [x] Add filter - `/sequencing-experiments?genomic_file_id=<GF KF ID>`
- [x] Add filter - `/genomic-files?sequencing_experiment_id=<SE KF ID>`
- [x] Create new migration for model change and include population of sequencing_experiment_genomic_file table w existing data from genomic_file table
- [x] Update ERD


dataservice/api/genomic_file/resources.py
```diff
@@ -42,6 +42,10 @@ def get(self, filter_params, after, limit):
         # Get study id and remove from model filter params
         study_id = filter_params.pop('study_id', None)
 
+        # Get read group id and remove from model filter params
```
fiendish: The comment is wrong. (These pop lines shouldn't have individual comments anyway, because they're all doing the same thing.)

dataservice/api/sequencing_experiment/models.py
```diff
@@ -62,18 +62,52 @@ class SequencingExperiment(db.Model, Base):
...
-                                        'sequencing_experiment',
-                                        lazy=True))
+
+    genomic_files = association_proxy(
+        'sequencing_experiment_genomic_files',
+        'genomic_file',
+        creator=lambda gf: SequencingExperimentGenomicFile(genomic_file=gf))
+
+    sequencing_experiment_genomic_files = db.relationship(
+        'SequencingExperimentGenomicFile',
```
fiendish, line 72: All other invocations of db.relationship use a class for the first argument, not a string.

dataservice/api/sequencing_experiment/models.py
```diff
@@ -62,18 +62,52 @@ class SequencingExperiment(db.Model, Base):
...
+                                          'genomic_file_id',),)
+    sequencing_experiment_id = db.Column(
+        KfId(),
+        db.ForeignKey('sequencing_experiment.kf_id'),
+        nullable=False)
+
+    genomic_file_id = db.Column(KfId(),
+                                db.ForeignKey('genomic_file.kf_id'),
+                                nullable=False)
+    external_id = db.Column(db.Text(),
```
fiendish, line 103: Relationship between entities is an artificial construct. Under what circumstance would it ever get an external ID?

fiendish (changes requested):

dataservice/api/sequencing_experiment/models.py
znatty22: You're right its not really an `external_id` assigned by the investigator, but all of our tables now have have this field which captures either an ID assigned by the investigator or a synthetic ID made by the ingest developer. For these many-to-many association tables, I usually concatenate the external IDs from both of the related tables. Its pretty useful for sanity checking my data once ingested. 

I do think at some point we should consider rethinking external_id. Maybe we have two fields, one to capture the submitter's ID if it exists and one to capture a synthetic ID which we would use for KF ID mapping

dataservice/api/sequencing_experiment/models.py
znatty22: SqlAlchemy allows either string or a class. If its a string, sqlalchemy will convert the string to the class at runtime. Sometimes we use strings to circumvent import conflicts and/or avoid having to define classes in a strict order.  In this case it's a string bc the class definition for SequencingExperimentGenomicFile is after the the use of SequencingExperimentGenomicFile in the db.relationship call. I guess I could move the definition of the SequencingExperimentGenomicFile class up before SequencingExperiment but meh

dataservice/api/genomic_file/resources.py
znatty22: 👍

fiendish (approved):

tests/genomic_file/test_genomic_file_resources.py
```diff
@@ -156,17 +106,18 @@ def test_new_indexd_error(client, entities):
         'data_type': 'Aligned Reads',
         'file_format': 'bam',
         'urls': ['s3://bucket/key'],
-        'hashes': {'md5': 'd418219b883fce3a085b1b7f38b01e37'},
-        'sequencing_experiment_id': 'SE_AAAAAAAA',
+        # 'hashes': {'md5': 'd418219b883fce3a085b1b7f38b01e37'},
```
parimalak: ^

tests/test_pagination.py
```diff
@@ -179,6 +180,7 @@ def participants(client):
         ('/tasks', 50),
         ('/task-genomic-files', 50),
         ('/read-group-genomic-files', 50),
+        # ('/sequencing-experiment-genomic-files', 50),
```
parimalak: ^

tests/genomic_file/test_genomic_file_resources.py
znatty22: Meant to delete that line. This test is supposed to ensure that when there is a problem with the input data, the post should fail and indexd should not create a file. Hashes is a required field so without it present the post should fail as expected

tests/test_pagination.py
znatty22: Woops, will uncomment 👍

baileyckelly:
Added SBG and chop BIX to the reviewer list.

parimalak (approved):
looks good 👍

rosibaj (approved):

trbica:
This works with us (SBG)! 👍

PR #507, opened 2019-04-10
🗃 Add indices to created_at

### LINKED ISSUE #506: Investigate potential impact of new indices on created_at
Dataservice has just started maxing the RDS cpu utilization during ETL. This is abnormal because the CPU very rarely jumps to 1-3%, especially over long periods.

Screen Shot 2019-04-10 at 9 13 10 AM

I suspect this is due to the recent addition of relationship inserted on the Marazita study which increased the number of entities to be paginated during etl drastically. I suspect that we may be seeing this because we do not have an index on our `created_at` field which is used to paginate entities. Will explore further.

dankolbman:
Adds indices to the `created_at` column to enable more performant pagination.
See the below issue for benchmarks.

Fixes #506


fiendish (approved):

znatty22 (approved):
💯

parimalak (approved):

Issue #509, opened 2019-04-10
Exclude listing of foreign key from the endpoints schema


parimalak:
we could exclude

- [ ]  `sequencing_experiment_genomic_files` and `read_group_genomic_files` from genomic-files endpoint. For example,
https://kf-api-dataservice.kidsfirstdrc.org/genomic-files?biospecimen_id=BS_5J4TMCYF
- [ ] `biospecimen_diagnoses` from biospecimens endpoint

PR #511, opened 2019-04-11
✨ Exclude fkey lists from genomic-files and biospecimens results

parimalak:
Address #509 
Removes this list:
Screen Shot 2019-04-11 at 2 52 02 PM


znatty22 (changes requested):
What's the motivation for this?

parimalak:
@znatty22 When we query cohort genomic files etc. the list is large and taking up the entire space  if accessed in browser; we get the links any way using _links. and also, other tables are excluded and accessible only through links.

znatty22 (approved):
👍 

I bet this will improve performance on requests to get a page of biospecimens/genomic files

fiendish (approved):

dankolbman (approved):
Good catch. We should probably add tests for field exclusions sometime in the future to make sure we aren't doing this elsewhere.

PR #515, opened 2019-05-03
🐳 Remove venv from docker

dankolbman:
This removes references to `venv` from the `run.sh` command which seems to only have been working because there was a `venv` in the build directory. I've also added a '.dockeringore' that will prevent these unused resources from being added to the docker build context.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+.circleci/
+tests/
+venv/
+data/
+tests/
+dumps/
+pyenv/
+*/__pycache__/
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN apk update && apk add py3-psycopg2 musl-dev \
  && pip install --upgrade pip
 
 RUN         pip install -r /app/requirements.txt
-ADD         . /app
+COPY        dataservice docs .git migrations bin manage.py config.py setup.py /app/
 RUN         python /app/setup.py install
 
 EXPOSE      80
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,7 @@ RUN apk update && apk add py3-psycopg2 musl-dev \
     nginx supervisor git \
     openssl ca-certificates \
     gcc postgresql-dev \
- && pip install --upgrade pip \
- && pip install virtualenv
+ && pip install --upgrade pip
 
 RUN         pip install -r /app/requirements.txt
 ADD         . /app
--- a/bin/run.sh
+++ b/bin/run.sh
@@ -1,4 +1,3 @@
 #!/bin/ash
-source venv/bin/activate
 flask db upgrade
 supervisord -c  /etc/supervisor/conf.d/supervisord.conf


.dockerignore
```diff
@@ -0,0 +1,6 @@
+venv/
```
znatty22, line 3: Hmm you could add a few more things like .circleci, docker-compose.yml, or you could also change the Dockerfile to only include what you need in the /app dir? Maybe that would be easier

znatty22 (changes requested):

Dockerfile
```diff
@@ -7,11 +7,17 @@ RUN apk update && apk add py3-psycopg2 musl-dev \
     nginx supervisor git \
     openssl ca-certificates \
     gcc postgresql-dev \
- && pip install --upgrade pip \
- && pip install virtualenv
+ && pip install --upgrade pip
 
 RUN         pip install -r /app/requirements.txt
-ADD         . /app
+ADD         dataservice /app
```
znatty22: Wondering if you really meant to use ADD vs COPY? See https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy

You can have just one ADD/COPY statement to copy all of the files/dirs you need. This will avoid creating a bunch of unnecessary image layers. (Maybe also consider refactoring the rest of the Dockerfile to use minimal RUN steps for the same reason)

Dockerfile
```diff
@@ -7,11 +7,17 @@ RUN apk update && apk add py3-psycopg2 musl-dev \
...
+ADD         docs /app
```
znatty22: Do you need docs and .git?

.dockerignore
```diff
@@ -0,0 +1,8 @@
+.circleci/
```
znatty22, line 1: I think you can remove everything except a line for pycache now that you're only copying what you need into the working dir

.dockerignore
```diff
@@ -0,0 +1,8 @@
...
+tests/
+venv/
+data/
+tests/
+dumps/
+pyenv/
+__pycache__/
```
znatty22: Ignore all `__pycache__` dirs below root
```suggestion
*/__pycache__ 
```

znatty22 (changes requested):

Dockerfile
dankolbman: I did mean to do that, I think it should be right now.

.dockerignore
dankolbman: I could, but docker will still send everything in the working directory to the build context which will add more time to the build. There's more about the build context [here](https://docs.docker.com/v17.09/engine/reference/builder/#usage).

Dockerfile
dankolbman: The status endpoint relies on .git for the commit hash [here](https://github.com/kids-first/kf-api-dataservice/blob/master/dataservice/__init__.py#L159-L179)
The swagger docs require the docs [here](https://github.com/kids-first/kf-api-dataservice/blob/5647b38b8e13c0b6fd839be1cd179ed452ccb5b1/dataservice/api/docs/resources.py#L11-L13)

.dockerignore
znatty22: Ahh interesting ok 👍

Dockerfile
```diff
@@ -7,11 +7,10 @@ RUN apk update && apk add py3-psycopg2 musl-dev \
     nginx supervisor git \
     openssl ca-certificates \
     gcc postgresql-dev \
- && pip install --upgrade pip \
- && pip install virtualenv
+ && pip install --upgrade pip
 
 RUN         pip install -r /app/requirements.txt
-ADD         . /app
+COPY        dataservice docs .git migrations bin manage.py config.py setup.py /app
```
znatty22: Docker needs that end slash if the dest is a directory and you're copying multiple things 🤷‍♀ 

```suggestion
COPY        dataservice docs .git migrations bin manage.py config.py setup.py /app/
```

znatty22 (changes requested):

znatty22 (approved):
👍

fiendish (approved):

parimalak (approved):

PR #518, opened 2019-05-09
🐛 Individually copy files to dockerfile

dankolbman:
`COPY dir1/ dir2/ /app/` only copies the contents of the directories to the image, not the actual directories themselves.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,7 @@ RUN         pip install -r /app/requirements.txt
 
 COPY        manage.py manage.py
 COPY        setup.py setup.py
+COPY        config.py config.py
 COPY        bin bin
 COPY        docs docs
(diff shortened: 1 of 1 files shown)


znatty22 (changes requested):
Hah sorry that was my fault for telling you to put everything in one line 😳. But shouldn't jinkies have caught this and prevented the status checks from passing? If not, we prob need to fix something

Since you're already ignoring the things you don't want in the build context, maybe just go back to `COPY . /app/`? Otherwise this is going to create a lot of image layers you probably don't need

dankolbman:
> Otherwise this is going to create a lot of image layers you probably don't need

What are the issues with having more layers? As far as I know, it won't change image size significantly and, if anything, will make pulling faster because layers may be pulled async.

PR #519, opened 2019-05-13
♻️ Make Dockerfile follow best practices

znatty22:
Refactor Dockerfile to follow best practices: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

Try to minimize image layers where possible
Try to leverage build cache (things that change less should come before things that change more so that rebuilding of image layers is minimized) 

Changes reduce rebuild time by about 50%:
```
Before
time docker build -t kf-ds-2 .  0.50s user 0.33s system 5% cpu 15.665 total

After
time docker build -t kf-ds .  0.77s user 0.56s system 18% cpu 7.383 total
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,7 @@ RUN         pip install -r /app/requirements.txt
 
 COPY        manage.py manage.py
 COPY        setup.py setup.py
+COPY        config.py config.py
 COPY        bin bin
 COPY        docs docs
 COPY        migrations migrations
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,15 @@ RUN apk update && apk add py3-psycopg2 musl-dev \
  && pip install --upgrade pip
 
 RUN         pip install -r /app/requirements.txt
-COPY        dataservice docs .git migrations bin manage.py config.py setup.py /app/
+
+COPY        manage.py manage.py
+COPY        setup.py setup.py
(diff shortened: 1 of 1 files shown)


dankolbman:
So this makes it take twice as long? :p

dankolbman:
Can you just merge this against master?

Dockerfile
```diff
@@ -1,34 +1,41 @@
...
+# Setup dataservice
+COPY        requirements.txt /app/
+RUN         pip install -r /app/requirements.txt
+
+COPY        manage.py setup.py config.py /app/
+COPY        bin bin
+COPY        docs docs
+COPY        migrations migrations
+COPY        dataservice  dataservice
+COPY        .git .git
```
znatty22, line 37: @dankolbman btw, why do we need .git?

dankolbman: We return the current commit hash on the status page

znatty22: Oh thats right...

Copying the entire .git folder seems like a lot just to get the commit hash for the release. We could probably write the commit hash to the package init or somewhere so that you don't need to copy the .git folder over. Just read the commit hash from the file. But that's for another PR

dankolbman: yeah we should do that (calls to git are slow anyway)

dankolbman (approved):

fiendish (approved):

PR #522, opened 2019-05-28
🐛 Add missing groupby for sequencing center

dankolbman:
Sequencing center was missing a groupby which caused incorrect 'total' to be returned when filtering by study_id.
See for example: https://kf-api-dataservice.kidsfirstdrc.org/sequencing-centers?study_id=SD_9PYZAHHE

### DIFF
--- a/tests/test_pagination.py
+++ b/tests/test_pagination.py
@@ -195,10 +195,7 @@ class TestPagination:
         resp = json.loads(resp.data.decode('utf-8'))
         assert len(resp['results']) == min(expected_total, 10)
         assert resp['limit'] == 10
-        if '/sequencing-centers' in endpoint:
-            assert resp['total'] == 50
-        else:
-            assert resp['total'] == expected_total
+        assert resp['total'] == expected_total
 
         ids_seen = []
         # Iterate through via the `next` link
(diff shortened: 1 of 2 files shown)


parimalak (approved):

parimalak:
@dankolbman count testis failing.

tests/test_pagination.py
```diff
@@ -195,10 +195,7 @@ def test_study_filter(self, client, participants,
         resp = json.loads(resp.data.decode('utf-8'))
         assert len(resp['results']) == min(expected_total, 10)
         assert resp['limit'] == 10
-        if '/sequencing-centers' in endpoint:
```
dankolbman, line 198: I'm not really sure why this exception is here? @parimalak

znatty22: Hmm, I don't know why its there, but it looks like it shouldn't be

parimalak: we added because we can create only one sequencing center with that name

znatty22: I think also the check at line 216 should be removed. Only keep the check thats at line 219 `assert len(ids_seen) == resp['total']`

znatty22 (changes requested):

znatty22 (approved):
👍

PR #529, opened 2019-07-24
✨ Add pagination by uuid

### LINKED ISSUE #528: Dataservice paging skips entries
It looks like it might happen when multiple entries with the same creation time cross page boundaries

https://kf-api-dataservice-qa.kidsfirstdrc.org/biospecimen-genomic-files?limit=100&after=1563807032.395032&study_id=SD_BHJXBDQK

returns a set of results starting with
```
{
_links: {
biospecimen: "/biospecimens/BS_7RQCH5Y7",
collection: "/biospecimen-genomic-files",
genomic_file: "/genomic-files/GF_31NCMYC7",
self: "/biospecimen-genomic-files/BG_35XCAMP5"
},
created_at: "2019-07-22T14:50:32.397926+00:00",
external_id: "BS_7RQCH5Y7-['s3://kf-study-us-east-1-prd-sd-bhjxbdqk/harmonized/copy-number-variations/25305290-a288-468d-a19a-56ec15569d57.png']",
kf_id: "BG_35XCAMP5",
modified_at: "2019-07-22T14:50:32.397935+00:00",
visible: true
},
```
and ending with 
```
{
_links: {
biospecimen: "/biospecimens/BS_5BNV7MCF",
collection: "/biospecimen-genomic-files",
genomic_file: "/genomic-files/GF_ZC3VWCXW",
self: "/biospecimen-genomic-files/BG_8SP4P5KS"
},
created_at: "2019-07-22T14:50:33.289567+00:00",
external_id: "BS_5BNV7MCF-['s3://kf-study-us-east-1-prd-sd-bhjxbdqk/harmonized/copy-number-variations/31ecacf7-ec04-4577-999e-c348786c7792.png']",
kf_id: "BG_8SP4P5KS",
modified_at: "2019-07-22T14:50:33.289577+00:00",
visible: true
}
```

the list links are
```
{
next: "/biospecimen-genomic-files?after=1563807033.289567&study_id=SD_BHJXBDQK",
self: "/biospecimen-genomic-files?after=1563807032.397916&study_id=SD_BHJXBDQK"
},
```

If you go to the next page at https://kf-api-dataservice-qa.kidsfirstdrc.org/biospecimen-genomic-files?limit=100&after=1563807033.289567&study_id=SD_BHJXBDQK you get results starting with
```
{
_links: {
biospecimen: "/biospecimens/BS_H2K7XX4X",
collection: "/biospecimen-genomic-files",
genomic_file: "/genomic-files/GF_CY8VDW4C",
self: "/biospecimen-genomic-files/BG_K50467H0"
},
created_at: "2019-07-22T14:50:33.290550+00:00",
external_id: "BS_H2K7XX4X-['s3://kf-study-us-east-1-prd-sd-bhjxbdqk/harmonized/copy-number-variations/33071a17-1c88-464a-ac98-17b5bc1bffaf.png']",
kf_id: "BG_K50467H0",
modified_at: "2019-07-22T14:50:33.290556+00:00",
visible: true
},
```

But the next BG should be
```
{
_links: {
biospecimen: "/biospecimens/BS_68TZMZH1",
collection: "/biospecimen-genomic-files",
genomic_file: "/genomic-files/GF_5P2EY7BA",
self: "/biospecimen-genomic-files/BG_4222E7RA"
},
created_at: "2019-07-22T14:50:33.289567+00:00",
external_id: "BS_68TZMZH1-['s3://kf-study-us-east-1-prd-sd-bhjxbdqk/harmonized/copy-number-variations/32e9b9c0-1a26-43fa-8fd0-90a42a08d197.CNVs.p.value.txt']",
kf_id: "BG_4222E7RA",
modified_at: "2019-07-22T14:50:33.289577+00:00",
visible: true
}
```

You will note that BG_4222E7RA and BG_8SP4P5KS have the same creation times.

dankolbman:
This changes pagination to default to using the uuid field. The old timestamp or date format will still be supported, but pagination links will use uuids as they will ensure that all rows are returned correctly.

Closes #528

### DIFF
--- a/dataservice/api/common/pagination.py
+++ b/dataservice/api/common/pagination.py
@@ -1,48 +1,76 @@
+from typing import Optional, Tuple
+from uuid import UUID
 from flask import request, current_app
 from functools import wraps
 from dateutil import parser
 from datetime import datetime
+from sqlalchemy import and_, or_
+
+
+After = Tuple[Optional[datetime], Optional[str]]
 
 
 def paginated(f):
 
     @wraps(f)
     def paginated_wrapper(*args, **kwargs):
+        """
+        Parses details about a page being requested by its urls parameters
+        and injects them into the wrapped function's kwargs
+
+        Handles parameters of the form:
+        ?after=1529089066.003078&after_uuid=e29fba44-6d39-4719-b600-97aadbe876a0&limit=10
+
+        Where the ?after parameter is a either a timestamp or parseable
+        datetime (as determined by the dateutil module).
+        The ?after_uuid parameter is the uuid used to resolve any conflicts
+        between rows with the same created_at datetime.
+        The ?limit parameter specifies how many results to return on a page
+        that occur after the specified ?after and ?after_uuid parameters
+        """
         def_limit = current_app.config['DEFAULT_PAGE_LIMIT']
         max_limit = current_app.config['MAX_PAGE_LIMIT']
         limit = min(request.args.get('limit', def_limit, type=int), max_limit)
-        after = request.args.get('after', None)
+        after_date = request.args.get('after', '')
+        after_uuid = request.args.get('after_uuid', None)
 
-        if type(after) is str:
-            # Parser won't recognize the timestamp with fractional seconds
-            if after.replace('.', '').isdigit():
-                after = float(after)
-                after = datetime.fromtimestamp(after)
-            else:
-                try:
-                    after = parser.parse(after)
-                # Parser couldn't derive a datetime from the string
-                except ValueError:
-                    after = None
+        # Assume timestamp if ?after is a number
+        if after_date.replace('.', '').isdigit():
+            after_date = float(after_date)
+            after_date = datetime.fromtimestamp(after_date)
+        # Otherwise, try to extract a datetime with dateutil parser
+        else:
+            try:
+                after_date = parser.parse(after_date)
+            # Parser couldn't derive a datetime from the string
+            except ValueError:
+                # Fallback to begining of the epoch if we can't parse
+                after_date = datetime.fromtimestamp(0)
 
-        # Default to the unix epoch
-        if after is None:
-            after = datetime.fromtimestamp(0)
+        # Try to parse a valid UUID, if there is one
+        if after_uuid is not None:
+            try:
+                after_uuid = str(UUID(after_uuid))
+            except ValueError:
+                after_uuid = None
 
-        return f(*args, **kwargs,
-                 after=after,
-                 limit=limit)
+        # Default the uuid to the zero uuid if none is specified
+        if after_uuid is None:
+            after_uuid = str(UUID(int=0))
+
+        after = (after_date, after_uuid)
+        return f(*args, **kwargs, after=after, limit=limit)
 
     return paginated_wrapper
 
 
 class Pagination(object):
     """
-    Object to help paginate through endpoints using the created_at field
+    Object to help paginate through endpoints using the created_at field and
+    uuid fields
     """
 
-    def __init__(self, query, after, limit):
-        assert type(after) is datetime
+    def __init__(self, query: str, after: After, limit: int):
         self.query = query
         self.after = after
         self.limit = limit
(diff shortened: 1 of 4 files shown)


dataservice/api/common/pagination.py
```diff
@@ -22,71 +23,90 @@ def paginated_wrapper(*args, **kwargs):
...
+            query.order_by(order_by_field.asc())
+            .filter(order_by_field > after)
+            .limit(limit)
+            .all()
+        )
 
     @property
     def prev_num(self):
-        """ Returns the timestamp of the first item """
+        """ Returns the uuid of the first item """
```
dankolbman: This function isn't used right now and could probably be removed until we support a `before` parameter.

dataservice/api/common/pagination.py
```diff
@@ -22,71 +23,90 @@ def paginated_wrapper(*args, **kwargs):
...
+            return str(self.items[0].uuid)
+        return str(UUID(int=0))
 
     @property
     def curr_num(self):
         """
-        Returns the timestamp of the first item minus some epsilon
-        so that the current timestamp will be included in the range
+        Returns the first item's UUID minus one so that the resulting param
+        will include the first item in it's query.
```
fiendish:
```suggestion
        will include the first item in its query.
```

fiendish (dismissed):
This will make iterating pages miss anything that was added after iteration began. Can we keep sorting by creation date?

dataservice/api/common/schemas.py
```diff
@@ -68,7 +68,7 @@ def wrap_envelope(self, data, many):
             _links = {}
 
             # If an 'after' param could not be parsed, don't include the param
-            after = None if p.after.timestamp() == 0 else p.curr_num
+            after = p.curr_num
```
fiendish: The comment on this line seems no longer valid

dankolbman:
@fiendish Right now the problem is that `created_at` only gives us so much precision so we can't reliably use it to reference pages. What do you mean by only *presenting* the UUID?

We could possibly get the best of both worlds by using a composite index where we sort by `(created_at, uuid)`

fiendish:
Like put UUID of the next item in the next link and read UUID in the after field, just don't sort by UUID.

fiendish:
I mean like first assign a row number (e.g. with row_number()) for all results sorted by creation date, find the row number of the given UUID, and then filter starting from the row number of the given UUID. That way you have date-ordered entries starting from a given UUID and you don't miss any and you _also_ don't insert anything in the middle of the already-paged region.

fiendish:
Like the `Using ROW_NUMBER() function for pagination` section on http://www.postgresqltutorial.com/postgresql-row_number/

fiendish:
> just don't sort by UUID

Actually, sort by creation date _and_ UUID such that creation date determines the main order and then sub-sort any entries that have the same creation date by UUID so that all entries have a fully defined order.

dankolbman:
I think we'll have to sort by some other key. I'm not entirely sure if only sorting by created_at will have deterministic order, so may not reliably be able to provide the uuid for the next item.
row number needs to enumerate the full result set so it's not going to be great for large queries. 

Perhaps we could limit it by first filtering the created_at date, but then the possible non-deterministic sorting of identical dates may cause issues.

fiendish:
> I think we'll have to sort by some other key.

I ninjaed you. See ^

dankolbman:
> > just don't sort by UUID
> 
> Actually, sort by creation date _and_ UUID such that creation date determines the main order and then sub-sort any entries that have the same creation date by UUID so that all entries have a defined order.

Yeah this is what I meant by the `(created_at, uuid)` sorting.

So should we accept this composite index in a single param, or split into multiple?
`/studies?after=1529089066.003078,e29fba44-6d39-4719-b600-97aadbe876a0`
vs
`/studies?after=1529089066.003078&uuid_after=e29fba44-6d39-4719-b600-97aadbe876a0`

The first way would be easier as we wouldn't need to modify every view we have.

fiendish:
I like just `?after=e29fba44-6d39-4719-b600-97aadbe876a0` like you currently have, since every uuid already implies a timestamp.

fiendish:
> row number needs to enumerate the full result set

Sorting needs to do that anyway. The query optimizer might keep it from happening twice.

dankolbman:
> > row number needs to enumerate the full result set
> 
> Sorting needs to do that anyway.

Not if it's indexed

dankolbman:
> I would like just `?after=e29fba44-6d39-4719-b600-97aadbe876a0`

I think we can only do this if
1) we only want to sort by uuid
or
2) we only provide uuid's of existing rows so we can look up their created_at and sort by it behind the scenes

fiendish:
I was thinking the second one, but I guess the url format doesnt really matter and `after=1529089066.003078,e29fba44-6d39-4719-b600-97aadbe876a0` is just as good _as long as_ you can use just a date if you don't know uuids.

Also consider changing `after=` to `from=` (inclusive start instead of exclusive start). It won't drop things when starting with only a timestamp.

dankolbman:
I went with the additional `after_uuid` parameter to resolve potential timestamp conflicts instead of the concatenated key as I realized that the `,` will be encoded for the url.
The `after` parameter should still function the same as before, but links will now be returned with the additional `after_uuid` fields

dataservice/api/common/pagination.py
```diff
@@ -1,92 +1,122 @@
...
-        return self._to_timestamp(datetime.utcnow())
+    def curr_num(self) -> Optional[After]:
+        """ Returns the after index of the current page """
+        return self.after
 
     @property
-    def next_num(self):
-        """ Returns the timestamp of the last item"""
+    def next_num(self) -> Optional[After]:
+        """ Returns the timestamp, uuid of the last item"""
```
fiendish, line 111: s/timestamp/datetime?

fiendish (approved):
seems ok

znatty22 (approved):
👍

PR #539, opened 2019-09-25
✨ Add Indexd authz field

### LINKED ISSUE #536: Pass through new indexd authz field
With the addition of the `authz` field in indexd, we will need to start passing this additional field to indexd for any update or creation actions and return it as part of any get.
We may also want to have functionality to merge previous acls with the new authz format.
See the [indexd docs](https://github.com/uc-cdis/indexd).

dankolbman:
## About

The `acl` field is being replaced with a new `authz` field which will correspond to policies created in [Arborist](https://github.com/uc-cdis/arborist). These policies will be created automatically according to our `users.yml` and `dbGaP` syncing, so we will not have to worry about creating them. There is no validation on them, either, so we can populate them regardless of whether they reference legitimate policies or not.

Moving fhe `acl` field is intended to be used as a fallback if a record has no `authz` field.
Because access is currently determined by the consuming application, it is possible that we may have a divergence in some services that continue to check the `acl` field even after the `authz` field has been added. Therefore, we should manage the `acl` as conservatively as possible. Locking down the `acl` to remain empty is by far the safest solution.

## Migrating

Gen3 Indexd has a migration script that will populate `authz` fields based on the `acl` field. This script will be run to update everything to `authz` policies.
After this has occurred, we will be able to manually set all record's `acl` to be empty to ensure that it is not used when determining access to the record.

### New `authz` Field

Adds a pass-through `authz` field for Indexd documents similar to other fields that are relayed through dataservice to/from indexd.
This field is not persisted inside of the dataservice, but proxied to and from Indexd where it is stored.

### Disable `acl` Field for Records

An error will be returned when a record is created or update with a new or non-empty `acl` field.
This will force any new data to be created using the new `authz` field instead of using the soon to be deprecated `acl` field.

```
$ curl -s -XPOST localhost:5000/genomic-files -d '{"acl": ["abc", "123"], "urls": ["s3://abc/abc"], "hashes": {"md5":"e4b246e5bac1f152b3b13047a4876b84"}, "size": 10}' | jq
{
  "_status": {
    "code": 400,
    "message": "could not create genomic_file: {'acl': ['The ACL field has been disabled. Use the authz field instead.']}"
  }
}
```

**NB: This will deprecate creating/updating the acl field and fail requests that specify one. Updates to existing documents that do not wish to change the acl should remove it from the body.**

Fixes  #536

### DIFF
--- a/dataservice/api/common/schemas.py
+++ b/dataservice/api/common/schemas.py
@@ -111,10 +111,19 @@ class BaseSchema(ma.ModelSchema):
             raise ValidationError('Unknown field', unknown)
 
 
+def acl_deprecation(v):
+    if v is not None and not v == []:
+        raise ValidationError(
+            "The ACL field has been disabled. Use the authz field instead."
+        )
+
+
 class IndexdFileSchema(Schema):
     urls = ma.List(ma.Str(), required=True)
     access_urls = ma.List(ma.Str(), dump_only=True)
-    acl = ma.List(ma.Str(), required=False)
+    acl = ma.List(
+        ma.Str(), required=False, validate=lambda v: acl_deprecation(v)
+    )
     authz = ma.List(ma.Str(), required=False)
     file_name = ma.Str()
     hashes = ma.Dict(required=True)
--- a/dataservice/api/common/schemas.py
+++ b/dataservice/api/common/schemas.py
@@ -115,6 +115,7 @@ class IndexdFileSchema(Schema):
     urls = ma.List(ma.Str(), required=True)
     access_urls = ma.List(ma.Str(), dump_only=True)
     acl = ma.List(ma.Str(), required=False)
+    authz = ma.List(ma.Str(), required=False)
     file_name = ma.Str()
     hashes = ma.Dict(required=True)
     metadata = ma.Dict(attribute='_metadata')
--- a/dataservice/extensions/flask_indexd.py
+++ b/dataservice/extensions/flask_indexd.py
@@ -3,7 +3,7 @@ import uuid
 
 from flask import current_app, abort
 from flask import _app_ctx_stack as stack
-from requests.exceptions import HTTPError
+from requests.exceptions import HTTPError, RequestException
 
 
 class RecordNotFound(HTTPError):
@@ -55,9 +55,12 @@ class Indexd(object):
             return record
 
         url = self.url + record.latest_did
-        resp = self.session.get(url)
-        self.check_response(resp)
-        resp.raise_for_status()
+        try:
+            resp = self.session.get(url)
+            self.check_response(resp)
+            resp.raise_for_status()
+        except RequestException as err:
+            abort(500, f"Problem getting record from Indexd")
 
         # update fields on the target record's object
         for prop, v in resp.json().items():
@@ -121,10 +124,12 @@ class Indexd(object):
         }
 
         # Register the file on indexd
-        resp = self.session.post(self.url,
-                                 json=req_body)
+        try:
+            resp = self.session.post(self.url, json=req_body)
+            resp.raise_for_status()
+        except RequestException as err:
+            abort(500, f"Problem saving file to Indexd")
 
-        resp.raise_for_status()
         resp = resp.json()
 
         # Update the record object with the id fields
@@ -146,7 +151,10 @@ class Indexd(object):
 
         # Fetch rev for the did
         url = self.url + record.latest_did
-        resp = self.session.get(url)
+        try:
+            resp = self.session.get(url)
+        except RequestException as err:
+            abort(500, "Problem getting record from Indexd")
         self.check_response(resp)
 
         old = resp.json()
@@ -174,12 +182,19 @@ class Indexd(object):
         if 'size' in req_body or 'hashes' in req_body:
             # Create a new version in indxed
             req_body['form'] = 'object'
-            resp = self.session.post(url, json=req_body)
-            did = resp.json()['did']
+            try:
+                resp = self.session.post(url, json=req_body)
+            except RequestException as err:
+                abort(500, "Problem creating a new record in Indexd")
+
+            did = resp.json()["did"]
             record.latest_did = did
         else:
             # Update the file on indexd
-            resp = self.session.put(url, json=req_body)
+            try:
+                resp = self.session.put(url, json=req_body)
+            except RequestException as err:
+                abort(500, "Problem updating record in Indexd")
 
         self.check_response(resp)
         resp.raise_for_status()
@@ -196,7 +211,10 @@ class Indexd(object):
                   'metadata', 'urls_metadata', 'rev']
 
         url = '{}{}/versions'.format(self.url, record.latest_did)
-        versions = self.session.get(url).json()
+        try:
+            versions = self.session.get(url).json()
+        except RequestException as err:
+            abort(500, "Problem updating record in Indexd")
         for version, doc in versions.items():
             if doc['acl'] != record.acl:
                 did = doc['did']
@@ -207,7 +225,10 @@ class Indexd(object):
                 url = '{}{}?rev={}'.format(self.url, did, doc['rev'])
                 # rev is not allowed in put schema
                 del doc['rev']
-                resp = self.session.put(url, json=doc)
+                try:
+                    resp = self.session.put(url, json=doc)
+                except RequestException as err:
+                    abort(500, "Problem updating record in Indexd")
                 # Update the record's rev if it's the record being modified
                 if record.latest_did == did:
                     record.rev = resp.json()['rev']
@@ -224,13 +245,16 @@ class Indexd(object):
             return record
 
         if record.rev is None:
-            r = self.session.get(self.url + record.latest_did)
+            try:
+                r = self.session.get(self.url + record.latest_did)
+            except RequestException as err:
+                abort(500, "Problem getting record in Indexd")
             record.rev = r.json()['rev']
 
         url = '{}{}?rev={}'.format(self.url, record.latest_did, record.rev)
-        resp = self.session.delete(url)
-        self.check_response(resp)
         try:
+            resp = self.session.delete(url)
+            self.check_response(resp)
             resp.raise_for_status()
         except HTTPError:
             message = 'could not get file record'
@@ -238,6 +262,8 @@ class Indexd(object):
             if 'error' in resp.json():
                 message = '{}: {}'.format(message, resp.json()['error'])
             abort(resp.status_code, message)
+        except RequestException as err:
+            abort(500, "Problem deleting record in Indexd")
 
         return record
 
--- a/dataservice/extensions/flask_indexd.py
+++ b/dataservice/extensions/flask_indexd.py
@@ -59,6 +59,8 @@ class Indexd(object):
             resp = self.session.get(url)
             self.check_response(resp)
             resp.raise_for_status()
+        except RecordNotFound as err:
+            raise err
         except RequestException as err:
             abort(500, f"Problem getting record from Indexd")
 
@@ -80,7 +82,8 @@ class Indexd(object):
 
           {
             "file_name": "my_file",
-            "acl": ["phs000000"],
+            "acl": [],
+            "authz": ["/programs/phs000000"],
             "hashes": {
               "md5": "0b7940593044dff8e74380476b2b27a9"
             },
@@ -119,6 +122,7 @@ class Indexd(object):
             "form": "object",
             "hashes": record.hashes,
             "acl": record.acl,
+            "authz": record.authz,
             "urls": record.urls,
             "metadata": meta
         }
@@ -165,6 +169,7 @@ class Indexd(object):
             "size": record.size,
             "hashes": record.hashes,
             "acl": record.acl,
+            "authz": record.authz,
             "urls": record.urls,
             "metadata": record._metadata
         }
@@ -174,8 +179,9 @@ class Indexd(object):
             del req_body['size']
             del req_body['hashes']
 
-        # If acl changed, update all previous version with new acl
-        if record.acl != old['acl']:
+        # Update all previous versions with the new authz or acl fields, if
+        # they changed
+        if record.authz != old.get('authz') or record.acl != old.get('acl'):
             self._update_all_acls(record)
 
         url = '{}{}?rev={}'.format(self.url, record.latest_did, record.rev)
@@ -203,11 +209,11 @@ class Indexd(object):
 
     def _update_all_acls(self, record):
         """
-        Update acls for all previous versions of a record and update the
-        target record's rev
+        Update acls and authz for all previous versions of a record and
+        update the target record's rev
         """
         # Only use fields allowed by the indexd PUT schema
-        fields = ['urls', 'acl', 'file_name', 'version',
+        fields = ['urls', 'acl', 'authz', 'file_name', 'version',
                   'metadata', 'urls_metadata', 'rev']
 
         url = '{}{}/versions'.format(self.url, record.latest_did)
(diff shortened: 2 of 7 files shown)


fiendish:
If the authz just prepends acls, why disable the acl field?

dankolbman:
The prepending is just the behavior of the initial migration, there's no guarantees that this will always be true, like when we will have to make use of the hierarchical policy structure in the future.

znatty22:
> The migration takes care of adding policies here: https://github.com/uc-cdis/indexd/blob/master/bin/migrate_acl_authz.py#L227-L231

Hmm, this code looks like it only creates resources in Arborist if they don't exist. I don't see anything that actually creates the roles and ties them to resources (a.k.a the policies). I know this doesn't really affect this PR but I'd like to know how that happens/how we plan on implementing that

dataservice/extensions/flask_indexd.py
```diff
@@ -55,9 +55,14 @@ def get(self, record):
...
         url = self.url + record.latest_did
-        resp = self.session.get(url)
-        self.check_response(resp)
-        resp.raise_for_status()
+        try:
+            resp = self.session.get(url)
+            self.check_response(resp)
+            resp.raise_for_status()
+        except RecordNotFound as err:
+            raise err
```
znatty22, line 63: Why raise an exception here instead of calling `abort`? 

And if you really do want to raise the exception then you could prob just call `check_response` outside the try except

dataservice/extensions/flask_indexd.py
```diff
@@ -116,15 +121,18 @@ def new(self, record):
             "form": "object",
             "hashes": record.hashes,
             "acl": record.acl,
+            "authz": record.authz,
```
znatty22, line 125: Maybe update docstring for this method to include authz

dataservice/extensions/flask_indexd.py
```diff
@@ -116,15 +121,18 @@ def new(self, record):
...
         }
 
         # Register the file on indexd
-        resp = self.session.post(self.url,
-                                 json=req_body)
+        try:
+            resp = self.session.post(self.url, json=req_body)
+            resp.raise_for_status()
+        except RequestException as err:
+            abort(500, f"Problem saving file to Indexd")
```
znatty22:
```suggestion
            abort(500, f"Problem creating new record in Indexd")
```

dataservice/extensions/flask_indexd.py
```diff
@@ -146,7 +154,10 @@ def update(self, record):
 
         # Fetch rev for the did
         url = self.url + record.latest_did
-        resp = self.session.get(url)
+        try:
+            resp = self.session.get(url)
```
znatty22, line 159: Do you need a `resp.raise_for_status()` here?

dataservice/extensions/flask_indexd.py
```diff
@@ -166,20 +178,34 @@ def update(self, record):
...
 
-        # If acl changed, update all previous version with new acl
+        # Updates to the acl have been disabled in favor of the authz field
         if record.acl != old['acl']:
+            abort(
+                400,
+                "Changes to the ACL field have been disabled. "
+                "Use the authz field instead.",
+            )
+        # Update all previous versions with the new auhtz field, if it changed
```
znatty22:
```suggestion
        # Update all previous versions with the new authz field, if it changed
```

dataservice/extensions/flask_indexd.py
```diff
@@ -166,20 +178,34 @@ def update(self, record):
...
             self._update_all_acls(record)
 
         url = '{}{}?rev={}'.format(self.url, record.latest_did, record.rev)
         if 'size' in req_body or 'hashes' in req_body:
             # Create a new version in indxed
             req_body['form'] = 'object'
-            resp = self.session.post(url, json=req_body)
-            did = resp.json()['did']
+            try:
+                resp = self.session.post(url, json=req_body)
```
znatty22, line 193: Do you need a `resp.raise_for_status()` here?

dataservice/extensions/flask_indexd.py
```diff
@@ -166,20 +178,34 @@ def update(self, record):
...
+            except RequestException as err:
+                abort(500, "Problem creating a new record in Indexd")
+
+            did = resp.json()["did"]
             record.latest_did = did
         else:
             # Update the file on indexd
-            resp = self.session.put(url, json=req_body)
+            try:
+                resp = self.session.put(url, json=req_body)
```
znatty22, line 203: Do you need a resp.raise_for_status() here?

dataservice/extensions/flask_indexd.py
```diff
@@ -192,22 +218,29 @@ def _update_all_acls(self, record):
...
         """
         # Only use fields allowed by the indexd PUT schema
-        fields = ['urls', 'acl', 'file_name', 'version',
+        fields = ['urls', 'acl', 'authz', 'file_name', 'version',
                   'metadata', 'urls_metadata', 'rev']
 
         url = '{}{}/versions'.format(self.url, record.latest_did)
-        versions = self.session.get(url).json()
+        try:
+            versions = self.session.get(url).json()
```
znatty22: Do you need a `resp.raise_for_status()` here?

dataservice/extensions/flask_indexd.py
```diff
@@ -166,20 +178,34 @@ def update(self, record):
             del req_body['size']
             del req_body['hashes']
 
-        # If acl changed, update all previous version with new acl
+        # Updates to the acl have been disabled in favor of the authz field
         if record.acl != old['acl']:
```
znatty22: Will `acl` always be present in the response or are they planning to remove it completely at some point? If the latter, should we change to `old.get('acl')`

dataservice/extensions/flask_indexd.py
```diff
@@ -192,22 +218,29 @@ def _update_all_acls(self, record):
...
 
         url = '{}{}/versions'.format(self.url, record.latest_did)
-        versions = self.session.get(url).json()
+        try:
+            versions = self.session.get(url).json()
+        except RequestException as err:
+            abort(500, "Problem updating record in Indexd")
         for version, doc in versions.items():
-            if doc['acl'] != record.acl:
+            if doc['acl'] != record.acl or doc['authz'] != record.authz:
```
znatty22, line 230: Will `acl` always be present in the response or are they planning to remove it completely at some point? If the latter, should we change to `old.get('acl')`

dataservice/extensions/flask_indexd.py
```diff
@@ -192,22 +218,29 @@ def _update_all_acls(self, record):
...
                 doc['acl'] = record.acl
+                doc['authz'] = record.authz
                 if doc['version'] is None:
                     del doc['version']
                 url = '{}{}?rev={}'.format(self.url, did, doc['rev'])
                 # rev is not allowed in put schema
                 del doc['rev']
-                resp = self.session.put(url, json=doc)
+                try:
+                    resp = self.session.put(url, json=doc)
```
znatty22, line 241: Do you need a `resp.raise_for_status()` here?

dataservice/extensions/flask_indexd.py
```diff
@@ -224,20 +257,25 @@ def delete(self, record):
             return record
 
         if record.rev is None:
-            r = self.session.get(self.url + record.latest_did)
+            try:
+                r = self.session.get(self.url + record.latest_did)
```
znatty22, line 262: Do you need a `resp.raise_for_status()` here? Also do you need a `check_response` call after the try/except?

dataservice/extensions/flask_indexd.py
```diff
@@ -224,20 +257,25 @@ def delete(self, record):
...
-        resp = self.session.delete(url)
-        self.check_response(resp)
         try:
+            resp = self.session.delete(url)
+            self.check_response(resp)
             resp.raise_for_status()
         except HTTPError:
             message = 'could not get file record'
             # Attach indexd error message, if there is one
             if 'error' in resp.json():
```
znatty22, line 275: Hmm why do we only attach the indexd error on delete errors? Seems like a good thing to do for all requests. 

I think its worth considering a common `send_request` method that can be called everywhere you send a request and do exception/error handling. It could take in the requests HTTP method, some request kwargs, and also an optional error message string to return if aborting.

dataservice/extensions/flask_indexd.py
znatty22: Also won't `record.acl` always be None now (since `acl_deprecation` validation would have raised an exception if it wasn't)?

znatty22 (changes requested):

dataservice/extensions/flask_indexd.py
dankolbman: There's some round-about logic that relies on the specific `RecordNotFound` exception when trying to resolve documents that no longer exist:
https://github.com/kids-first/kf-api-dataservice/blob/b81e68648ec9d1a880deb1e6970862a6e46199e5/dataservice/api/common/model.py#L136-L140

dankolbman: I could move the `check_response()` call outside of the try/except block, but then I will need two try/excepts for both the `check_response()` and `raise_for_status()`.

dataservice/extensions/flask_indexd.py
```diff
@@ -55,9 +55,14 @@ def get(self, record):
...
-        self.check_response(resp)
-        resp.raise_for_status()
+        try:
+            resp = self.session.get(url)
+            self.check_response(resp)
+            resp.raise_for_status()
+        except RecordNotFound as err:
+            raise err
+        except RequestException as err:
+            abort(500, f"Problem getting record from Indexd")
```
fiendish: non-specific. what happens to err?

dataservice/extensions/flask_indexd.py
znatty22: > There's some round-about logic that relies on the specific RecordNotFound exception when trying to resolve documents that no longer exist

Ohhh ok, I forgot about that 👍

znatty22: > I could move the check_response() call outside of the try/except block, but then I will need two try/excepts for both the check_response() and raise_for_status().

Why? `check_response` just raises RecordNotFound which you're just re-raising when you catch it right?

dankolbman: yeah, you're right

dankolbman: Actually, I would still need to try/catch blocks because both `raise_for_status()` and `get()` may raise `RequestException`.

dankolbman: I will just get rid of the `RecordNotFound` handler here.

dataservice/extensions/flask_indexd.py
dankolbman: This is true, I believe I can remove this logic here because the schema validation will always catch when it's not None.

dataservice/extensions/flask_indexd.py
```diff
@@ -78,6 +83,7 @@ def new(self, record):
           {
             "file_name": "my_file",
             "acl": ["phs000000"],
```
znatty22: Shouldn't this be: 
```suggestion
            "acl": None,
```

dataservice/extensions/flask_indexd.py
```diff
@@ -166,20 +180,31 @@ def update(self, record):
             del req_body['size']
             del req_body['hashes']
 
-        # If acl changed, update all previous version with new acl
-        if record.acl != old['acl']:
+        # Update all previous versions with the new authz or acl fields, if
+        # they changed
+        if record.authz != old.get('authz') or (
+            old.get('acl') is not None and record.acl != old.get('acl')
```
znatty22: Do you need this condition? Bc now you'll never be able to create or update the `acl` attribute right (since every request will pass through the acl deprecation validator)? `record.acl` will always be None.

dataservice/extensions/flask_indexd.py
```diff
@@ -166,20 +180,31 @@ def update(self, record):
...
             self._update_all_acls(record)
 
         url = '{}{}?rev={}'.format(self.url, record.latest_did, record.rev)
         if 'size' in req_body or 'hashes' in req_body:
             # Create a new version in indxed
             req_body['form'] = 'object'
-            resp = self.session.post(url, json=req_body)
-            did = resp.json()['did']
+            try:
+                resp = self.session.post(url, json=req_body)
```
znatty22, line 193: Do you want a `resp.raise_for_status()` after this line?

dataservice/extensions/flask_indexd.py
```diff
@@ -224,20 +259,25 @@ def delete(self, record):
...
             record.rev = r.json()['rev']
 
         url = '{}{}?rev={}'.format(self.url, record.latest_did, record.rev)
-        resp = self.session.delete(url)
-        self.check_response(resp)
         try:
+            resp = self.session.delete(url)
+            self.check_response(resp)
             resp.raise_for_status()
         except HTTPError:
```
znatty22, line 272: Same question as before...why do we only attach the indexd error on delete errors? Seems like a good thing to do for all requests.

znatty22 (changes requested):

dankolbman:
In light of some new information on how SBG is handling permissions, I'm going to icebox this until I've sketched out a design document outlining how we are going to move foreword with assigning ACLs and authz fields.

dankolbman:
The `acl` -> `authz` update is intended to be backwards-compatible wherein documents that do not have a `authz` field set will fall back to the `acl` field when checking permissions.
However, because this depends on the client service to v

# Release 1

First release will prevent new documents from being created with anything but an empty `acl` field. This will ensure that any new documents will not depend on the acl field.

# Release 2

Second release will disable changes and updates to the `acl` field, migrate `acl` to `authz`, and set all `acl` fields to be empty.
The flow will be:
1. Run Gen3 acl migration script to populate `authz` field
2. Set all `acl` fields to `[]`
3. Deploy release to disable updates to the `acl` field

dataservice/extensions/flask_indexd.py
dankolbman: The acl deprecation validator will allow _only_ `[]` or `None` values for the `acl` field. This will allow a user to update an `acl` from _something_ to `[]` or `None`. A user will not be able to change it to anything else, however.
In the case that this update does happen, we will need to patch all the old versions of the document with the new empty acl.

dankolbman:
@znatty22 To address the policy creation question, I've been told that all policies, resources, groups, etc will be created from the 'user sync' cron that is responsible for parsing `users.yml` and dbGaP telemetry to create those entities.

See the updated description for migration plan.

dataservice/api/common/schemas.py
```diff
@@ -111,10 +111,20 @@ def check_unknown_fields(self, data, original_data):
             raise ValidationError('Unknown field', unknown)
 
 
+def acl_deprecation(v):
+    if v is not None and not v == []:
```
znatty22, line 115: Couldn't this just be `if v`?

znatty22 (changes requested):

dataservice/api/common/schemas.py
dankolbman: What if `v = False`?

fiendish: sure, or "", but can v be those things?

dankolbman: v is only allowed to be `[]` or undefined (`None`)

fiendish: before or after this point? if before then if v would be ok :)

I meant _can_ be not _allowed_ _to_ be. Like, whether it's possible for a v reaching this function to be not a list or none

dankolbman: `v` comes from the request body that the user sends us, so it _can_ be anything.
This function defines the constraint that we want to enforce (`v ∈ ([], None)`)

znatty22: Is `if v is not None and not v == []` = `not (v is None or v is [])` right?....De morgans or whatever that boolean negation law is lol? If yes, then I think this PR is ok

znatty22 (approved):
✅

dataservice/extensions/flask_indexd.py
```diff
@@ -224,20 +258,25 @@ def delete(self, record):
...
             message = 'could not get file record'
             # Attach indexd error message, if there is one
             if 'error' in resp.json():
                 message = '{}: {}'.format(message, resp.json()['error'])
             abort(resp.status_code, message)
+        except RequestException as err:
+            abort(500, "Problem deleting record in Indexd")
```
fiendish: other invocations like this include {err}

fiendish (changes requested):

fiendish (approved):

Issue #542, opened 2019-10-10
New field to distinguish human and canine participants


allisonheath:
We have canine data incoming, so participant needs a new field that takes two values:
`Homo sapiens` (default)
`Canis lupus familiaris` (domesticated dog)

My initial thought was to call this `species`, but really the latter is a subspecies, so was considering calling this field `taxonomy`? Or any other good options for this?

cc @nicolevasilevsky perhaps?

nicolevasilevsky:
+1 to taxonomy. We use NCBI taxonomy for organism annotations.

allisonheath:
Ok great @znatty22 let's go with `taxonomy`. 

@nicolevasilevsky as in https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=9615 right?

nicolevasilevsky:
Yes, exactly!

znatty22:
Completed

Issue #541, opened 2019-10-09
Add field to capture the DUO code to Biospecimen


allisonheath:
The value is an ontology code defined: https://github.com/EBISPOT/DUO

Not sure if we want it to be `duo_id` or `duo_id_biospecimen` to match other ontology fields?

znatty22:
@allisonheath Should this field be the DUO ID (e.g. DUO:0000021) or the DUO shorthand code (e.g. IRB)? See: https://github.com/EBISPOT/DUO/blob/master/src/ontology/duo.csv

znatty22:
Also, is this field replacing the `dbgap_consent_code` field we already have? See https://github.com/kids-first/kf-api-dataservice/blob/5d52f76e94097528360a08e99230a04d68bc865d/dataservice/api/biospecimen/models.py#L109

baileyckelly:
I dont think the dbgap consent code would go away (at least not right now). That is where we store the code that we get directly from dbgap xml when we run the ACL tool. The dbgap consent codes map to the Consent Types - which can be specific to the study. However, the dbgap consent types would map to a more general DUO code perhaps?

Take the Marazita Orofacial cleft study. They have the following consent types:

- Health/Medical/Biomedical (MDS)
-- Consent Type Code: HMB-MDS
-- dbgap Consent Code: c4
-- Potential Duo Code: DUO:0000006 / HMB
- Disease-Specific (Orofacial birth defects and related phenotypes, MDS)
-- Consent Type Code: DS-OBDR-MDS
-- dbgap Consent Code: c1
-- Potential Duo Code: DS / disease specific research
- Disease-Specific (Orofacial birth defects, MDS)
-- Consent Type Code: DS-OBD-MDS
-- dbgap Consent Code: c2
-- Potential Duo Code: DS / disease specific research
- Disease-Specific (Oral Clefts, PUB, MDS)
-- Consent Type Code: DS-OC-PUB-MDS
-- dbgap Consent Code: c3
-- Potential Duo Code: DS / disease specific research AND DUO:0000019	PUB
- General Research Use
-- Consent Type Code: GRU
-- dbgap Consent Code: c5
-- Potential Duo Code: DUO:0000042 / GRU

Thoughts @allisonheath ?

baileyckelly:
[...]

PR #543, opened 2019-10-11
✨ Add new attributes to participant, biospecimen, seq exp

### LINKED ISSUE #540: Add `library_selection` and `library_prep` field to SequencingExperiment
The `library_selection` enums can come from GDC: https://docs.gdc.cancer.gov/Data_Dictionary/viewer/#?view=table-definition-view&id=read_group

@jharenza @yuankunzhu what values do you want `library_prep` to be able to accept? The only one I have is `polyA` - also maybe could you give a bit of a description we can use in the field description?

znatty22:
Closes #540, #541, #542 

**Note on DUO IDs**

- Added `biospecimen.duo_ids` which is a list
- Can query for biospecimens like this:
```
http://localhost:5000/biospecimens?duo_ids=DUO:0000021,DUO:0000025
```

The query will return all biospecimens that have both duo IDs in the biospecimen.duo_ids column

Returns something that looks like this:
```
{
    "_links": {
        "self": "/biospecimens"
    },
    "_status": {
        "code": 200,
        "message": "success"
    },
    "limit": 10,
    "results": [
        {
            "_links": {
                "biospecimen_diagnoses": "/biospecimen-diagnoses?biospecimen_id=BS_BS4K54ZN",
                "biospecimen_genomic_files": "/biospecimen-genomic-files?biospecimen_id=BS_BS4K54ZN",
                "collection": "/biospecimens",
                "diagnoses": "/diagnoses?biospecimen_id=BS_BS4K54ZN",
                "genomic_files": "/genomic-files?biospecimen_id=BS_BS4K54ZN",
                "participant": "/participants/PT_ZRFZ9E1C",
                "self": "/biospecimens/BS_BS4K54ZN",
                "sequencing_center": "/sequencing-centers/SC_DGDDQVYR"
            },
            "age_at_event_days": null,
            "analyte_type": "DNA",
            "composition": null,
            "concentration_mg_per_ml": null,
            "consent_type": null,
            "created_at": "2019-10-22T16:38:02.049002+00:00",
            "dbgap_consent_code": null,
            "duo_ids": [
                "DUO:0000021",
                "DUO:0000025"
            ],
            "external_aliquot_id": null,
            "external_sample_id": null,
            "kf_id": "BS_BS4K54ZN",
            "method_of_sample_procurement": null,
            "modified_at": "2019-10-22T16:38:02.049010+00:00",
            "ncit_id_anatomical_site": null,
            "ncit_id_tissue_type": null,
            "shipment_date": null,
            "shipment_origin": null,
            "source_text_anatomical_site": null,
            "source_text_tissue_type": null,
            "source_text_tumor_descriptor": null,
            "spatial_descriptor": null,
            "uberon_id_anatomical_site": null,
            "visible": true,
            "volume_ul": null
        }
    ],
    "total": 1
}
```


dataservice/api/biospecimen/models.py
```diff
@@ -79,6 +79,10 @@ class Biospecimen(db.Model, Base):
                                           'Uber-anatomy ontology which '
                                           'represents harmonized anatomical'
                                           ' ontologies')
+    duo_id_biospecimen = db.Column(db.Text(),
```
dankolbman: Why not just `duo_id` since it's already on the biospecimen?

dataservice/api/biospecimen/schemas.py
```diff
@@ -19,6 +19,15 @@
 SAMPLE_PROCUREMENT_ENUM = {"Autopsy", "Biopsy", "Subtotal Resections",
                            "Gross Total Resections", "Blood Draw",
                            "Other"}
+# Codes from https://github.com/EBISPOT/DUO/blob/master/src/ontology/duo.csv
```
dankolbman: This url may be the best source of truth, even though I think it's essentially the same as that csv: http://purl.obolibrary.org/obo/duo.owl
We should probably avoid dynamic generation for now so in the case a change occurs that requires some sort of manual migration.

dataservice/api/biospecimen/models.py
znatty22: 🤷‍♀ I just took it from the ticket: https://github.com/kids-first/kf-api-dataservice/issues/541#issue-504742357

Yea I guess `_biospecimen` is redundant though

dataservice/api/common/validation.py
```diff
@@ -47,3 +47,54 @@ def enum_validation_generator(_enum, common=True):
...
+        :raises: ValidationError when validation fails
+        """
+        assert isinstance(input_items, list), (
+            'Parameter `input_items` must be a list'
+        )
+        # Don't do validation for this case, just return. See NOTE in docstring
+        if len(input_items) == 1 and any([d in input_items[0]
+                                          for d in [',', ' ', ';']]):
+            return
+
```
znatty22, line 90: We could make this a case insensitive validation. But all of our other enum validations are case sensitive so I followed the same pattern.

dankolbman (commented):
Regarding the DUO field, why have we decided to put it on the biospecimen? @allisonheath @baileyckelly 
The community seems to have designed DUO for the dataset (study) level, which won't work for us. Biospecimen does seem more appropriate, but what about when we have cohort analyses with heterogeneous consent? Would it not be more accurate to annotate the files themselves with DUO, or do we try to dynamically derive this value from biospecimen. Or perhaps we don't wish to use DUO for any sort of access, but merely for search and discovery?

allisonheath:
Currently this is "merely" for search and discovery and examining issues like the ones you bring up. So for now it should go on biospecimen because that's where the dbGaP consents are attached as well. 

I agree, much more thought would need to go into using it for any sort of access control, but this is the first step.

tests/test_filter_params.py
```diff
@@ -27,19 +27,35 @@ def test_filter_params(self, client, entities, model):
...
+            else:
+                query_params[k] = v
+
+        # Non-list type filter params
+        expected_total = q.filter_by(**query_params).count()
 
         # Make query string
-        qs = urlencode(filter_params)
+        for k, v in list_params.items():
+            query_params[k] = ','.join(v)
```
liberaliscomputing, line 47: This is a minor suggestion, but since the iteration of `list_params` is not necessary, 

```python
q = model.query

query_params = {}
expected_total = 0
for k, v in filter_params.items():
    if isinstance(v, list):
        q = q.filter(getattr(model, k).contains(v))
        query_params[k] = ','.join(v)
        expected_total += 1
    else:
        query_params[k] = v

expected_total = q.filter_by(**query_params).count() - expected_total
```

dataservice/api/common/validation.py
liberaliscomputing: I am also down for the case insensitive validation. We may be able to lowercase query params and enumeration constants and validate if they match. Once matched, we can map the source value to the controlled value in  enumeration.

dataservice/api/common/validation.py
```diff
@@ -47,3 +47,54 @@ def enum_validation_generator(_enum, common=True):
...
+        valid_set = {str(valid_item) for valid_item in valid_items}
+        invalid_set = {item for item in input_items
+                       if str(item) not in valid_set}
```
dankolbman: I think you can avoid the comprehensions here and reduce to one line:
```
invalid_set = set(input_items) - set(valid_items)
```

tests/test_filter_params.py
```diff
@@ -2,7 +2,7 @@
 import json
 from dateutil import parser, tz
 from urllib.parse import urlencode
-
+from pprint import pprint
```
dankolbman, line 5: Do you still need this?

znatty22: Nope

tests/test_filter_params.py
znatty22: Hey @liberaliscomputing so this won't work unfortunately ... let me give you an example:

let's say we're querying Biospcimen and the `filter_params` equals: 

```
{
    'analyte_type': 'DNA',
    'duo_ids': ['DUO:0000021', 'DUO:0000025']
} 
```

then query params will equal:

```
{
    'analyte_type': 'DNA',
    'duo_ids': 'DUO:0000021,DUO:0000025'
} 
```

which will go into the `filter_by` method and result in a SQL query roughly like
this:

```
select * from biospecimen
where analyte_type='DNA' and duo_ids='DUO:0000021,DUO:0000025' and
<however the .contains resolves to SQL>
```

Since `duo_ids` column has an ARRAY type that query ^ won't actually get us 
what we want: specimens with analyte_type = DNA and whose duo_ids list
contains both `DUO:0000021`, and `DUO:0000025`

dataservice/api/common/validation.py
znatty22: So I thought about it some more and I've decided against case insensitive validation for a couple reasons: 

- I want to follow the same validation pattern as with enum validation where
we currently don't do case insensitive validation

- Allowing case insensitive validation in general would cause us to have to change/add other code that I don't think is worth it right now. 

Basically, if the validators ignore case then people can put in things like  DNA or dna or DnA. That's fine but then if we want values to "look uniform" in  the db then we have to do something to modify (e.g. title case everything) text values before they get into the db. Or if its really just about the values looking uniform on the portal, then we'd have to clean the text values on serialization.  

I'm sure there's various ways of doing this (post deserialization in marshmallow, or in the base model.py class on all db.Text type columns, etc) but like I said I don't think its worth it to figure this out now.

dataservice/api/common/validation.py
znatty22: Ok 👍 I implemented ^ bc at first I was doing case insensitive validation. I've decided against doing case insensitive validation (see reason in [comment](https://github.com/kids-first/kf-api-dataservice/pull/543#discussion_r340392528) earlier) so this should be fine

dankolbman (approved):
💎

liberaliscomputing (approved):
🎉

PR #544, opened 2019-10-22
⬆️ Bump pytest to 5.2.0

znatty22:
CircleCI is failing: 
image

**Notes**

- Upgrading to this version of pytest solves the problem. See details on issue on SO here: 
https://stackoverflow.com/questions/58189683/typeerror-attrib-got-an-unexpected-keyword-argument-convert

- As a result we need to change some test code bc pytest 5.2 deprecated the feature where you can call fixtures directly. See https://docs.pytest.org/en/latest/deprecations.html#calling-fixtures-directly

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,5 +1,5 @@
 codacy-coverage==1.3.10
-pytest==3.3.1
+pytest==5.2.0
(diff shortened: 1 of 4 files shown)


dankolbman (approved):
💾

PR #545, opened 2019-10-22
📝 Update README w new ERAlchemy install instruction

### LINKED ISSUE #110: relationships among entities on ER diagram are wrong
for example, "participant" and "diagnosis", the label on the "participant" side should be {0,1}, which means a diagnosis belongs to one participant. Also, the label on the "diagnosis" side should be {0, N}, which means one participant could have zero or more diagnosis

At the same time, I question 0 in multiplicity. If a diagnosis exists, it should belong to a participant. "0" here doesn't make sense

znatty22:
Fixes #110


dankolbman (approved):
🥇

PR #548, opened 2019-11-14
⏪ Revert authz changes

dankolbman:
Undo, ctrl+z

Reverts kids-first/kf-api-dataservice#539


znatty22 (commented):
Lol why?

znatty22 (approved):
Talked to allison. I guess we're not going to upgrade to new Gen3, so seems fine now

PR #550, opened 2019-12-06
✨ Add study fields: short_code, program, domain

### LINKED ISSUE #547: New field for study code
Right now for a study we have:
- `kf_id` (good for systems, not good for humans)
- `name` (often long, more like a paper title, not good for systems)
- `short_name` (shorter, but still not memorable, not good for systems)

None of these fit two increasing needs of:
1) Visual display on the portal, the graphs especially have gotten unwieldy with multiple studies, but also the main study facet

Screen Shot 2019-11-06 at 9 11 54 AM
 
2) Human reference when talking about studies, the study IDs we can't remember, names or investigator names often get mixed up

I had been a bit resistant to creating what felt like yet another identifier with a study code, as well as it has some implicit semantic meaning typically, but it now seems more like the right solution. So proposing we created a `study_code` field in data service on the study. Then we can generate TCGA/ICGC like codes for each e.g.:

`Kids First: Orofacial Cleft - European Ancestry` -> `KF-OFCE`
`Pediatric Brain Tumor Atlas: CBTTC ` -> `PBTA-CBTTC`

And then for those from other places that already have it, we can reuse those, e.g.
`TARGET: Neuroblastoma` -> `TARGET-NBL`

👍 👎 ?

znatty22:
Closes #547

### DIFF
--- a/dataservice/api/study/models.py
+++ b/dataservice/api/study/models.py
@@ -36,6 +36,18 @@ class Study(db.Model, Base):
                         doc='dbGaP version')
     name = db.Column(db.Text(),
                      doc='Name or title of study')
+    program = db.Column(
+        db.Text(),
+        doc='The project/program the study is a part of'
+    )
+    domain = db.Column(
+        db.Text(),
+        doc='The diagnosis domain of the study (e.g. Cancer)'
+    )
+    short_code = db.Column(
+        db.Text(),
+        doc='Alphanumeric string to identify the study'
+    )
     short_name = db.Column(db.Text(),
                            doc='Short name for study')
     attribution = db.Column(db.Text(),
--- a/migrations/versions/475214ed802d_.py
+++ b/migrations/versions/475214ed802d_.py
@@ -1,0 +1,27 @@
+"""
+1.14.0 Add short_code, program, domain to Study
+
+Revision ID: 475214ed802d
+Revises: 12850dab14e9
+Create Date: 2021-06-09 00:28:30.170282
+"""
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = '475214ed802d'
+down_revision = '12850dab14e9'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('study', sa.Column('short_code', sa.Text(), nullable=True))
+    op.add_column('study', sa.Column('domain', sa.Text(), nullable=True))
+    op.add_column('study', sa.Column('program', sa.Text(), nullable=True))
+
+
+def downgrade():
+    op.drop_column('study', 'short_code')
+    op.drop_column('study', 'program')
+    op.drop_column('study', 'domain')
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -165,6 +165,8 @@ def make_entities(client):
                             'filter_params')[endpoint]['valid'])
                 if model == Participant:
                     data['external_id'] = 'Participant_{}'.format(i)
+                if model == Study:
+                    data['short_code'] = f'KF_ST{i}'
                 if model == SequencingCenter:
                     _name = 'SequencingCenter_{}'.format(i)
                     data['name'] = _name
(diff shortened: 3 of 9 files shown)


dataservice/api/common/validation.py
```diff
@@ -97,3 +97,39 @@ def validate_list(input_items):
                 f'valid set: {", ".join(valid_items)}'
             )
     return validate_list
+
+
+def alphanumeric_str_validation_generator(field_name, quantifier_regex='+'):
```
dankolbman: Why do you need to take `quantifier_regex` as a parameter if it seems to always be constant?

dataservice/api/common/validation.py
```diff
@@ -97,3 +97,39 @@ def validate_list(input_items):
...
+    length of 3 and max length of 6.
+
+    :param field_name: Name of the field/attribute that is being validated
+    :type field_name: str
+    :param quantifier_regex: Regex pattern to check for min, max length.
+    :type quantifier_regex: str
+
+    :returns: validation function
+    """
+    assert isinstance(quantifier_regex, str), (
```
dankolbman: Maybe you'd be safer asserting that this is valid regex via `re.compile` or similar?

dankolbman (commented):
Just want to note that we won't be able to make new studies until we've accounted for this change in the study creator and data tracker.

dataservice/api/common/validation.py
znatty22: What do you mean by `always be constant`? If we have another field which requires alphanumeric str validation but has a different min/max length requirement, then you'd need to specify this to the validation func generator. Or am I not understanding the question?

dataservice/api/common/validation.py
znatty22: Good idea 👍

dataservice/api/common/validation.py
dankolbman: Do we have another field that requires it? If not, why not simply validate the length requirements statically instead of having an unecessary abstraction?

znatty22: We don't have one now, but I thought it doesn't hurt to implement this in a generic way so that it could be used in the future if needed

If we're going to make the length validation constant then we may as well get rid of the validation func generator and just implement a specific `validate_study_code` method in `dataservice/study/schema.py`

dataservice/api/study/schemas.py
```diff
@@ -9,6 +12,25 @@
 RELEASE_STATUS_ENUM = {'Pending', 'Waiting', 'Running', 'Staged',
                        'Publishing', 'Published', 'Failed', 'Canceled'}
 
+PROGRAM_ENUM = {
```
znatty22: Don't make this an enum since Data Tracker now supports multi tenants. The tenant roughly corresponds to the program and can be added at runtime to Data Tracker.

dataservice/api/study/models.py
```diff
@@ -36,6 +36,16 @@ class Study(db.Model, Base):
...
                      doc='Name or title of study')
+    program = db.Column(
+        db.Text(),
+        doc='The project/program the study is a part of'
+    )
+    domain = db.Column(
+        db.Text(),
+        doc='The diagnosis domain of the study (e.g. Cancer)'
+    )
+    short_code = db.Column(db.Text(), unique=True, nullable=False,
```
znatty22: Might be better to remove the unique and nullable constraints here so that we can do the data migration and apply constraints on the study creator side. This would probably simplify things a lot

znatty22: Also just learned that some studies won't have a code/domain/program

dataservice/api/study/schemas.py
```diff
@@ -9,6 +12,25 @@
...
+    'Kids First', 'Pediatric Brain Tumor Atlas', 'TARGET', 'ICR'
+}
+DOMAIN_ENUM = {
```
znatty22: Add Other

dankolbman (approved):
# 🐲

Looks good! Looks like this one has been sitting around for a while.

gsantia (approved):

PR #565, opened 2021-01-14
🐛 Allow larger values for SE total_reads 

fiendish:
I'm getting numbers that require 32 bits, but a 4byte signed int field only accepts up to 31.


dankolbman (commented):
I think you need to generate a migration for this?

dankolbman (approved):
# 🍟

LGTM

PR #566, opened 2021-01-19
📝 Update development init instructions

fiendish:
Postgres added a password requirement in a bug fix update. See https://github.com/docker-library/postgres/issues/681


dankolbman (approved):
# 🌯

Nice. I think this is something that changed with newer docker images of postgres and we may have some others that need to be updated.

PR #600, opened 2023-03-07
🔥 remove polya_stranded library_prep

chris-s-friedman:
we have columns in the ds for library_prep and for stranded-ness. having a library prep enum for polya_stranded combines these two columns when we want to keep them seperated.


youngnm (approved):
Looks good!

PR #608, opened 2023-05-08
💥 Major API upgrade

znatty22:
Closes #612 

- Collects all of the upgrades in a single branch so that we can keep testing the full upgrade in an isolated environment. 
- We can also use this branch to cherry pick the commits we want to iteratively merge into master via PRs that mirror the ones merging into this branch

PR #613, opened 2023-05-10
🔼 Upgrade migration tooling

znatty22:
## Todo

- [x] Do a test migration to ensure things work as expected
- [ ] Needs to be rebased on master after #611 is merged

## Motivation
We have not upgraded our migration tooling since 2018

## Approach
- Updated migration engine: `alembic`
    - Includes new auto-generated config (`alembic.ini`) and script (`env.py`) in `migrations/`
- Updated web framework integration with migration engine: `Flask-Migrate`
- ^ Required us to also updated ORM to min version 1.3: `SQLAlchemy`
- Fixed some tests that broke bc SQLAlchemy fixed some bugs and is stricter with how it does some things


znatty22:
Closing due to change in direction/priorities. We are not moving forward with upgrade. Any upgrade work that we choose to do in the future will be in #608. From there we can cherry pick commits and make PRs against master if we so desire

PR #614, opened 2023-05-16
📝 point documentation at the kidsfirstdrc.org domain, not kids-first.io

chris-s-friedman:
The documentation pointed to kids-first.io. Update that documentation to reference kidsfirstdrc.org

### DIFF
--- a/dataservice/api/README.md
+++ b/dataservice/api/README.md
@@ -19,12 +19,12 @@ Curl is the quickest way to communicate with the api from a unix shell
 
 Get the system status:
 ```bash
-curl -H "Content-Type: application/json" http://kf-api-dataservice-qa.kids-first.io/status
+curl -H "Content-Type: application/json" https://kf-api-dataservice.kidsfirstdrc.org/status
 ```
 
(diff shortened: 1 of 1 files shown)


calkinsh (approved):

znatty22 (changes requested):
I'm wondering why all of the urls are referring to the QA deployment. Probably should be prod? So can just remove the `-qa` from `https://kf-api-dataservice`

chris-s-friedman:
@znatty22 unless you feel particularly strong about it, I'm leaving the two commits separate. I see them logically different : 

1. [2cd3437](https://github.com/kids-first/kf-api-dataservice/pull/614/commits/2cd3437505266120059925357192f839a2d92b9f) to fix the documentation to use the correct domain
2. [1e3f627](https://github.com/kids-first/kf-api-dataservice/pull/614/commits/1e3f627eaf40d5629c858226ff90e4456085087d) change the documentation to reference prod instead of qa

znatty22 (approved):
Just need to rebase on master, and then its good to go

PR #622, opened 2023-06-06
🐛 Fix indexd POST error handling

### LINKED ISSUE #523: indexd.py does not correctly propagate indexd error messages to final response
POSTing a new genomic file with invalid indexd fields (invalid hash algo, wrong field type, missing req field, etc) should propagate the 400 status code and error message that comes from indexd. Right now the original error message gets lost and a 500 is returned from the dataservice with the default html Internal Server Error message. 

This is likely due to this line: https://github.com/kids-first/kf-api-dataservice/blob/88dec39d4d8c22b0e1606a3c3853502a07de347e/dataservice/extensions/flask_indexd.py#L127

Any `raise_for_status` call should probably be in a try/except like this part of the code here: 
https://github.com/kids-first/kf-api-dataservice/blob/88dec39d4d8c22b0e1606a3c3853502a07de347e/dataservice/extensions/flask_indexd.py#L233

znatty22:
Closes #523

### DIFF
--- a/dataservice/extensions/flask_indexd.py
+++ b/dataservice/extensions/flask_indexd.py
@@ -124,7 +124,18 @@ class Indexd(object):
         resp = self.session.post(self.url,
                                  json=req_body)
 
-        resp.raise_for_status()
+        try:
+            resp.raise_for_status()
+        except HTTPError:
+            # Attach indexd error message, if there is one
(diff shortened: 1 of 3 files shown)


chris-s-friedman (approved):
👍

PR #624, opened 2023-06-13
✨ Store ACL in new authz field

znatty22:
## Motivation

Authorization policy information for indexd files is currently stored in the `acl` list field. However, the `acl` field is being replaced with a new `authz` list field. The values in the `authz` list will correspond to policies created in [Arborist](https://github.com/uc-cdis/arborist). 

Some important things to note:

- If the `authz` field is populated, any values in the `acl` field are ignored
- If the `authz` field is NOT populated, the values in the `acl` field are used as a fallback
- Because of the above two facts, we should manage the deprecated `acl` conservatively by setting it to an empty list (no access) by default

#### New `authz` Field

Add an`authz` field for Indexd documents similar to other fields that are relayed through dataservice to/from indexd.
This field is not persisted inside of the dataservice, but proxied to and from Indexd where it is stored.

#### Disable `acl` Field for New Records

An error will be returned when a record is created with a non-empty `acl` field.
This will force any new data to be created using the new `authz` field instead of using the soon to be deprecated `acl` field.

```shell
$ curl -X POST localhost:5000/genomic-files -d '{"acl": ["abc", "123"], "urls": ["s3://abc/abc"], "hashes": {"md5":"e4b246e5bac1f152b3b13047a4876b84"}, "size": 10}' | jq
{
  "_status": {
    "code": 400,
    "message": "could not create genomic_file: {'acl': ['The ACL field has been deprecated. Please use the authz field instead.']}"
  }
}
```

## Migrations

Gen3 Indexd has a migration script that will populate `authz` fields based on the `acl` field. This script will be run to update everything to `authz` policies. After this has occurred, we will be able to manually set all record's `acl` to be empty to ensure that it is not used when determining access to the record.

### DIFF
--- a/dataservice/extensions/flask_indexd.py
+++ b/dataservice/extensions/flask_indexd.py
@@ -182,7 +182,7 @@ class Indexd(object):
 
         # If acl changed, update all previous version with new acl
         if record.acl != old['acl']:
-            self._update_all_acls(record)
+            self._update_all_acls(record, key="acl")
 
         # If authz changed, update all previous version with new authz
         if record.authz != old['authz']:
@@ -204,10 +204,18 @@ class Indexd(object):
 
         return record
 
-    def _update_all_acls(self, record, key="acl"):
+    def _update_all_acls(self, record, key="authz"):
         """
-        Update acls for all previous versions of a record and update the
+        Update ACL values for all previous versions of a record and update the
         target record's rev
+
+        Since the GenomicFile.acl field is being deprecated in favor of the
+        new GenomicFile.authz field, the default behavior of this method will
+        be to update the ACL values in the authz field
+
+        Until the GenomicFile.acl field is completely removed from the code
+        base and all GenomicFile.acl values in the DB are migrated into the
+        GenomicFile.authz field, this method will be used to update both fields
         """
         # Only use fields allowed by the indexd PUT schema
         fields = ['urls', 'acl', 'authz', 'file_name', 'version',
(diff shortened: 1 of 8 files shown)


dataservice/extensions/flask_indexd.py
```diff
@@ -197,22 +204,22 @@ def update(self, record):
 
         return record
 
-    def _update_all_acls(self, record):
+    def _update_all_acls(self, record, key="acl"):
```
chris-s-friedman: 1. if we're keeping the default for `key` to be "acl" for backwards compatibility and going forward, `authz` should be used, can you document that in the docstring for this function, that the expectation is to use `authz` unless there is some particular reason why "acl" should be used. 
2. If we are deprecating acl, we should change the name of this function to represent that change. If the worry is backwards compatibility, but going forward `"authz"` should be the `key` value used, perhaps we rename this function but add a pointer with the old function name, i.e.:

```python
    def _update_all_access_things(self, record, key='authz'):
        """
        update key field's values for all previous versions of a record and
        update the target record's rev. 
        """
    ...


    def _update_all_acls(self, record):
        """
        Update acls for all previous versions of a record and
        update the target record's rev. Pointer to 
        _update_all_access_things for backwards compatibility
        """
        self._update_all_access_things(record = record, key="acl")
```

chris-s-friedman (changes requested):
Looks great, Natasha! Besides my one comment, the other thing I'd want to see is a notice in the swagger docs that the acl field is deprecated. When consulting the docs, users should see a warning that if they try to register something with a value in the acl field, they will get a 400 error.

znatty22:
@chris-s-friedman The best I could do to document the ACL field deprecation in the API docs is:

image

Unfortunately the swagger generation lib has changed too much for me to find the old version of the docs and figure out how to annotate the deprecation on the actual schema definition for POSTing genomic files :(

chris-s-friedman (approved):
Looks great!

PR #629, opened 2023-08-08
🔧 Increase max and default page limits

znatty22:
## Motivation
The kf-api-study-creator API often queries the Dataservice when it syncs studies between Dataservice and itself. Unfortunately there is a bug in study creator where it only ever requests the first 100 studies since this is the maximum # of resources you can fetch in a single request. 

Additionally, several other API consumers have long been requesting the maximum page size be increased from 100 to something larger so that they can send less requests.

## Approach

- Increase the maximum page size from 100 to 1000
- Increase the default page size from 10 to 100


README.md
```diff
@@ -90,6 +90,11 @@ https://realpython.com/intro-to-pyenv/
 # Setup python environment and install dependencies
 pyenv virtualenv 3.7.11 dataservice_venv
 pyenv local dataservice_venv
+
+# Important but *temporary*
+# See https://github.com/yaml/pyyaml/issues/724
+pip install "cython<3.0.0" && pip install --no-build-isolation "pyyaml==5.4.0"
```
devbyaccident, line 96: That's an annoying issue. Is there something that we need pyyaml at `5.4.0` for, or can we take a newer version like `pyyaml<6.0`? Would rather not hard-code a specific version to change later if we can avoid it.

Also, https://github.com/yaml/pyyaml/issues/724#issuecomment-1640073508

znatty22: @devbyaccident Lol that image is so accurate. 

We've always had pyyaml pinned to 5.4.0, and I believe its bc a while ago, something broke and caused us to pin it. So I don't want to change anything that we don't absolutely have to in this PR. I can try changing it in another PR so the change isolated

devbyaccident: Gotchya, that makes sense. Ok, I'm happy to give it the 👍 , but I can't speak as much for the non-docker/package changes. Would still like someone from @kids-first/dataservice-reviewers to take a look.

devbyaccident (approved):
👍

alubneuski (approved):

PR #637, opened 2023-10-20
:heavy_plus_sign: adding ONT WGS strategy enum

awarkow:
adding ONT WGS as seq strategy for use with long reads files. Was left out of the previous release mistakenly


Christina-J-Diaz:
looks good!

Christina-J-Diaz (approved):

PR #643, opened 2023-12-12
✨ Add Sample and Container

znatty22:
## Motivation

The current biospecimen model does not adequately capture lineage of biospecimens or biological equivalency among groups of biospecimens. This information is important for several reasons. One reason is when someone needs to know the total remaining quantity of  aliquots/biospecimens that can be used for additional sequencing or processing. 

## Approach
Introduce 2 new tables: `Sample` and `Container`. These tables will be populated by deriving containers and samples from the incoming Biospecimen in a POST/PATCH request (#645). 

- `Sample`: represents a biologically distinct amount of material from a single person. A sample will also be used to group biospecimens (also known as aliquots or containers) that are biologically equivalent. 
- `Container`: an extracted portion of a Sample

sample-container-erd

**Note:** We wanted to take this approach to minimize or eliminate any changes to the current biospecimen table so that current functionality is not disrupted in any way.

### Database
- Create a `Sample` table 
- Create a `Container` table

#### Delete Life Cycles
- When a Participant is deleted its child Samples are also deleted
- When a Biospecimen is deleted its associated Container is deleted
- When a Sample is deleted its child Containers are also deleted
- When a Sample has 0 child Containers, it will be deleted. This is considered an orphan sample

### API
- Implement REST API for both `Sample` and `Container`
- In the future we may remove POST/PUT/PATCH/DELETE from the API for these two concepts since eventually the create, update, delete lifecycles will be managed internally by the API.

## Todo

- [ ] Add db schema migrations once finalized
- [x] (#645) Implement logic that will automatically create/update/delete rows in `Sample` and `Container` each time a Biospecimen is created/updated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,3 +107,7 @@ ENV/
 
 # hidden Mac Desktop Services Store files
 .DS_Store
+
+# vim
+*.swo 
+*.swp
--- a/dataservice/__init__.py
+++ b/dataservice/__init__.py
@@ -19,6 +19,8 @@ from dataservice.api.family_relationship.models import FamilyRelationship
 from dataservice.api.outcome.models import Outcome
 from dataservice.api.phenotype.models import Phenotype
 from dataservice.api.biospecimen.models import Biospecimen
+from dataservice.api.sample.models import Sample
+from dataservice.api.container.models import Container
 from dataservice.api.diagnosis.models import Diagnosis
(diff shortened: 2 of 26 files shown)


znatty22:
Closing for now. New approach is to implement the Sample table only and keep it simple. This is an MVP to meet Portal Beta requirements

PR #670, opened 2024-06-06
♻️ Refactor biobank_contact to be two columns

znatty22:
FHIR Research study needs both the name and the email for the biobank contact


Christina-J-Diaz (approved):
looks good!

PR #672, opened 2024-06-11
♻️ Allow parent or child sample to be null

### LINKED ISSUE #671: Parent should be nullable in Sample Relationship
We've decided that the easiest way to identify roots of sample trees is to have a row for the root where it appears in the child column with a null parent. We need to be able to identify roots of sample trees in order to correctly load the samples into the FHIR server and also to be able to easily identify the roots for query/analysis purposes.

znatty22:
Closes #671


calkinsh:
I dont know if we want child sample to ever be null. Just parent? I can't think of a use case where we'd have a parent with null child, only cases where we'd have a child with null parent (being the "top" of the tree)

dataservice/api/sample_relationship/models.py
```diff
@@ -32,18 +32,23 @@ class SampleRelationship(db.Model, Base):
...
         db.ForeignKey('sample.kf_id'),
-        nullable=False,
+        nullable=True,
         doc='kf_id of one sample in the relationship')
 
     child_id = db.Column(
         KfId(),
         db.ForeignKey('sample.kf_id'),
-        nullable=False,
+        nullable=True,
```
calkinsh, line 41: keep this false, I think

dataservice/api/sample_relationship/models.py
```diff
@@ -119,35 +124,47 @@ def validate_sample_relationship(target):
...
 
-    # Get samples in relationship by id
-    parent = Sample.query.get(target.parent_id)
-    child = Sample.query.get(target.child_id)
+    parent_id = target.parent_id
+    child_id = target.child_id
 
-    # Check that both are existing samples
-    if not (parent and child):
+    # Check that at least 1 sample ID is non-null
```
calkinsh, line 130: check that at least child sample ID is not null

calkinsh (changes requested):

dataservice/api/sample_relationship/models.py
znatty22: I wasn't sure what to do about this. I thought maybe we make this nullable too bc maybe there is a time when a parent sample gets added in but it doesn't have children yet? Idk if that is realistic but that was the motivation

calkinsh (approved):

PR #681, opened 2024-08-28
✨ D3B-674 Add cavatica related gf fields

chris-s-friedman:


HuangXiaoyan0106 (approved):

znatty22 (commented):
Need to add a test to the genomic file API tests. If you add new fields to https://github.com/kids-first/kf-api-dataservice/blob/f96f93aaa073c16668a4861757d897be04cf3aa6/tests/genomic_file/test_genomic_file_resources.py#L512, that should take care of `test_genomic_file_resources.py`

znatty22 (commented):
Also please add the new fields to `data.json`. This gets used in the general API tests which test for  things like filters, bad values, etc: https://github.com/kids-first/kf-api-dataservice/blob/f96f93aaa073c16668a4861757d897be04cf3aa6/tests/data.json#L112

znatty22 (commented):
We also need the db schema migration:

```
# this creates the migration
flask db migrate -m "name of migration"

# this applies the migration to the db
flask db upgrade
```

Make sure to add the semantic version number in the docstring of the migration py file

migrations/versions/869f891d97ad_add_cavatica_gf_fields.py
```diff
@@ -0,0 +1,32 @@
+"""add cavatica gf fields
```
znatty22:
```suggestion
"""
1.25.0 Add cavatica gf fields
```

znatty22 (approved):
Looks good, just one minor comment on the migration docstring!

PR #684, opened 2024-09-05
👨🏻‍💻 Always bind all interfaces when app runs in docker-compose

znatty22:
Make it easier for developers to run the app when they're using docker-compose. 

Previously, the developer had to make some changes to the code in order to run and develop the API inside the docker container. 

The changes here make it so that the developer doesn't have to change anything when running the app in the docker-compose stack.

**NOTE** If running the API in a docker container, binding to all interfaces does not expose your API to the world. 
See https://www.reddit.com/r/docker/comments/xwfm08/why_do_i_need_to_specify_host0000_when_running_a/


haynescd (approved):
👍

PR #688, opened 2024-09-09
:bug: Update Tag of dataservice image

haynescd:


.github/workflows/deploy.yml
```diff
@@ -38,9 +38,15 @@ jobs:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set git tag and SHA
+        run: |
+          VERSION=$(git tag --sort=committerdate | grep -E '[0-9]' | tail -1 | cut -b 1-7)
+          SHORT_SHA=${GITHUB_SHA:0:7}
+          TAG="${VERSION}-${SHORT_SHA}"
```
znatty22, line 45: Does the next block in the action have access to the `TAG` variable? or do you have to export it `export TAG=foo`?

haynescd: echo "TAG=${TAG}" >> $GITHUB_ENV

This line should be what saves the env var for other steps to use

znatty22 (approved):
Looks good!

